PR #214, opened 2023-09-18
Create dashboards on installation and require `warehouse_id` on installation

nfx:
This PR adds framework to automatically add Redash queries / vizualiations / widgets / dashboards from simple file definitions


larsgeorge-db (commented):
lgtm

larsgeorge-db:
This raises a good point... should the dashboard operate on the summaries the assessment step creates, or on the inventory table(s). I am not for the latter right now as it means implementing the logic to detect things twice, once in the assessment code and then again in the dashboard code. I would rather prefer to have this being created by the assessment step and stored in a report table, then the dashboard simply visualizes the aggregates. Thoughts @nfx @FastLee?

nfx:
Dashboards are created from SQL files
image

nfx:
image
readme notebook gets one more link

nfx:
Assessment report is refreshed after all jobs
image

nfx:
and dashboard is auto-refereshed:
image

PR #217, opened 2023-09-19
Allow empty group filter in install.py (#216)

### LINKED ISSUE #216: Allow empty group filter in `install.py`
The script right now wants users to enter a list of groups to migrate. Allow to not enter anything to trigger the automatic discovery of groups instead.

larsgeorge-db:
Also fixes stripping whitespaces around groups when enters like `g1, g2, g3` etc. Previously the code expected to get no whitespaces.

Fixes #216


nfx (approved):
👍

PR #221, opened 2023-09-19
Remove .python-version as it pins the python version which is unneces…

dmoore247:
…sary

Removed `.python-version` as this was causing `./install.sh` problems


nfx (approved):
👍

PR #223, opened 2023-09-19
Update README.md

dmoore247:
Remind user to clone the repo (fork is better) prior to running install.sh (otherwise the user won't have the install.sh file.)


nfx (approved):
👍

Issue #229, opened 2023-09-20
Deeply nested types fail with Error: `type expected at the position 34508 of 'string:bigint:string:string:string:string:string:int:int:int:array<string>:array...`


nfx:
We have extreme error conditions, like this one:

```
[...]

PR #238, opened 2023-09-20
Fix describe throwing errors when listing

william-conti:
Fixes https://github.com/databrickslabs/ucx/issues/229


nfx (approved):
Thank you!

Issue #199, opened 2023-09-14
Tune-up the documentation for UCX


tamilselvanveeramani:
tune-up the documentation for UCX

PR #239, opened 2023-09-20
Refine the documentation to include the 'hive_metastore' catalog and 'ucx' schema in all the output generated by the assignment tool

tamilselvanveeramani:


tamilselvanveeramani:
Part of issue #199

src/databricks/labs/ucx/runtime.py
```diff
@@ -28,18 +28,18 @@ def crawl_tables(_: MigrationConfig):
     storage location details.
 
     The gathered metadata is then subsequently organized and documented within a designated storage entity referred to
-    as the `$inventory.tables` table. This table serves as an extensive inventory, offering a well-structured and
-    readily accessible point of reference for users, data engineers, and administrators."""
+    as the `hive_metastore.ucx.tables` table. This table serves as an extensive inventory, offering a well-structured
```
nfx, line 31: keep the `$inventory` variable here. See

https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/install.py#L280-L281

nfx (changes requested):
why is this needed?

tamilselvanveeramani:
as it's has been taken care by @nfx part of https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/install.py#L280-L281

PR #246, opened 2023-09-20
Adding safe return on grants

william-conti:
Adding a safe return for grants to handle corner cases

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/grants.py
+++ b/src/databricks/labs/ucx/hive_metastore/grants.py
@@ -6,6 +6,10 @@ from databricks.labs.ucx.framework.crawlers import CrawlerBase
 from databricks.labs.ucx.framework.parallel import ThreadedExecution
 from databricks.labs.ucx.hive_metastore.tables import TablesCrawler
 
+import logging
+
+logger = logging.getLogger(__name__)
+
 
 @dataclass(frozen=True)
 class Grant:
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/hive_metastore/grants.py
```diff
@@ -219,20 +222,27 @@ def _grants(
...
+        try:
+            object_type_normalization = {"SCHEMA": "DATABASE", "CATALOG$": "CATALOG"}
+            for row in self._fetch(f"SHOW GRANTS ON {on_type} {key}"):
+                (principal, action_type, object_type, _) = row
+                if object_type in object_type_normalization:
+                    object_type = object_type_normalization[object_type]
+                if on_type != object_type:
+                    logger.warning(
+                        f"Types for object {key} doesn't match after listing grants {object_type} != {on_type}"
+                    )
```
nfx, line 234:
```suggestion
                    logger.warning(
                        f"Types for object {key} doesn't match after listing grants {object_type} != {on_type}"
                    )
```
this will make a lot of noise if it's warning - there. either remove it or make it debug

william-conti: It's a corner case where the object type isn't the same as the one requested. For example you request the grant for a view and you get the grants for a table, it's not a normal behaviour and should be highlighted

nfx (approved):
👍

PR #248, opened 2023-09-20
Speedup and stabilise workspace listing algorithm

nfx:


nfx:
this PR is not relevant anymore

PR #259, opened 2023-09-21
Fixed `guess_external_locations` failure with `AttributeError: as_dict` and added an integration test

### LINKED ISSUE #256: `guess_external_locations` fails with `AttributeError: as_dict`
Failed on the test run:

```
     55 tables = self._backend.fetch(f"SELECT location FROM {self._schema}.tables WHERE location IS NOT NULL")
     56 mounts = Mounts(self._backend, self._ws, self._schema).snapshot()
---> 57 return self._external_locations(list(tables), list(mounts))

...

     25 external_locations = []
     26 for table in tables:
---> 27     location = table.as_dict()["location"]
     28     if location is not None and len(location) > 0:
     29         if location.startswith("dbfs:/mnt"):
```

Preceeding tasks succeeded:
image

FastLee:
Created unit testing
Still seeing weird behavior of the Row class.
Possibly related to Python version.

Fix #256

### DIFF
--- a/tests/integration/assessment/test_external_locations.py
+++ b/tests/integration/assessment/test_external_locations.py
@@ -1,0 +1,30 @@
+import logging
+import os
+
+from databricks.labs.ucx.framework.crawlers import StatementExecutionBackend
+from databricks.labs.ucx.hive_metastore.data_objects import ExternalLocationCrawler
+from databricks.labs.ucx.hive_metastore.list_mounts import Mount
+from databricks.labs.ucx.hive_metastore.tables import Table
+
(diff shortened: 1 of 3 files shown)


tests/integration/assessment/test_external_locations.py
```diff
@@ -0,0 +1,30 @@
+import logging
```
nfx, line 1: why this file is in `assessment` folder? it has to be in `hive_metastore`. please fix.

nfx (changes requested):

nfx (approved):

PR #275, opened 2023-09-22
Add inventory_database name check

dmoore247:
Check that database name is a word, loop until correct.

### DIFF
--- a/src/databricks/labs/ucx/install.py
+++ b/src/databricks/labs/ucx/install.py
@@ -163,7 +163,12 @@ class Installer:
                 raise err
 
         logger.info("Please answer a couple of questions to configure Unity Catalog migration")
-        inventory_database = self._question("Inventory Database", default="ucx")
+        while True:
+            inventory_database = self._question("Inventory Database stored in hive_metastore", default="ucx")
+            if re.match(r'^\w+$', inventory_database):
+                break
+            else:
+                print(f"{inventory_database} is not a valid database name")
 
         pro_warehouses = {"[Create new PRO SQL warehouse]": "create_new"} | {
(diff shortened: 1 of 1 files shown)


nfx (changes requested):

src/databricks/labs/ucx/install.py
```diff
@@ -163,7 +163,12 @@ def _configure(self):
                 raise err
 
         logger.info("Please answer a couple of questions to configure Unity Catalog migration")
-        inventory_database = self._question("Inventory Database", default="ucx")
+        while True:
```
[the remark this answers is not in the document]
dmoore247: :)

dmoore247:
Working on feedback, waiting for assessment job before re-testing installer.

nfx:
@dmoore247 you know you can kill installer by CTRL+C (or CMD+D, don't remember), right?..

larsgeorge-db (commented):
Not to say this is not useful, but this should also then be done for most of the other inputs, no? And even then, you can always edit the config file in the workspace afterwards.

dmoore247:
I think I can get to it next week.

nfx:
Following up - is this issue still relevant?

install
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+# This script will eventually be replaced with `databricks labs install ucx` command.
```
nfx, line 3: why do we need this file? please remove.

nfx (changes requested):
remove `install` in the root, otherwise lgtm

install
dmoore247: Removed

dmoore247 (commented):
removed `install` file

Issue #283, opened 2023-09-23
Allow disabling/skipping of workflow tasks via config


larsgeorge-db:
Some tasks in a workflow (here the group inventorizing) can take a long time but is not required for an assessment run. Add a config entry that allows to skip such tasks temproarily:

```
...
skip_tasks:
- crawl_grants
- crawl_mounts
```

nfx:
We must not allow that, otherwise users will disable what prepares the important dataset and things will break unexpectedly.

There's "Repair Run" button in the workflows ui that restarts only the failed tasks. Probably that needs a documentation, as some users don't know about the feature.

If you want ad-hoc experimentation for a task, there's debug notebook

larsgeorge-db:
This is not for repair runs, this is for advanced users that want to control the flow of jobs. Not all tasks are necessary given certain use-cases. We can leave this, if you prefer, undocumented and use this only under guidance.

nfx:
What is the use-case? The flow of jobs is always static and in a certain required order.

PR #284, opened 2023-09-23
Allow disabling/skipping of workflow tasks via config

larsgeorge-db:
Implements #283

### DIFF
--- a/src/databricks/labs/ucx/config.py
+++ b/src/databricks/labs/ucx/config.py
@@ -81,6 +81,7 @@ class MigrationConfig:
     connect: ConnectConfig | None = None
     num_threads: int | None = 10
     log_level: str | None = "INFO"
+    skip_tasks: list[str] | None = None
 
     # Starting path for notebooks and directories crawler
     workspace_start_path: str = "/"
@@ -124,6 +125,7 @@ class MigrationConfig:
             warehouse_id=raw.get("warehouse_id", None),
             num_threads=raw.get("num_threads", 10),
             log_level=raw.get("log_level", "INFO"),
+            skip_tasks=raw.get("skip_tasks", []),
         )
 
     @classmethod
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/config.py
```diff
@@ -124,6 +125,7 @@ def from_dict(cls, raw: dict) -> "MigrationConfig":
             warehouse_id=raw.get("warehouse_id", None),
             num_threads=raw.get("num_threads", 10),
             log_level=raw.get("log_level", "INFO"),
+            skip_tasks=raw.get("skip_tasks", []),
```
nfx, line 128: I am not convinced we need it.

nfx (changes requested):

src/databricks/labs/ucx/config.py
FastLee: I am. Most customers will start with Assessment only where the other tasks may come weeks and sometimes Months later.

nfx: then we need to split the jobs and not skip tasks.

FastLee: I would like them to be able to start with a quick assessment not messing up with the other tasks. Then redeploy when they are ready, making sure they have the latest code.

nfx: we can split the jobs with enough justification.

FastLee: Some customers may want to not deploy the group migration. Even look at it as a risk.
The assessment is a read only job and considered low risk.

larsgeorge-db:
Closing this for now, since this requires either a larger discussion on how to address slow tasks or how configure the assessment step. For now, refreshing the dashboard manually is ok.

PR #286, opened 2023-09-23
Keep order of tasks in the README notebook

### LINKED ISSUE #285: Keep order of tasks in the README notebook
Right now the pydoc snippets are inserted randomly for each step and random across steps.

larsgeorge-db:
Resolves #285

### DIFF
--- a/src/databricks/labs/ucx/framework/tasks.py
+++ b/src/databricks/labs/ucx/framework/tasks.py
@@ -8,6 +8,7 @@ from databricks.labs.ucx.config import WorkspaceConfig
 from databricks.labs.ucx.framework.logger import _install
 
 _TASKS: dict[str, "Task"] = {}
+_TASK_ORDER: list[str] = []
 
 
 @dataclass
@@ -63,6 +64,7 @@ def task(workflow, *, depends_on=None, job_cluster="main", notebook: str | None 
             notebook=notebook,
             dashboard=dashboard,
         )
+        _TASK_ORDER.append(func.__name__)
 
         return wrapper
 
--- a/src/databricks/labs/ucx/framework/tasks.py
+++ b/src/databricks/labs/ucx/framework/tasks.py
@@ -8,7 +8,6 @@ from databricks.labs.ucx.config import WorkspaceConfig
 from databricks.labs.ucx.framework.logger import _install
 
 _TASKS: dict[str, "Task"] = {}
-_TASK_ID_GENERATOR: int = 0
 
 
 @dataclass
@@ -55,11 +54,8 @@ def task(workflow, *, depends_on=None, job_cluster="main", notebook: str | None 
             msg = f"Task {func.__name__} must have documentation"
             raise SyntaxError(msg)
 
-        global _TASK_ID_GENERATOR
-        task_id = _TASK_ID_GENERATOR
-        _TASK_ID_GENERATOR += 1
         _TASKS[func.__name__] = Task(
-            task_id=task_id,
+            task_id=len(_TASKS),
             workflow=workflow,
             name=func.__name__,
             doc=func.__doc__,
--- a/src/databricks/labs/ucx/framework/tasks.py
+++ b/src/databricks/labs/ucx/framework/tasks.py
@@ -8,11 +8,12 @@ from databricks.labs.ucx.config import WorkspaceConfig
 from databricks.labs.ucx.framework.logger import _install
 
 _TASKS: dict[str, "Task"] = {}
-_TASK_ORDER: list[str] = []
+_TASK_ID_GENERATOR: int = 0
 
 
 @dataclass
 class Task:
+    task_id: int
     workflow: str
     name: str
     doc: str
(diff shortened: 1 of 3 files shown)


larsgeorge-db:
Works

image

src/databricks/labs/ucx/framework/tasks.py
```diff
@@ -63,6 +64,7 @@ def wrapper(*args, **kwargs):
             notebook=notebook,
             dashboard=dashboard,
         )
+        _TASK_ORDER.append(func.__name__)
```
nfx, line 67: Why do we need to introduce a separate data structure, that has more potential for getting out of sync due to a bug?

Why not a field in the Task class, that holds the definition order (and sorting the tasks before the readme generation) or a topological sort based on depends_on field?

src/databricks/labs/ucx/install.py
```diff
@@ -248,21 +248,29 @@ def _create_jobs(self):
...
     def _create_readme(self):
         md = [
             "# UCX - The Unity Catalog Migration Assistant",
             f'To troubleshoot, see [debug notebook]({self._notebook_link(f"{self._install_folder}/DEBUG.py")}).\n',
             "Here are the URL and descriptions of jobs that trigger's various stages of migration.",
             "All jobs are defined with necessary cluster configurations and DBR versions.",
         ]
-        for step_name, job_id in self._deployed_steps.items():
+        for step_name in self._step_list():
+            job_id = self._deployed_steps[step_name]
```
nfx, line 265: And what will happen if step is not deployed?

tests/unit/test_install.py
```diff
@@ -220,3 +220,55 @@ def test_choices_happy(mocker):
...
     res = install._choice("foo", ["a", "b"])
     assert "b" == res
+
+
+def test_step_list(mocker):
+    ws = mocker.Mock()
+    install = Installer(ws)
+    from databricks.labs.ucx.framework.tasks import _TASKS, Task
+
+    _TASKS.clear()
```
nfx, line 230: Unit tests may be run from multiple threads in a random order, so this one is not deterministic.

nfx (changes requested):

nfx:
Also change the PR title to be changelog-friendly

src/databricks/labs/ucx/framework/tasks.py
larsgeorge-db: The latter will not help with the steps ordering, only task ordering. I am open for suggestions on how to do this better.

src/databricks/labs/ucx/install.py
larsgeorge-db: Understood, will address asap.

tests/unit/test_install.py
larsgeorge-db: Darn, I need to find a way to make this testable... in general, using protected fields seems to be complained about but some linters.

src/databricks/labs/ucx/framework/tasks.py
larsgeorge-db: How about an extra field that holds a generated ID, which is starting at 1 and gets increased for every task added. Then we can sort on that field? @nfx

nfx: > How about an extra field that holds a generated ID, which is starting at 1 and gets increased for every task added. Then we can sort on that field? 

that's exactly what I'm asking for :) but start with 0, to be consistent with array indexing

src/databricks/labs/ucx/framework/tasks.py
```diff
@@ -53,7 +55,11 @@ def wrapper(*args, **kwargs):
             msg = f"Task {func.__name__} must have documentation"
             raise SyntaxError(msg)
 
+        global _TASK_ID_GENERATOR
```
larsgeorge-db, line 58: @nfx The linter does not like this:
```
src/databricks/labs/ucx/framework/tasks.py:58:9: PLW0603 Using the global statement to update `_TASK_ID_GENERATOR` is discouraged
```
Is there a better way?

nfx: You can len(_TASKS)

nfx: no additional counter to track, right?

src/databricks/labs/ucx/install.py
```diff
@@ -248,21 +248,35 @@ def _create_jobs(self):
         self._create_readme()
         self._create_debug(remote_wheel)
 
+    @staticmethod
+    def _sort_task_list() -> list[Task]:
```
nfx, line 252:
```suggestion
    def _sorted_tasks() -> list[Task]:
```

`_sort` implies no retun

src/databricks/labs/ucx/install.py
```diff
@@ -248,21 +248,35 @@ def _create_jobs(self):
...
+        return sorted(_TASKS.values(), key=lambda x: x.task_id)
+
+    @classmethod
+    def _step_list(cls) -> list[str]:
+        step_list = []
+        [step_list.append(task.workflow) for task in cls._sort_task_list() if task.workflow not in step_list]
```
nfx, line 258: can you rewrite it as a for-loop, as you're not assigning a result of list comprehension to a value

nfx (changes requested):

src/databricks/labs/ucx/install.py
larsgeorge-db: Does it? Seems arbitrary.

nfx (approved):
LGTM for implementation, but fix the tests

Issue #288, opened 2023-09-25
Ensure fixed casing for object names in inventory tables.


larsgeorge-db:
Names like catalog, databases and tables etc. should be uppercased in `tables.scala`. This makes it easier to process them downstream in the jobs.

nfx:
@larsgeorge-db maybe they should be lowercased?

pohlposition:
This is 2 weeks old.

Is this something we need to do?  Do we have naming conventions for things like this?

larsgeorge-db:
I do not think we do... I am actually looking at this and was asking myself the same question. Which column should be converted to what?

| col_name | data_type | casing |
| --------- | --------| ------- |
| `catalog` | `string` | lower |
| `database` | `string` | lower |
| `name` | `string` | lower |
| `object_type` | `string` | upper |
| `table_format` | `string` | upper |
| `location` | `string` | leave |
| `view_text` | `string` | leave |
| `upgraded_to` | `string` | leave |

Looking at what we have so far, `object_type` is upper already, so matching `table_format` to it makes sense. Given SQL objects are treated case-insensitive by DBSQL we could lower case the names, making them easier on the eyes mostly. But, this does not apply to all RDBMSs, some are case-sensitive, like Oracle, so names of federated tables are likely sensitive too. Assuming we deal with the HMS side here, even a federated table would be in the connection/location details, so the case-insensitivity still applies here.

larsgeorge-db:
Looking at our dashboard queries, the only column we have to `UPPER()` is `table_format`. In the code we `LOWER()` the database and table names at a few places. Do we expect any impact on applying the above conversions directly while scanning HMS and taking stock of its data structures?

HariGS-DB:
@nfx Have started working on it today. Had one question to confirm.
 - The requirement is to convert all fields of all tables created by ucx to lower case? and ensure read operation dont do any upper anywhere like in dashboard queries. 
 - only the tables used by tables.scala (tables and table_failures)?

Issue #249, opened 2023-09-20
Create inventory of service principals and direct files access in Azure


birbalgithub:
In Azure, data access is authorized using service principals and it can be configured in either of the settings below- 
Cluster conf
Spark session
Cluster policy
DLT settings

Today, there is no automated way of knowing how many service principal are being used in the workspace and where they are being set (cluster/cluster policies/dlt/notebook ?). As such, it is hard for the customer is figure out how many STORAGE CREDENTIALS needs to be created as part of UC upgrade. Furthermore, they could be accessing the files or mount points directly using service principal credentials and as such it is important to list all those files and mount points to estimates the EXTERNAL LOCATIONS that needs to be created. Additionally, it is also useful to list the cluster/cluster policies/dlt/notebook wherever the service principal credentials are being set to identify their owner/user. This would help  customer map the permission to the required group on Storage Credential and External locations.

We need a feature in the tool to create an inventory of all service principals and direct files/mount points that are currently being used in the workspace along with the objects (cluster/cluster policies/dlt/notebook) that are using them .  The direct files/mount points can be found by scanning their code (Notebook/scala/java/sql) and the service principal can be found by scanning the following four settings.

1. Cluster conf settings
spark.hadoop.fs.azure.account.auth.type OAuth
spark.hadoop.fs.azure.account.oauth.provider.type org.apache.hadoop.fs.azurebfs.oauth2.ClientCredsTokenProvider
spark.hadoop.fs.azure.account.oauth2.client.id <client-id>
spark.hadoop.fs.azure.account.oauth2.client.secret {{secrets/<secret-scope>/<secret-name>}}
spark.hadoop.fs.azure.account.oauth2.client.endpoint https://login.microsoftonline.com/<tenant-id>/oauth2/token

2. Spark session level settings
spark.conf.set("fs.azure.account.auth.type","OAuth")
spark.conf.set("fs.azure.account.oauth.provider.type","org.apache.hadoop.fs.azurebfs.oauth2.ClientCredsTokenProvider")
spark.conf.set("fs.azure.account.oauth2.client.id","<client-id>")
spark.conf.set("fs.azure.account.oauth2.client.secret",dbutils.secrets.get(scope="<secret-scope>",key="<service-credential-key>"))
spark.conf.set("fs.azure.account.oauth2.client.endpoint","https://login.microsoftonline.com/<tenant-id>/oauth2/token")

3. DLT settings
"configuration": {
[...]

PR #299, opened 2023-09-26
Detect conflicts between workspaces in the account for groups and tables

### LINKED ISSUE #22: Add ability to compare workspace and account group size and members

### LINKED ISSUE #83: Groups with same name across workspaces
For the group migration when we migrate the groups at the account level, how do we handle that scenario where two different workspaces has groups with the exact same names.

nfx:
Fixes #83
Fixes #22

### DIFF
--- a/src/databricks/labs/ucx/account/conflicts.py
+++ b/src/databricks/labs/ucx/account/conflicts.py
@@ -1,0 +1,127 @@
+import collections
+import logging
+
+from databricks.labs.ucx.account.workspaces import Workspaces
+from databricks.labs.ucx.config import WorkspaceConfig
+from databricks.labs.ucx.framework.crawlers import StatementExecutionBackend
+from databricks.labs.ucx.hive_metastore import TablesCrawler
+from databricks.labs.ucx.workspace_access.groups import GroupCrawler
+
+logger = logging.getLogger(__name__)
+
+
+def groups(workspaces: Workspaces):
+    workspace_members = {}
+    workspace_names = {}
+    for workspace in workspaces.configured_workspaces():
+        workspace_names[workspace.workspace_id] = workspace.workspace_name
+        ws = workspaces.client_for(workspace)
+        group_crawler = GroupCrawler(ws)
+        logger.debug(f"scanning users from {workspace.workspace_name}")
+        workspace_members[workspace.workspace_id] = group_crawler.membership()
+    seen = set()
+    no_conflicts = set()
+    for workspace_id, membership in workspace_members.items():
+        for group_name, usernames in membership.items():
+            if group_name in seen:
+                continue
+            diffs = {}
+            # this is O((n^3)/2), but maybe there's a better algo
+            for other_workspace_id in workspace_members.keys():
+                if workspace_id == other_workspace_id:
+                    continue  # self-link
+                other_usernames = workspace_members[other_workspace_id][group_name]
+                diff_a = usernames.difference(other_usernames)
+                diff_b = other_usernames.difference(usernames)
+                # record membership differences between workspace groups
+                diffs[other_workspace_id] = (diff_a, diff_b)
+            no_diff_counter = 0
+            for other_workspace_id, (diff_a, diff_b) in diffs.items():
+                if len(diff_a) == 0 and len(diff_b) == 0:
+                    no_diff_counter += 1
+                    continue
+                workspace_name = workspace_names[workspace_id]
+                other_workspace_name = workspace_names[other_workspace_id]
+                msg = []
+                if diff_a:
+                    msg.append(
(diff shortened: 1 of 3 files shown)


nfx:
Depends on https://github.com/databrickslabs/ucx/issues/288, https://github.com/databrickslabs/ucx/issues/249,

FastLee:
Should we consider a "warehouse" for assessment/migration? Pulling from all workspaces?

nfx:
@FastLee so how warehouse should work across 1000 workspaces? Do we want to select a warehouse from each of those?.. what if warehouses have different names or not existing? 

We also may need to create a "main workspace" for migration, but i am not sure if it's simpler than storing configuration on a laptop

zpappa:
Customers will not love the idea of creating a workspace to run a tool. 

We should follow simple and predictable logic when trying to reconcile workspace local groups with account groups, and document this so people know how it works.

Here are the scenarios that come to mind.

**Scenario 1**
Workspace Local SCIM is configured
Workspace Local Group Is an External Group
Workspace Local Group Does not exist in the account console
Result: Migrate workspace local group, notify user that this group needs to be added to account console SCIM connector

**Scenario 2**
Workspace Local SCIM is configured
Workspace Local Group is an External Group
Workspace Local Group exists in the account console and is an external group
Result: Do not migrate workspace local group, but use the account group that already exists

**Scenario 3**
Workspace Local SCIM is not configured
Workspace local group is not an external group
Workspace Local group does not exist in the account console
Result: Create the workspace local group in the account console with the same members

**Scenario 4**
Workspace Local SCIM is not configured
Workspace local group is not an external group
Workspace Local group exists in the account console and does not have the same members
Result: Create a new group with the workspace name as a prefix, and create in the account console

**Scenario 5**
Workspace Local SCIM is not configured
Workspace local group is not an external group
Workspace Local group exists in the account console and has the same members
Result: Use the existing group in the account console

nfx:
We have no way of knowing if SCIM sync is configured, because it's the external tool calling our APIs. Scim sync is something done by identity admins

nfx:
> upgraded all their workspaces at once
The requirement is to run assessment job to collect metadata about HMS and flag inconveniences/conflicts. Then we collect accout/local groups and flag inconveniences/conflicts. It's an easier single ticket for Identity Engineering (or Administrative) team about making changes to Okta/Aad/Terraform for user management (i'd say, majority of UC target audience).

> (identity admins) Create the workspace local group in the account console with the same members
I think that the single ticket to that team has to be as simple as "install Databricks CLI and run the 'databricks labs ucx report-groups' and modify okta/aad/tf to change groups according to output".

We need an issue for this scenario, please create.

> (identity admins) Create the workspace local group in the account console with the same members
We need an issue for this scenario, please create. More than that - it's exactly the same scenario as "Create a new group with the workspace name as a prefix, and create in the account console", but without the need to create a prefix in most of the cases.

> Use the existing group in the account console

**Migration probably would happen in workspace-by-workspace job-in-UI triggering. We can automate that also, if customers asks/votes.**

That's for groups.

-------

We didn't cover service principals at all 😉 on aws, account-level and service level spns may have the same/purpose, but different application_id, that is recorded in both grants and generic permissions.

--------

Now for tables, there also needs to be a report on table/db inconsistency - like 
A: db1.tbl1, db1.tbl3
B: db1.tbl2

And the team(s) that are driving UC Migration within account would make a decision after some time in review (of excel spreadsheet). By the way, we can split UCX installation across different Azure Subscriptions. And every installation would just focus on defining target catalog mapping per database. But here are unanswered questions:

- two workspaces, same dbs, all different tables and columns (all managed tables, effectively)
- two workspaces, same dbs, 90% same tables, 10% are different tables 
- two workspaces, two different dbs

We can technically support both db_to_catalog and workspace_to_catalog, and even at the same time, but db_to_catalog will override workspace_to_catalog. We also need default_catalog_for_workspace, if workspace_to_catalog is set (default catalog for all workspaces is set per metastore)..

We can also do another override for tables, but we have unanswered questions:

- what if same db, same workspace, same table, but different columns/order/types? Ignore and keep in hive metastore? And then rerun the scan for tables and grants?
- what if during migration catalog/database/table were deleted either from hms and/or uc?

Speaking of metastores, in the beginning, there needs to be workspace_to_metastore mapping with default_metastore_for_workspace. Can we come up with a good default  mapping here? Coarse or fine grained? Select between the two? Ask for inline input? How many conflicts we expect to justify the need to create/support custom mapping?

the last very important question is what future-proof configuration format might we need for this mapping. That's why I don't want any configuration after the assessment step.

nfx:
Workspace to metastore mapping can be solved with region:

> We should ideally be checking if there is a metastore in region, and in region only to the workspace, not in general if there is one setup.

nfx:
stale PR, will need to start from scratch

PR #306, opened 2023-09-27
Added a column to `$inventory.tables` to specify if a table might have been synchronised to Unity Catalog already or not

### LINKED ISSUE #303: Add upgrade status and upgrade target to the "tables" assessment table
As part of the assessment process.
Add the following columns to the "tables" assessment table.
upgrade_status | int | Upgrade Status, 1 for upgraded 0 for not | default:0
upgrade_target | string | Target Table for (3 level) of upgrade | default:""

Set upgrade_status to 1 if upgraded_to property is set on the table properties.

FastLee:
Closes #303


src/databricks/labs/ucx/assessment/queries/all_tables.sql
```diff
@@ -12,5 +12,7 @@ SELECT `database`,
            ELSE "EXTERNAL"
        END AS storage,
        IF(format = "delta", "Yes", "No") AS is_delta,
-       location
+       location,
+       upgrade_status,
+       upgrade_target
```
nfx, line 17: we cannot have these columns here. this table is overwritten every assessment run and we lose all the state: `df.write.mode("overwrite")`. The only proper way to store it is in the table properties of the source and target table.

nfx (changes requested):

src/databricks/labs/ucx/assessment/queries/all_tables.sql
nfx: It doesn't belong to this task/workflow, which is run on a workspace without UC metastore. What are you trying to show?

src/databricks/labs/ucx/assessment/queries/database_summary.sql
```diff
@@ -1,11 +1,13 @@
--- viz type=table, name=Database Summary, columns=database,tables,views,dbfs_root,delta_tables,upgrade
+-- viz type=table, name=Database Summary, columns=database,tables,views,dbfs_root,delta_tables,upgraded_status,upgrade
 -- widget title=Database Summary, col=0, row=9, size_x=6, size_y=8
 SELECT `database`,
        SUM(is_table) AS tables,
        SUM(is_view) AS views,
        SUM(is_dbfs_root) AS dbfs_root,
        SUM(is_delta) AS delta_tables,
+       SUM(upgrade_status) AS upgraded_tables,
```
nfx, line 8: This is not part of the assessment. Upgrade has to be a different dashboard

nfx (changes requested):

src/databricks/labs/ucx/framework/dashboards.py
```diff
@@ -130,7 +130,7 @@ def _install_widget(self, query: SimpleQuery):
     def _get_widget_options(self, query: SimpleQuery):
         self._pos += 1
         widget_options = WidgetOptions(
-            title=query.widget.get("title", ""),
+            title=query.widget["title"],
```
nfx, line 133: will break if title is not there

nfx (changes requested):
Please rebase on main - you're deleting features.

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -22,6 +22,8 @@ class Table:
 
     location: str = None
     view_text: str = None
+    upgrade_status: int = 0
```
nfx, line 25:
```suggestion
```
we don't need this column

src/databricks/labs/ucx/hive_metastore/tables.scala
```diff
@@ -36,8 +36,10 @@ def metadataForAllTables(databases: Seq[String], queue: ConcurrentLinkedQueue[Ta
           failures.add(TableError("hive_metastore", databaseName, tableName, s"result is null"))
           None
         } else {
+          val upgrade_to=table.properties.get("upgraded_to")
           Some(TableDetails("hive_metastore", databaseName, tableName, table.tableType.name, table.provider.orNull,
-            table.storage.locationUri.map(_.toString).orNull, table.viewText.orNull))
+            table.storage.locationUri.map(_.toString).orNull, table.viewText.orNull,
+              upgrade_to match {case Some(target) => 1 case None => 0}, upgrade_to match {case Some(target) => target case None => ""}))
```
nfx, line 42: remove the upgrade status column, we can infer it from `upgrade_target`. Please answer what would happen if the table is removed from the target catalog.


also add the same parsing to `_crawl` method in the `TablesCrawler` python's implementation, so that we can properly run the integration tests.

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -70,6 +73,13 @@ def sql_alter_from(self, catalog):
         )
 
 
+def parse_table_props(tbl_props: string) -> {}:
```
nfx, line 76: don't let this public function dangling on a module level, it has to be a private method in the `TablesCrawler`. also add a unit test, so that we know when parsing of table properties is broken.

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -140,6 +150,7 @@ def _describe(self, catalog: str, database: str, table: str) -> Table | None:
                 table_format=describe.get("Provider", "").upper(),
                 location=describe.get("Location", None),
                 view_text=describe.get("View Text", None),
+                upgraded_to=parse_table_props(describe.get("Table Properties", "")).get("upgraded_to", ""),
```
nfx, line 153: add unit test. check cases when props are broken

src/databricks/labs/ucx/hive_metastore/tables.scala
```diff
@@ -36,8 +36,10 @@ def metadataForAllTables(databases: Seq[String], queue: ConcurrentLinkedQueue[Ta
           failures.add(TableError("hive_metastore", databaseName, tableName, s"result is null"))
           None
         } else {
+          val upgraded_to=table.properties.get("upgraded_to")
           Some(TableDetails("hive_metastore", databaseName, tableName, table.tableType.name, table.provider.orNull,
-            table.storage.locationUri.map(_.toString).orNull, table.viewText.orNull))
+            table.storage.locationUri.map(_.toString).orNull, table.viewText.orNull,
+              upgraded_to match {case Some(target) => target case None => ""}))
```
nfx, line 42: has to be NULL, not empty string.

nfx (changes requested):

nfx (approved):

nfx (approved):

Issue #297, opened 2023-09-25
assess_jobs fails with “PySparkValueError: [CANNOT_BE_NONE] Argument `obj` can not be None.”


tamilselvanveeramani:
in Azure demo workspace assess_jobs fails with “PySparkValueError: [CANNOT_BE_NONE] Argument `obj` can not be None.”

Here is the error traces

PySparkValueError: [CANNOT_BE_NONE] Argument `obj` can not be None.
---------------------------------------------------------------------------
PySparkValueError                         Traceback (most recent call last)
File ~/.ipykernel/1386/command--1-1316133125:18
     15 entry = [ep for ep in metadata.distribution("databricks_labs_ucx").entry_points if ep.name == "runtime"]
     16 if entry:
     17   # Load and execute the entrypoint, assumes no parameters
---> 18   entry[0].load()()
     19 else:
     20   import databricks_labs_ucx

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/runtime.py:211, in main()
    210 def main():
--> 211     trigger(*sys.argv)

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/framework/tasks.py:91, in trigger(*argv)
     88 cfg = WorkspaceConfig.from_file(Path(args["config"]))
     89 logging.getLogger("databricks").setLevel(cfg.log_level)
---> 91 current_task.fn(cfg)

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/runtime.py:104, in assess_jobs(cfg)
    102 ws = WorkspaceClient(config=cfg.to_databricks_config())
    103 crawler = JobsCrawler(ws, RuntimeBackend(), cfg.inventory_database)
--> 104 crawler.snapshot()

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/assessment/crawlers.py:146, in JobsCrawler.snapshot(self)
    145 def snapshot(self) -> list[ClusterInfo]:
--> 146     return self._snapshot(self._try_fetch, self._crawl)

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/framework/crawlers.py:215, in CrawlerBase._snapshot(self, fetcher, loader)
[...]

PR #356, opened 2023-10-02
Fixed appending to tables by adding filtering of `None` rows

william-conti:
This PR aims to fix https://github.com/databrickslabs/ucx/issues/297 and https://github.com/databrickslabs/ucx/issues/346

It adds a utility method to filter rows that have a column containing None, this will help Crawlers to not throw an error when a column is None.

it also checks if the column in the class is Nullable or not, if it's nullable and the value is None, it's ignored

### DIFF
--- a/src/databricks/labs/ucx/framework/crawlers.py
+++ b/src/databricks/labs/ucx/framework/crawlers.py
@@ -48,13 +48,12 @@ class SqlBackend(ABC):
             row_contains_none = False
             for k, v in dataclasses.asdict(row).items():
                 if v is None:
-                    print(f"Field {k} is None, filtering")
+                    logger.debug(f"Field {k} is None, filtering row")
                     row_contains_none = True
                     break
             if not row_contains_none:
                 results.append(row)
         return results
-
 
 
 class StatementExecutionBackend(SqlBackend):
--- a/src/databricks/labs/ucx/framework/crawlers.py
+++ b/src/databricks/labs/ucx/framework/crawlers.py
@@ -44,13 +44,20 @@ class SqlBackend(ABC):
     @classmethod
     def _filter_none_rows(cls, rows):
         results = []
+        nullable_fields = []
+
+        for field in dataclasses.fields(rows[0]):
+            if field.default is None:
+                nullable_fields.append(field.name)
+
         for row in rows:
             row_contains_none = False
-            for k, v in dataclasses.asdict(row).items():
-                if v is None:
-                    logger.debug(f"Field {k} is None, filtering row")
+            for column, value in dataclasses.asdict(row).items():
+                if value is None and column not in nullable_fields:
+                    logger.debug(f"Field {column} is None, filtering row")
                     row_contains_none = True
                     break
+
             if not row_contains_none:
                 results.append(row)
         return results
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -41,6 +41,27 @@ def _schema_for(cls, klass):
...
+    @classmethod
+    def _filter_none_rows(cls, rows):
+        results = []
+        nullable_fields = []
+
+        for field in dataclasses.fields(rows[0]):
+            if field.default is None:
+                nullable_fields.append(field.name)
+
+        for row in rows:
```
nfx, line 53: Also add `if row is None: continue` - might be enough to fix the issue, but it's good to be more defensive

src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -41,6 +41,27 @@ def _schema_for(cls, klass):
...
+            row_contains_none = False
+            for column, value in dataclasses.asdict(row).items():
+                if value is None and column not in nullable_fields:
+                    logger.debug(f"Field {column} is None, filtering row")
```
nfx, line 57:
```suggestion
                    logger.warning(f"[{full_table_name}] Field {column} is None, filtering row")
```

also propagate the `full_table_name` down here, so that the logs are more clear.

src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -41,6 +41,27 @@ def _schema_for(cls, klass):
             fields.append(f"{f.name} {spark_type}{not_null}")
         return ", ".join(fields)
 
+    @classmethod
+    def _filter_none_rows(cls, rows):
+        results = []
+        nullable_fields = []
```
nfx, line 47:
```suggestion
        nullable_fields = set()
```
so that the lookup is faster for 10k rows

src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -115,6 +136,8 @@ def fetch(self, sql) -> Iterator[any]:
         return self._spark.sql(sql).collect()
 
     def save_table(self, full_name: str, rows: list[any], mode: str = "append"):
+        rows = self._filter_none_rows(rows)
```
nfx, line 139: can you also add that to `StatementExecutionBackend`?

tests/unit/framework/test_crawlers.py
```diff
@@ -204,3 +210,39 @@ def test_runtime_backend_save_table(mocker):
...
+def test_runtime_backend_save_table_with_row_containing_none(mocker):
+    from unittest import mock
+
+    with mock.patch.dict(os.environ, {"DATABRICKS_RUNTIME_VERSION": "14.0"}):
+        pyspark_sql_session = mocker.Mock()
+        sys.modules["pyspark.sql.session"] = pyspark_sql_session
+
+        rb = RuntimeBackend()
+
+        rb.save_table("a.b.c", [Foo("aaa", True), Foo("bbb", False), Foo("bbb", None)])
```
nfx, line 224:
```suggestion
        rb.save_table("a.b.c", [Foo("aaa", True), Foo("bbb", False), Foo("ccc", None)])
```
so that it's more clear from reading the test

PR #357, opened 2023-10-02
Remove exception and added proper logging for groups in the list that…

### LINKED ISSUE #312: Groups migration fails if a local group is not found on the account level, when we explicitly select that group in config
I run the tool with selected group. One of the group did not exist in the account. The tool failed and stopped.

Enclosed stacktrace and standard output.
[stacktrace.txt](https://github.com/databrickslabs/ucx/files/12746443/stacktrace.txt)

### LINKED ISSUE #313: Groups migration fails if the selected group does not exist
If the groups selected in the config do not exist the groups-migration job fails. 

```
selected:
- non_existent
```

FastLee:
Closes #312 
Closes #313

### DIFF
--- a/src/databricks/labs/ucx/workspace_access/groups.py
+++ b/src/databricks/labs/ucx/workspace_access/groups.py
@@ -161,25 +161,24 @@ class GroupManager:
             "exist and are of the correct type. If some temporary groups are missing, they'll be created"
         )
         group_names = self.config.selected
+        valid_group_names = []
         if group_names:
             logger.info("Using the provided group listing")
 
             for g in group_names:
                 if g in self.SYSTEM_GROUPS:
                     logger.info(f"Cannot migrate system group {g}. {g} will be skipped.")
-                    group_names.remove(g)
                     continue
                 if not self._get_group(g, "workspace"):
                     logger.info(f"Group {g} not found on the workspace level. {g} will be skipped.")
-                    group_names.remove(g)
                     continue
                 if not self._get_group(g, "account"):
                     logger.info(
                         f"Group {g} not found on the account level. {g} will be skipped. You can add {g} "
                         f"to the account and rerun the job."
                     )
-                    group_names.remove(g)
                     continue
+                valid_group_names.append(g)
         else:
             logger.info(
                 "No group listing provided, all available workspace-level groups that have an account-level "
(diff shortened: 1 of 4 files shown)


src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -165,11 +165,22 @@ def prepare_groups_in_environment(self):
...
 
             for g in group_names:
-                assert g not in self.SYSTEM_GROUPS, f"Cannot migrate system group {g}"
-                assert self._get_group(g, "workspace"), f"Group {g} not found on the workspace level"
-                assert self._get_group(g, "account"), f"Group {g} not found on the account level"
-
-        if not group_names:
+                if g in self.SYSTEM_GROUPS:
+                    logger.info(f"Cannot migrate system group {g}. {g} will be skipped.")
+                    group_names.remove(g)
```
nfx, line 170: don't modify the list while iterating, as it's one of the most common sources for bugs. e.g. don't use `group_names.remove`, it doesn't give deterministic results:

```
>>> x = [1,2,3,4,5]
>>> for i in x:
...   if i > 3:
...     x.remove(i)
... 
>>> x
[1, 2, 3, 5] <=== 5 should have been removed

>>> for i in x: x.remove(i)
... 
>>> x
[2, 4] <==== should have been empty
```

instead, create a copy of the list, but with invalid elements removed.

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -165,11 +165,22 @@ def prepare_groups_in_environment(self):
...
+                    continue
+                if not self._get_group(g, "workspace"):
+                    logger.info(f"Group {g} not found on the workspace level. {g} will be skipped.")
+                    group_names.remove(g)
```
nfx, line 174: can you add a test to make sure that when the groups are either not found or invalid - they don't get to a selected list?

nfx (changes requested):
don't modify the list while iterating, as it's one of the most common sources for bugs

src/databricks/labs/ucx/workspace_access/groups.py
FastLee: Done

nfx (approved):
👍

PR #366, opened 2023-10-03
Crawler for RunSubmit API usages from External Orchestrators (ADF/Airflow)

### LINKED ISSUE #266: Assessment for `RunSubmit` API usages
**Problem statement**

Existing processes that create `RunSubmit` Jobs (the so-called ephemeral jobs) with the following properties:

- DBR > 11.X
- [AWS-specific] Uses `instance_profile` 
- Has no `data_security_mode` property specified in the job creation request
- [Azure-specific] some configurations in the `spark_conf` (To be confirmed)

May be broken when UC is enabled in a given workspace (meaning when workspace is assigned to the metastore). 

**Why this happens?**

When UC is enabled, all `RunSubmit` Jobs with DBR 11.X+ will by default UC-enabled. If there is a conflict in permissions between UC and service principal (e.g. instance profile), the job will fail. 

The reason and the change is described [here](https://docs.databricks.com/en/archive/runtime-release-notes/11.1.html#job-clusters-default-to-single-user-access-mode-with-databricks-runtime-111-and-higher). 

**How can we identify the identical `RunSubmit`?**

This requires additional internal discussion.

# TODO:

- [ ] get a list of all (persisted) jobs
- [ ] get a list of all job runs
- [ ] find job runs that have no persisted job (out of workflow)
- [ ] group all job runs from non-persisted jobs to approximate the number of unique airflow/azure data factory DAGs.
- [ ] identify job runs that do not include data_security_mode in the job creation request and are run against 11.x  compute

zpappa:
Resolves #266 

**Added JobRunsCrawler**
Added a crawler to look at JobRuns from the SDK and determine which of the job runs are from the RunsSubmit API

**Added Unit Tests**
Added tests to cover basic logic, further tests pending.


larsgeorge-db (approved):
lgtm

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -249,3 +259,47 @@ def snapshot(self) -> list[ClusterInfo]:
...
             yield JobInfo(*row)
+
+
+class JobsRunCrawler(CrawlerBase):
+    def __init__(self, ws: WorkspaceClient, sbe: SqlBackend, schema):
+        super().__init__(sbe, "hive_metastore", schema, "job_runs")
+        self._ws = ws
+
+    def _crawl(self) -> list[JobRunInfo]:
+        no_of_days_back = datetime.timedelta(days=30)  # todo make configurable in yaml?
```
nfx, line 270:
```suggestion
        no_of_days_back = datetime.timedelta(days=90)  # todo make configurable in yaml?
```

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -249,3 +259,47 @@ def snapshot(self) -> list[ClusterInfo]:
...
+        start_time_from = datetime.datetime.now() - no_of_days_back
+        # todo figure out if we need to specify a default timezone
+        all_job_runs = list(
+            self._ws.jobs.list_runs(start_time_from=start_time_from, start_time_to=datetime.datetime.now())
```
nfx, line 274: two questions:
1. why do we need to `list(self._ws.jobs.list_runs(..))`? SDK returns the iterator and here we load everything to memory immediately.
2. do we need to specify the `start_time_to` param?

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -249,3 +259,47 @@ def snapshot(self) -> list[ClusterInfo]:
...
+        return self._assess_job_runs(all_clusters, all_job_runs)
+
+    def _assess_job_runs(self, all_clusters, all_job_runs):
+        all_job_run_info = []
+        job_runs_without_job_id = list(filter(lambda jr: jr.job_id is None, all_job_runs))
+        for job_run in job_runs_without_job_id:
+            for job_run_cluster in job_run.job_clusters:
+                cluster_key = job_run_cluster.job_cluster_key
+                cluster: ClusterDetails = (
+                    all_clusters[cluster_key] if cluster_key is not None else job_run_cluster.new_cluster
```
nfx, line 286: a) the logic is wrong - `cluster_key` is an alias, that is valid in scope of a job, not an id. `existing_cluster_id ` is something that has real cluster id.
b) there's a logic for getting cluster confs - for example, in the job assessment. Why don't you reuse that? :) 

image

https://docs.databricks.com/api/workspace/jobs/create

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -249,3 +259,47 @@ def snapshot(self) -> list[ClusterInfo]:
...
+                )
+                spark_version = cluster.spark_version
+                data_security_mode = cluster.data_security_mode
+                job_run_info = JobRunInfo(
+                    run_id=job_run.run_id,
+                    run_type=str(job_run.run_type.value),
```
nfx, line 292: why do we need to `str()` it? our `SqlBackend` framework picks up enums automatically and converts them to strings. but why do we need to pull this run_type at all?..

nfx (changes requested):
missing logic for hashing/deduping the similar run-submit jobs and has some logic errors

src/databricks/labs/ucx/assessment/crawlers.py
zpappa: I believe this has been changed and is even more complete, but I am not using the code in the job assessment because it's in an internal method, do you have a pattern for reusing code you want us to follow? Util class etc.?

src/databricks/labs/ucx/assessment/crawlers.py
zpappa: Got it - wasn't sure this was handled, defensive coding...

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -250,3 +269,150 @@ def snapshot(self) -> list[ClusterInfo]:
     def _try_fetch(self) -> list[ClusterInfo]:
         for row in self._fetch(f"SELECT * FROM {self._schema}.{self._table}"):
             yield JobInfo(*row)
+
+
+class ExternallyOrchestratedJobTaskCrawler(CrawlerBase):
+    def __init__(self, ws: WorkspaceClient, sbe: SqlBackend, schema):
+        super().__init__(sbe, "hive_metastore", schema, "job_runs", ExternallyOrchestratedJobTask)
```
nfx, line 276: `job_runs` may be too generic here for a table name

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -250,3 +269,150 @@ def snapshot(self) -> list[ClusterInfo]:
...
+        if task.python_wheel_task is not None:
+            hash_values.append(task.python_wheel_task.package_name)
+            hash_values.append(task.python_wheel_task.entry_point)
+        if task.sql_task is not None:
+            hash_values.append(task.sql_task.file.path)
+            hash_values.append(task.sql_task.alert.alert_id)
+            hash_values.append(task.sql_task.dashboard.dashboard_id)
+            hash_values.append(task.sql_task.query.query_id)
+        if task.dbt_task is not None:
+            task.dbt_task.commands.sort()
```
nfx, line 322: is it really needed here?

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -250,3 +269,150 @@ def snapshot(self) -> list[ClusterInfo]:
...
+            hash_values.append(task.dbt_task.schema)
+            hash_values.append(task.dbt_task.catalog)
+            hash_values.append(task.dbt_task.warehouse_id)
+            hash_values.append(task.dbt_task.project_directory)
+            hash_values.append(",".join(task.dbt_task.commands))
```
nfx, line 327:
```suggestion
            hash_values.append(",".join(sorted(task.dbt_task.commands)))
```
this might be a bit more deterministic

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -250,3 +269,150 @@ def snapshot(self) -> list[ClusterInfo]:
...
+            hash_values.append(task.condition_task.right)
+            hash_values.append(task.condition_task.left)
+            hash_values.append(task.condition_task.outcome)
+        if task.run_job_task is not None:
+            hash_values.append(task.run_job_task.job_id)
+        if task.git_source is not None:
+            hash_values.append(task.git_source.git_url)
+            hash_values.append(task.git_source.git_tag)
+            hash_values.append(task.git_source.git_branch)
+            hash_values.append(task.git_source.git_commit)
```
nfx, line 339:
```suggestion
```
these might vary a lot, no need to have them in hash

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -250,3 +269,150 @@ def snapshot(self) -> list[ClusterInfo]:
...
+            hash_values.append(task.git_source.git_provider)
+            hash_values.append(task.git_source.git_snapshot.used_commit)
```
nfx, line 341:
```suggestion
```

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -250,3 +269,150 @@ def snapshot(self) -> list[ClusterInfo]:
...
+
+        hash_value_string = "|".join([value for value in hash_values if value is not None])
```
nfx, line 343:
```suggestion
        hash_value_string = "|".join([str(value) for value in hash_values if value is not None])
```
as some of the things may not be strings. python is a language that allows sloppy code and we have to be very defensive against the language itself. probably surround it with `try .. except` to be extra careful :D

nfx (changes requested):
this starts to take shape. minor comments remaining - please also add the integration for this, so that we have some happy path calling the real apis.

nfx (changes requested):

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -250,3 +269,150 @@ def snapshot(self) -> list[ClusterInfo]:
...
+        ext_orc_job_tasks: list[ExternallyOrchestratedJobTask] = list[ExternallyOrchestratedJobTask]()
+        all_persisted_job_ids = [x.job_id for x in all_jobs]
+        not_persisted_job_runs = list(
+            filter(lambda jr: jr.job_id is None or jr.job_id not in all_persisted_job_ids, all_job_runs)
+        )
+        for job_run in not_persisted_job_runs:
+            for task in job_run.tasks:
+                spark_version = self._get_spark_version_from_task(task, job_run, all_clusters)
+                data_security_mode = self._get_data_security_mode_from_task(task, job_run, all_clusters)
+                hashed_id = self._create_hash_from_job_run_task(task)
```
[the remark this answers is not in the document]
zpappa: I figured we can do dedupe in the actual report based on the hash.

Airflow creates job runs yes, but each task is triggered in a workflow, so a job run might exist that only has one task out of 5 that ran successfully, also each task can have it's own cluster configuration, and that's what we're checking.

src/databricks/labs/ucx/assessment/crawlers.py
zpappa: Shouldn't the SDK be enforcing typing?

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -250,3 +269,150 @@ def snapshot(self) -> list[ClusterInfo]:
...
+        pipe separate and hash the value to provide a uniqueness identifier
+        - specifically ignore parameters as these change.
+        """
+        hash_values = []
+        if task.notebook_task is not None:
+            hash_values.append(task.notebook_task.notebook_path)
+        if task.spark_python_task is not None:
+            hash_values.append(task.spark_python_task.python_file)
+        if task.spark_submit_task is not None:
+            hash_values.append(task.spark_submit_task.parameters)
```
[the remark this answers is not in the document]
zpappa: 🤦 parameters shouldn't be in there

But python file should largely not change if it's the same DAG, unless they're doing some clever metadata driven selection

src/databricks/labs/ucx/assessment/crawlers.py
zpappa: Defensive posture :)

src/databricks/labs/ucx/assessment/crawlers.py
zpappa: okay, will opt for this instead

src/databricks/labs/ucx/assessment/crawlers.py
zpappa: Agreed

src/databricks/labs/ucx/assessment/crawlers.py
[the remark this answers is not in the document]
nfx: two things:
- this table has to get as little noise as possible, so we have to dedupe before we write to table.
- we have to do dedupe based on the job_run, not the task, because that's how multi-task jobs are designed.

nfx (changes requested):

src/databricks/labs/ucx/assessment/crawlers.py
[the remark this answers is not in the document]
zpappa: Yes, but cluster definitions live on the task as mentioned, trying to imagine a user running this..

Potentially we run the risk, in particularly large multi-task workflows, of directing the user to the wrong task to address.

If there are multiple tasks that would fail, then a user has to potentially change multiple cluster configurations. If we roll this up to one result, then we run the risk of telling the user to fix "1" thing.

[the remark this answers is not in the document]
nfx: For job with 50 tasks we have to have only one record in this table.

nfx (changes requested):
this also needs an integration test

src/databricks/labs/ucx/assessment/crawlers.py
zpappa: will rename it

src/databricks/labs/ucx/assessment/crawlers.py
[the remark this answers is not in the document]
zpappa: Will remove it, it shouldn't be there

src/databricks/labs/ucx/assessment/crawlers.py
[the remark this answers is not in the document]
zpappa: The challenge is around usability, if this shows up in failure JSON, returning a useful report from this is hard.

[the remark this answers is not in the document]
zpappa: Especially if there's 50 tasks ☝️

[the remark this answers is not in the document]
zpappa: I'll change the grain for now so we can move this along.

zpappa:
> this also needs an integration test

Haven't even bothered trying to set one up yet, the amount of data required is high to run this, will work on it...

src/databricks/labs/ucx/assessment/crawlers.py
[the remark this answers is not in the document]
zpappa: But we'll have to make a decision about which cluster version info we surface.

A multi task cluster can use multiple cluster definitions.

For now I'll just choose the lowest dbr, but this can be misleading.

nfx:
@zpappa just create few job runs with fixtures - e.g. calling the `ws.jobs.submit_run(...)`

zpappa:
Closing in favor of #395 after a rebase gone wrong.

Issue #374, opened 2023-10-04
Delete backup groups does not work if run as separate step


mwojtyczka:
migrate-groups-cleanup job does not delete any groups and just reports `Prepared 0 groups for migration`.

The `delete_backup_groups` func is using migration state which is empty if the groups were already migrated. This function only works if the func is run after the migration step. If the func is run as a separate job, the migration state has to be built again and since the groups at this point are migrated the migration state is empty.

mwojtyczka:
fixed as part of https://github.com/databrickslabs/ucx/pull/375

Issue #359, opened 2023-10-03
`crawl_permissions` run fails


mwojtyczka:
Crawl permissions task fails with the following error:

`DatabricksError: java.io.IOException: java.util.concurrent.TimeoutException: Timed out after 5 seconds
---------------------------------------------------------------------------
DatabricksError                           Traceback (most recent call last)
File ~/.ipykernel/1209/command--1-1024872113:18
     15 entry = [ep for ep in metadata.distribution("databricks_labs_ucx").entry_points if ep.name == "runtime"]
     16 if entry:
     17   # Load and execute the entrypoint, assumes no parameters
---> 18   entry[0].load()()
     19 else:
     20   import databricks_labs_ucx

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/runtime.py:213, in main()
    212 def main():
--> 213     trigger(*sys.argv)

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/framework/tasks.py:93, in trigger(*argv)
     90 cfg = WorkspaceConfig.from_file(Path(args["config"]))
     91 logging.getLogger("databricks").setLevel(cfg.log_level)
---> 93 current_task.fn(cfg)

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/runtime.py:137, in crawl_permissions(cfg)
    135 toolkit = GroupMigrationToolkit(cfg)
    136 toolkit.cleanup_inventory_table()
--> 137 toolkit.inventorize_permissions()

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/workspace_access/migration.py:124, in GroupMigrationToolkit.inventorize_permissions(self)
    123 def inventorize_permissions(self):
--> 124     self._permissions_manager.inventorize_permissions()

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/workspace_access/manager.py:26, in PermissionManager.inventorize_permissions(self)
     24 crawler_tasks = list(self._get_crawler_tasks())
     25 logger.info(f"Starting to crawl permissions. Total tasks: {len(crawler_tasks)}")
---> 26 results = ThreadedExecution.gather("crawl permissions", crawler_tasks)
[...]

Issue #353, opened 2023-10-02
Tool is failing creating new groups


mwojtyczka:
The tool is failing sometimes with the creation of groups. 

`DatabricksError: The service at /api/2.0/preview/scim/v2/Groups is taking too long to process your request. Please try again later or try a faster operation. [TraceId: 00-8bbf887439cb887433f92a1197f8bf36-4d79ab1b4df6ed88-00]
---------------------------------------------------------------------------
DatabricksError                           Traceback (most recent call last)
File ~/.ipykernel/1041/command--1-1857924431:18
     15 entry = [ep for ep in metadata.distribution("databricks_labs_ucx").entry_points if ep.name == "runtime"]
     16 if entry:
     17   # Load and execute the entrypoint, assumes no parameters
---> 18   entry[0].load()()
     19 else:
     20   import databricks_labs_ucx

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/runtime.py:213, in main()
    212 def main():
--> 213     trigger(*sys.argv)

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/framework/tasks.py:93, in trigger(*argv)
     90 cfg = WorkspaceConfig.from_file(Path(args["config"]))
     91 logging.getLogger("databricks").setLevel(cfg.log_level)
---> 93 current_task.fn(cfg)

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/runtime.py:192, in migrate_permissions(cfg)
    168 """As we embark on the complex journey of migrating from Hive Metastore to the Databricks Unity Catalog,
    169 a crucial phase in this transition involves the careful management of permissions.
    170 This intricate process entails several key steps: first, applying permissions to designated backup groups;
   (...)
    189 
    190 See [interactive tutorial here](https://app.getreprise.com/launch/myM3VNn/)."""
    191 toolkit = GroupMigrationToolkit(cfg)
--> 192 toolkit.prepare_environment()
    193 toolkit.apply_permissions_to_backup_groups()
    194 toolkit.replace_workspace_groups_with_account_groups()

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/workspace_access/migration.py:118, in GroupMigrationToolkit.prepare_environment(self)
[...]

PR #375, opened 2023-10-04
Improved cleanup for workspace backup groups by adding more retries on errors

mwojtyczka:
Fixed deletion of backup groups [issue #374].
Added rate limits and retries to group operations [issue #353].
Temp fix for issue #359
Added log messages for better visibility.
Added useful troubleshooting snippets to the docs.


src/databricks/labs/ucx/install.py
```diff
@@ -56,6 +56,36 @@
...
+]
+account_groups = [g for g in account_groups if g.display_name not in ["users", "admins", "account users"]]
+print(f"Found {{len(account_groups)}} account groups")
+
+# COMMAND ----------
+
+ws_group_names = {{_.display_name for _ in workspace_groups}}
+ac_group_names = {{_.display_name for _ in account_groups}}
+group_names = list(ws_group_names.intersection(ac_group_names))
+print(f"Found {{len(group_names)}} groups to migrate")
```
nfx, line 88: why do we need this in all debug nodebooks?

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -90,6 +91,8 @@ def _get_group(self, group_name, level: GroupLevel) -> iam.Group | None:
             if group.display_name == group_name:
                 return group
 
+    @retried(on=[IOError])
```
nfx, line 94: shouldn't you retry on `DatabrocksError` instead? what exception message were you getting to add retries, btw? we'll probably be getting more concrete exceptions in SDK soon, so i'd like comments with messages :) 

see https://github.com/databricks/databricks-sdk-py/pull/376

nfx (changes requested):
can you also add an integration test for this?

src/databricks/labs/ucx/install.py
mwojtyczka: I found it useful to have the possibility to check how many groups are still to be migrated (e.g. in case things are not working or fail in the middle of migration).

src/databricks/labs/ucx/workspace_access/groups.py
mwojtyczka: I agree it would be better to use the specific `DatabrocksError `. Let me update this.

mwojtyczka: I double-checked and indeed it was `DatabrocksError `. Code was updated accordingly.

mwojtyczka:
> can you also add an integration test for this?

There are existing tests for delete of backup groups but they don't cover the scenario where this is executed as a separate job. I will add this to the integration tests as a separate PR as I need to familiarize myself first with the setup.

src/databricks/labs/ucx/install.py
nfx: imagine that someone doesn't know what groups migration is or not doing that anymore - does it add value for them in this debug notebook? maybe it should be in one of the markdown files in `docs/` folder?..

mwojtyczka: sure I can move it under docs. I probably miss the overall vision of the tool. The local group migration is what this toll is currently used for.

src/databricks/labs/ucx/workspace_access/groups.py
mwojtyczka: fixed now

src/databricks/labs/ucx/install.py
mwojtyczka: i moved it to the docs/local-group-migration.md

nfx (approved):

Issue #89, opened 2023-08-18
Errors from `ThreadedExecution` are not handled


nfx:
In the current way of dealing with threads, we ignore reporting the exceptions. We should at least log them with a connection to a task pool.

https://superfastpython.com/threadpoolexecutor-exception-handling/

PR #376, opened 2023-10-04
Added logging of errors during threadpool operations.

larsgeorge-db:
Fixes: #89

### DIFF
--- a/src/databricks/labs/ucx/framework/parallel.py
+++ b/src/databricks/labs/ucx/framework/parallel.py
@@ -1,5 +1,6 @@
 import concurrent
 import datetime as dt
+import functools
 import logging
 from collections.abc import Callable
 from concurrent.futures import ALL_COMPLETED, ThreadPoolExecutor
@@ -32,10 +33,12 @@ class ProgressReporter:
 class ThreadedExecution(Generic[ExecutableResult]):
     def __init__(
         self,
+        name,
         executables: list[ExecutableFunction],
         num_threads: int | None = 4,
         progress_reporter: ProgressReporter | None = None,
     ):
+        self._name = name
         self._num_threads = num_threads
         self._executables = executables
         self._futures = []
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/framework/parallel.py
```diff
@@ -45,22 +48,36 @@ def __init__(
...
         logger.debug(f"Starting {len(self._executables)} tasks in {self._num_threads} threads")
 
+        def error_handler(func, name):
+            @functools.wraps(func)
+            def inner(*args, **kwargs):
+                try:
+                    return func(*args, **kwargs)
+                except Exception as ex:
+                    msg = getattr(ex, "message", ex)
+                    logger.warning(f"An error occurred while '{name}', message: {msg}")
```
nfx, line 63:
```suggestion
                    logger.warning(f"{name} task failed: {str(ex)}")
```
python exceptions don't always have `message`. and we need more concise log messages.

src/databricks/labs/ucx/framework/parallel.py
```diff
@@ -45,22 +48,36 @@ def __init__(
...
+                future = executor.submit(error_handler(executable, self._name))
                 future.add_done_callback(self._done_callback)
-                # TODO: errors are not handled yet - https://github.com/databricks/UC-Upgrade/issues/89
                 self._futures.append(future)
 
             results = concurrent.futures.wait(self._futures, return_when=ALL_COMPLETED)
 
         logger.debug("Collecting the results from threaded execution")
-        collected = [future.result() for future in results.done]
+        collected = [future.result() for future in results.done if future.result()]
```
nfx, line 77: i think there's a bug here, as `.result()` may not always be available the second time we call it. Just for counting, introduce a separate variable.

nfx (changes requested):

src/databricks/labs/ucx/framework/parallel.py
larsgeorge-db: Really? I thought (by reading the `_base.py` code) that only reading exceptions has side effects?

src/databricks/labs/ucx/framework/parallel.py
larsgeorge-db: But that is why I have the `getattr()` with the default value. Why call `str()` on the exception, a simple `{ex}` does the same, no (that is, call `__str__`)?

larsgeorge-db:
```
============================= test session starts ==============================
collecting ... collected 1 item
21:35  WARN [d.l.ucx.framework.parallel] testing task failed: failed
21:35  WARN [d.l.ucx.framework.parallel] The parallel run of 'testing' incurred 1 failed tasks

unit/framework/test_parallel.py::test_threaded_execution_gather_with_failed_task PASSED

============================== 1 passed in 0.22s ===============================

Process finished with exit code 0
```

nfx (approved):

PR #395, opened 2023-10-06
Crawler for Externally Orchestrated Jobs with Failing Configuration

### LINKED ISSUE #266: Assessment for `RunSubmit` API usages

zpappa:
Resolves #266

**Added ExternallyOrchestratedJobsWithFailingConfigCrawler**
Added a crawler to look at JobRuns from the SDK and determine which of the job runs are from the RunsSubmit API

**Added Unit Tests**
Added tests to cover basic logic and some edge cases

**Integration Tests Pending**


nfx (changes requested):
needs a passing integration test. please attach a screenshot once you get it to work locally

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -57,6 +61,14 @@ class PipelineInfo:
     failures: str
 
 
+@dataclass
```
FastLee, line 64: Do we want to capture only the failing ones or all of them with the failing one anotated?

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -65,6 +77,29 @@ def _azure_sp_conf_present_check(config: dict) -> bool:
     return False
 
 
+def is_custom_image(version_string: str):
+    """
+    Is this a custom version?
+    """
```
FastLee, line 83: Does it require implementation?

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -65,6 +77,29 @@ def _azure_sp_conf_present_check(config: dict) -> bool:
...
+    Is left version greater than right version
+    """
+    if left_version is None or is_custom_image(left_version):
+        return False
+    if right_version is None or is_custom_image(right_version):
+        return True
+    pattern = r"^(?P<major>\d+)?\.(?P<minor>\d+)?\.(?P<patch>[\dx]+)?.*"
+    lvg = re.match(pattern, left_version)
+    rvg = re.match(pattern, right_version)
+    left = (int(lvg.group("major")), int(lvg.group("minor")))
```
FastLee, line 98: Do we have a proper unit test for that?

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -84,6 +119,26 @@ def spark_version_compatibility(spark_version: str) -> str:
     return "supported"
 
 
+def get_job_cluster_from_task(
+    task: RunTask, job_run: BaseRun, all_clusters: dict[str, ClusterDetails]
```
FastLee, line 123: Re factor job cluster to use the same mechanism

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -250,3 +305,142 @@ def snapshot(self) -> list[ClusterInfo]:
...
+            sbe,
+            "hive_metastore",
+            schema,
+            "ext_orc_job_runs_with_failing_config",
+            ExternallyOrchestratedJobRunWithFailingConfiguration,
+        )
+        self._ws = ws
+
+    def _crawl(self) -> list[ExternallyOrchestratedJobRunWithFailingConfiguration]:
+        no_of_days_back = datetime.timedelta(days=30)  # todo make configurable in yaml?
```
FastLee, line 332: Make timedelta externally configurable

FastLee (changes requested):

Issue #36, opened 2023-06-14
Validation Suite & Report to confirm what is available in inventory vs migrated


pradeepravi87:
we need to have a validation suite which should validate the inventory vs migrated, because currently we are assuming the coverage is 100 percent by doing some sample validation and it’s quite manual. Can we have a feature which does the validation and provide some report at the end of the process?
Note - At times, validation might fail for some reasonable scenarios. For example, the account level group would have had more users compared to workspace level group . But that’s fine , it helps to retrospect and reconcile

## Current situation
```
@task("migrate-groups", depends_on=[crawl_permissions])
def migrate_permissions(cfg: WorkspaceConfig):
    toolkit = GroupMigrationToolkit(cfg)
    toolkit.prepare_environment()
    if toolkit.has_groups():
        toolkit.apply_permissions_to_backup_groups()
        toolkit.replace_workspace_groups_with_account_groups()
        toolkit.apply_permissions_to_account_groups()
    else:
        logger.info("Skipping group migration as no groups were found.")
```

## Expected outcome

Scope of issue is to split `migrate-groups` into a couple of workflows, adding rigorous checks after every state change. We already have `databricks.labs.ucx.workspace_access.verification.VerificationManager`, that should be used in `verify_permissions_applied_to_backup_groups`, `verify_workspace_groups_replaced_with_account_groups`, and `verify_permissions_applied_to_account_groups`.

```
@task("apply-permissions-to-backup-groups", depends_on=[crawl_permissions])
def apply_permissions_to_backup_groups(cfg: WorkspaceConfig):
    toolkit = GroupMigrationToolkit(cfg)
    toolkit.prepare_environment()
    if not toolkit.has_groups():
        logger.info("Skipping group migration as no groups were found.")
    toolkit.apply_permissions_to_backup_groups()

    # Added feature:
    toolkit.verify_permissions_applied_to_backup_groups() # <<<<<< ADDED CODE

@task("replace_workspace-groups-with-account-groups", depends_on=[apply_permissions_to_backup_groups])
def replace_workspace_groups_with_account_groups(cfg: WorkspaceConfig):
    toolkit = GroupMigrationToolkit(cfg)
    toolkit.prepare_environment()
    if not toolkit.has_groups():
        logger.info("Skipping group migration as no groups were found.")
    toolkit.replace_workspace_groups_with_account_groups()

    # Added feature:
    toolkit.verify_workspace_groups_replaced_with_account_groups() # <<<<<< ADDED CODE

[...]

PR #421, opened 2023-10-10
Check if permissions are up to date

william-conti:
Attempt to fix https://github.com/databrickslabs/ucx/issues/36

Goal of this PR is to add a validation in each class that implements the Crawler and Applier class.
This validation step will verify that a permissions that has been applied to an object is correct, if it's not the case, it attemps to re-apply this permission multiple times


larsgeorge-db:
How can this happen in the first place? Could you add some info around this use-case? Or edge case?

src/databricks/labs/ucx/workspace_access/base.py
```diff
@@ -57,3 +59,14 @@ def noop():
                 pass
 
             return partial(noop)
+
+    def response_to_request(self, acls: Optional["list[AccessControlResponse]"] = None) -> list[AccessControlRequest]:
```
nfx, line 63: it does not belong to this class. it only belongs to GenericPermissionsSupport. this has to be a private method.

src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -60,8 +60,17 @@ def _get_apply_task(
 
     @rate_limited(max_requests=30)
     def _applier_task(self, object_type: str, object_id: str, acl: list[iam.AccessControlRequest]):
-        self._ws.permissions.update(object_type, object_id, access_control_list=acl)
-        return True
+        permission = self._ws.permissions.update(object_type, object_id, access_control_list=acl)
+        request_acl = self.response_to_request(permission.access_control_list)
+
+        if acl == request_acl:
```
nfx, line 66: 1. comparing with response is not enough to detect configuration drift. it has to be a separate `get`.
2. plain `==` comparison does not work for it, because there might be other groups and/or users and/or service principals with permissions, that are not part of request. compare in a loop.
3. log a message - e.g. f'{object_type} {group} permissions {object_id} do not match: requested={requested} vs returned={returned}'

nfx (changes requested):

src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -58,10 +59,35 @@ def _get_apply_task(
         )
         return partial(self._applier_task, item.object_type, item.object_id, new_acl)
 
+    @staticmethod
+    def response_to_request(acls: Optional["list[iam.AccessControlResponse]"] = None) -> list[iam.AccessControlRequest]:
```
nfx, line 63: make method private

src/databricks/labs/ucx/workspace_access/generic.py
william-conti: Need to define the error message to make it easier, right now I feel I'm printing the entire group information for scim.py

src/databricks/labs/ucx/workspace_access/scim.py
```diff
@@ -12,6 +13,7 @@
     Crawler,
     Destination,
     Permissions,
+    logger,
```
nfx, line 16: Create logger in this module, don't import it from other modules, as we'll have wrong error messages

nfx (commented):
Few minor comments before we can merge this

src/databricks/labs/ucx/workspace_access/redash.py
```diff
@@ -83,8 +84,21 @@ def _applier_task(self, object_type: sql.ObjectTypePlural, object_id: str, acl:
         Please note that we only have SET option (DBSQL Permissions API doesn't support UPDATE operation).
         This affects the way how we prepare the new ACL request.
         """
-        self._ws.dbsql_permissions.set(object_type=object_type, object_id=object_id, access_control_list=acl)
-        return True
+        for _i in range(0, 3):
+            self._ws.dbsql_permissions.set(object_type=object_type, object_id=object_id, access_control_list=acl)
+
+            remote_permission = self._safe_get_dbsql_permissions(object_type, object_id)
+            if all(elem in remote_permission.access_control_list for elem in acl):
```
nfx, line 91: Are you sure it's not required to convert from response to request?

tests/integration/workspace_access/test_generic.py
```diff
@@ -0,0 +1,164 @@
...
+    assert len(applied_permissions.access_control_list) == 4
+    assert list_equal_unordered(
+        applied_permissions.access_control_list,
+        [
+            # TODO: Find what /directories/4218545820690421 corresponds to
+            iam.AccessControlResponse(
+                all_permissions=[
+                    iam.Permission(
+                        inherited=True,
+                        inherited_from_object=["/directories/4218545820690421"],
```
nfx, line 55: Number after directories is object id for workspace. This is a hardcoded id and cannot be merged, as this test will fail on the other workspace

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,106 @@
...
+            sql.AccessControl(group_name=ws_group.display_name, permission_level=sql.PermissionLevel.CAN_RUN)
+        ],
+    )
+
+    redash_acl_listing = [redash.redash_listing_wrapper(ws.queries.list, sql.ObjectTypePlural.QUERIES)]
+    sql_support = redash.SqlPermissionsSupport(ws, redash_acl_listing)
+
+    tasks = list(sql_support.get_crawler_tasks())
+    # Only one query exist in the workspace, so one crawler task
+    assert len(tasks) == 1
```
nfx, line 36: This will fail when multiple integration suits are run in parallel or there's some residue from previous tests

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,106 @@
...
+    value = apply_task()
+
+    # Validate that no errors has been thrown when applying permission to backup group
+    assert value
+
+    applied_permissions = ws.dbsql_permissions.get(object_type=ObjectTypePlural.QUERIES, object_id=query.id)
+
+    # Validate that permissions has been applied properly to the backup group
+    # and the old group permission has been revoked
+    assert len(applied_permissions.access_control_list) == 3
```
nfx, line 48: Don't rely on concrete numbers for a shared system

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,106 @@
...
+            sql.AccessControl(group_name=ws_group.display_name, permission_level=sql.PermissionLevel.CAN_RUN)
+        ],
+    )
+
+    redash_acl_listing = [redash.redash_listing_wrapper(ws.queries.list, sql.ObjectTypePlural.QUERIES)]
+    sql_support = redash.SqlPermissionsSupport(ws, redash_acl_listing)
+
+    tasks = list(sql_support.get_crawler_tasks())
+    # Only one query exist in the workspace, so one crawler task
+    assert len(tasks) == 1
```
nfx, line 84: Same

tests/integration/workspace_access/test_scim.py
```diff
@@ -0,0 +1,60 @@
...
+    backup_group = make_group(display_name=backup_group_name)
+
+    migration_state = GroupMigrationState()
+    migration_state.add(group=MigrationGroupInfo(workspace=ws_group, account=acc_group, backup=backup_group))
+
+    sup = ScimSupport(ws=ws)
+
+    tasks = list(sup.get_crawler_tasks())
+    # Crawler tasks should return the users, admins, and the ws_group as there are no other groups present
+    assert len(tasks) == 3
```
nfx, line 44: Will fail on race conditions

nfx (changes requested):
Few small changes

src/databricks/labs/ucx/workspace_access/redash.py
william-conti: remote_permission.access_control_list is type List[sql.AccessControl] and acl is also list[sql.AccessControl]

nfx (approved):

PR #424, opened 2023-10-10
disaster recovery from temp state

fannijako:
This fix helps to recover from a state, where 

- the temporary groups were created, 
- the account groups were not,
- the original workspace-local groups were deleted.

The fix **recreates the original state**, with _recreating all the workspace-local groups, which does have a temp pair but not an account level pair attached to the workspace_.

This helps to recreate the state of the migration with the already existing group_manager.migration_groups_provider


mwojtyczka:
The permissions should be taken from the inventory, not the backup group itself. The backup groups might not have all permissions applied correctly.

PR #444, opened 2023-10-12
Fixed `crawl_permissions` task to respect 'workspace_start_path' config

pritishpai:
The crawl_permissions task in assessment would not respect the config value of workspace_start_path as it was not present in the 'from_dict' classmethod. Added an extraction for the config.

Removed dbconnect from Contributing README as it is no longer needed.


nfx (approved):
👍

PR #458, opened 2023-10-13
Added `make install-dev` and a stronger `make clean` for easier dev on-boarding and release upgrades.

dmoore247:
… easier on-boarding experience

The problem was `make test` was reporting incompatible versions. The `hatch` environment needed cleaning up. Also added `make install-dev` as a convention to install the developer environment (hatch).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,11 @@ clean:
 	rm -fr htmlcov .mypy_cache .pytest_cache .ruff_cache .coverage coverage.xml
 	hatch env remove unit
 
-install-dev:
+dev:
 	pip install hatch
 	hatch env create
 	hatch run pip install -e '.[test]'
+	hatch run which python
 
 lint:
 	hatch run lint:verify
@@ -17,6 +18,9 @@ fmt:
 
 test:
 	hatch run unit:test
+
+integration:
+	hatch run integration:test
 
 test-cov:
 	hatch run unit:test-cov-report && open htmlcov/index.html
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,13 @@
+all: clean lint fmt test
+
 clean:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,5 +1,13 @@
+all: clean lint fmt test
+
 clean:
 	rm -fr htmlcov .mypy_cache .pytest_cache .ruff_cache .coverage coverage.xml
+	hatch env remove unit
+
+install-dev:
```
nfx, line 7:
```suggestion
dev:
```
We use this for python sdk

Makefile
```diff
@@ -11,4 +19,5 @@ test:
 	hatch run unit:test
 
 test-cov:
```
nfx, line 21: Perhaps call it "coverage"

nfx (changes requested):
Can you also modify the contributing.md?

Makefile
```diff
@@ -1,5 +1,13 @@
+all: clean lint fmt test
+
 clean:
 	rm -fr htmlcov .mypy_cache .pytest_cache .ruff_cache .coverage coverage.xml
+	hatch env remove unit
+
+install-dev:
+	pip install hatch
+	hatch env create
+	hatch run pip install -e '.[test]'
```
nfx, line 10: Can you output the path to python interpreter here?

nfx (changes requested):
Please also modify the contributing.md

Makefile
dmoore247: as in `which python` ?

nfx (approved):

Makefile
larsgeorge-db: I think Serge means this:
```
hatch run python -c "import sys; print(sys.executable)"
```

Issue #338, opened 2023-09-29
Report AWS Instance Profile to UC Storage Credentials (via query)


nfx:
We should automat migration ACLs based on Instance Profiles / Service Principals and other legacy security mechanisms

Related issues:
- https://github.com/databrickslabs/ucx/issues/100

pohlposition:
As part of this, we should output a report that tells the user what to do on the AWS side for each IAM role to enable UC to assume it, by establishing the trust relationship:
https://docs.databricks.com/en/data-governance/unity-catalog/get-started.html#configure-a-storage-bucket-and-iam-role-in-aws

zpappa:
# Background
As part of the migration effort, customers need to migrate policies tied to instance profiles to UC Storage Credentials.

# Pre-requisites
- Customer will need to provide an AWS account in order to get details about the AWS instance profiles attached? 
- OR the assumption is that an existing instance profile can perform these policy lookup operations

# Output
- Customer should be provided a report based on crawler inventory.
**aws_instance_profiles_to_be_migrated**
Field | Comment
| -- | -- |
instance_profile | the name of the instance profile
policy_json | generated policy json for an inline policy
policy_update_command | generated update command to update an existing policy or create based on the policy json

# Considerations
**Inline vs attached policy**
- If a policy is inline, we should generate a new IAM role JSON in the JSON field based on the inline policy
- If a policy is attached, i.e. an existing IAM role we should provide a command in the command field to establish a trust relationship with UC for assumption

pohlposition:
relates to https://github.com/databrickslabs/ucx/issues/94

pohlposition:
We have a list of Instance Profiles in the inventory.
 - Need to create a report / query (for assessment)
 - Need to include prescriptive next steps for AWS Admin
 - Need to create a new issue for mapping from old instance profile to new credential (customer may want to consolidate)

dmoore247:
I don't exactly follow the request here, but I do have code to [crawl through instance profiles and policy documents.](https://github.com/dmoore247/permissions-graph/blob/main/extract-triples/roles.py)

Also [code to walk glue catalogs](https://github.com/dmoore247/permissions-graph/blob/main/extract-triples/glue.py)

dmoore247:
@pohlposition @zpappa Is this to be added to the assessment report? This thing should be tagged as such.
Is this assignment to implement the crawler, or flesh out the details for all the steps?

dmoore247:
Jobs to be done:

[...]

PR #459, opened 2023-10-15
Added AWS instance profile crawler

dmoore247:
May close [#338](https://github.com/databrickslabs/ucx/issues/338)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,13 @@
+all: clean lint fmt test
+
 clean:
 	rm -fr htmlcov .mypy_cache .pytest_cache .ruff_cache .coverage coverage.xml
+	hatch env remove unit
+
+install-dev:
+	pip install hatch
+	hatch env create
+	hatch run pip install -e '.[test]'
 
 lint:
 	hatch run lint:verify
@@ -12,3 +20,4 @@ test:
 
 test-cov:
 	hatch run unit:test-cov-report && open htmlcov/index.html
+
--- a/src/databricks/labs/ucx/assessment/aws_instance_profiles.py
+++ b/src/databricks/labs/ucx/assessment/aws_instance_profiles.py
@@ -1,0 +1,47 @@
+from dataclasses import dataclass
+from databricks.sdk import WorkspaceClient
+from databricks.labs.ucx.framework.crawlers import CrawlerBase, SqlBackend
+
+@dataclass
+class AWSInstanceProfileInfo:
+    # The AWS ARN of the instance profile to register with Databricks.
+    # This field is required.
+    instance_profile_arn: str
+    # The AWS IAM role ARN of the role associated with the instance profile.
+    # This field is required if your role name and
+    # instance profile name do not match and you want to use
+    # the instance profile with Databricks SQL Serverless.
+    iam_role_arn: str = None
+    # Boolean flag indicating whether the instance profile should only be
+    # used in credential passthrough scenarios.
+    # If true, it means the instance profile contains an meta IAM role
+    # which could assume a wide range of roles. Therefore it should
+    # always be used with authorization. This field is optional,
+    # the default value is false.
+    is_meta_instance_profile: bool = False
+
+
+
+class AWSInstanceProfileCrawler(CrawlerBase):
+    """Crawl AWS workspace for Instance profiles"""
+
+    def __init__(self, ws: WorkspaceClient, sbe: SqlBackend, schema):
+        super().__init__(sbe, "hive_metastore", schema, "aws_instance_profiles", AWSInstanceProfileInfo)
+        self._ws = ws
+
+    def _crawl(self) -> list[AWSInstanceProfileInfo]:
+        instance_profiles = self._ws.instance_profiles.list()
+
+        for ips in instance_profiles:
+            yield AWSInstanceProfileInfo(
+                iam_role_arn=ips.iam_role_arn,
+                instance_profile_arn=ips.instance_profile_arn,
+                is_meta_instance_profile=ips.is_meta_instance_profile,
+            )
+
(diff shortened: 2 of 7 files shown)


dmoore247:
GitGuardian is failing this because of a pre-existing secret which is used in unit tests.

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -61,6 +61,25 @@ class PipelineInfo:
     failures: str
 
 
+@dataclass
+class AWSInstanceProfileInfo:
```
nfx, line 65: This file size is getting out of control, could you put the aws instance profiles related code into aws_instance_profiles.py in the same folder? :) we discussed it last week

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -173,6 +192,31 @@ def _try_fetch(self) -> list[GlobalInitScriptInfo]:
...
 
+class AWSInstanceProfileCrawler(CrawlerBase):
+    """Crawl AWS workspace for Instance profiles"""
+
+    def __init__(self, ws: WorkspaceClient, sbe: SqlBackend, schema):
+        super().__init__(sbe, "hive_metastore", schema, "aws_instance_profiles", AWSInstanceProfileInfo)
+        self._ws = ws
+
+    def _crawl(self) -> list[AWSInstanceProfileInfo]:
+        instance_profiles = self._ws.instance_profiles.list()
```
nfx, line 203:
```suggestion
        return self._ws.instance_profiles.list()
```
I wonder if we can just do this.

src/databricks/labs/ucx/runtime.py
```diff
@@ -138,6 +139,20 @@ def assess_pipelines(cfg: WorkspaceConfig):
     crawler.snapshot()
 
 
+@task("assessment", depends_on=[setup_schema])
+def assess_aws_instance_profiles(cfg: WorkspaceConfig):
```
nfx, line 143:
```suggestion
def crawl_aws_instance_profiles(cfg: WorkspaceConfig):
```
There's no assessment done there

Makefile
```diff
@@ -4,6 +4,7 @@ clean:
 	rm -fr htmlcov .mypy_cache .pytest_cache .ruff_cache .coverage coverage.xml
 	hatch env remove unit
 
+
```
nfx, line 7: Revert, irrelevant

src/databricks/labs/ucx/runtime.py
```diff
@@ -138,6 +139,20 @@ def assess_pipelines(cfg: WorkspaceConfig):
     crawler.snapshot()
 
 
+@task("assessment", depends_on=[setup_schema])
```
nfx, line 142:
```suggestion
@task("assessment", depends_on=[setup_schema], is_aws=True)
```

Here's the real blocker for this PR: at the moment, this task will always break on Azure and GCP because that API is not supported outside of aws. You have to modify install.py to add tasks only "if task.is_aws and self._ws.config.is_aws". Same for queries. For queries we can use file prefixes.

This won't install AWS related checks on azure at all

nfx (changes requested):
Also make pr title changelog-friendly

src/databricks/labs/ucx/assessment/crawlers.py
dmoore247: Makes sense

src/databricks/labs/ucx/runtime.py
dmoore247: @nfx I was wondering how this was accomplished!
Just following the azure example which didn't differentiate and should be fixed.

src/databricks/labs/ucx/assessment/crawlers.py
dmoore247: Considered that!
Following the Azure example, a new class was created. The new data class lets us mark fields as NULLable which lets us get through the null record filter and create a derived field `account_id` like @pohlposition might be asking for.

src/databricks/labs/ucx/runtime.py
dmoore247: Oh wait, you want me to add the cloud differentiator into this PR? That should have been done prior to the  AzureServicePrinciple commit. How about we create that differentiator in a separate PR and merge that in first.

dmoore247: Let's move this to https://github.com/databrickslabs/ucx/issues/460

nfx:
Please upgrade to the latest version. It seems that the version you're referring is too old.

src/databricks/labs/ucx/runtime.py
dmoore247: #460 replaced by https://github.com/databrickslabs/ucx/pull/610

src/databricks/labs/ucx/assessment/aws_instance_profiles.py
```diff
@@ -0,0 +1,47 @@
...
+    # the default value is false.
+    is_meta_instance_profile: bool = False
+
+
+
+class AWSInstanceProfileCrawler(CrawlerBase):
+    """Crawl AWS workspace for Instance profiles"""
+
+    def __init__(self, ws: WorkspaceClient, sbe: SqlBackend, schema):
+        super().__init__(sbe, "hive_metastore", schema, "aws_instance_profiles", AWSInstanceProfileInfo)
```
nfx, line 29: why do we need a table for this?... does `self._ws.instance_profiles.list()` call take more than few minutes or something?...

nfx (changes requested):
why do we need this table?... can't the dependent task be done with simply `self._ws.instance_profiles.list()`?...

PR #473, opened 2023-10-18
Added documentation to describe interaction with external Hive Metastores

FastLee:
Per our discussion.
Analysis of External HMS integration.

### DIFF
--- a/docs/external_hms.md
+++ b/docs/external_hms.md
@@ -27,6 +27,7 @@ The UCX toolkit by default relies on the internal workspace HMS as a source for 
 - We cannot support multiple HMS
 - Using an external HMS to persist UCX tables will break functionality for a second workspace using UCX
 - We should consider using a pattern similar to our integration testing to rename the target database to allow persisting from multiple workspaces. For example WS1 --> UCX_ABC, WS2 --> UCX_DEF.
+- With external HMS it is likely that some of the tables will not be accessible by some of the workspaces. We may need to migrate certain databases from certain workspaces.
 
 ### Suggested flow
 1. Start the installer.
--- a/docs/external_hms.md
+++ b/docs/external_hms.md
@@ -18,7 +18,7 @@ The UCX toolkit by default relies on the internal workspace HMS as a source for 
   - Init scripts
 
 ### Design Decisions
-- We should set up a single HMS for UCX
+- Should we set up a single HMS for UCX?
 - We should suggest copying the setup from an existing Cluster/Cluster policy
 - We shouldn't override the set up for the DB Warehouses (that may break functionality)
 - We should allow overriding cluster settings and instance profile setting to accommodate novel settings.
--- a/docs/external_hms.md
+++ b/docs/external_hms.md
@@ -1,0 +1,37 @@
+# External HMS Integration
+### TL;DR
+The UCX toolkit by default relies on the internal workspace HMS as a source for tables and views.
+<br/>Many DB users utilize an external HMS instead of the Workspace HMS provided by DB.
+<br/>A popular external HMS is Amazon Glue.
+<br/>This document describes the considerations for UCX integration with external HMS.
+
+### Current Considerations
+- Integration with external HMS is set up on individual clusters.
+- Theoretically we can integrate separate clusters in a workspace with different HMS repositories.
+- In reality most customers use a single (internal or external) HMS within a workspace.
+- When migrating from an external HMS we have to consider that it is used by more than one workspace.
+- Integration with external HMS has to be set on all DB Warehouses together.
+- HMS connectivity is set, usually, on cluster policy.
+- Typically external HMS setup relies on:
+  - Spark Config
+  - Instance Profiles
+  - Init scripts
+
+### Design Decisions
+- We should set up a single HMS for UCX
+- We should suggest copying the setup from an existing Cluster/Cluster policy
+- We shouldn't override the set up for the DB Warehouses (that may break functionality)
+- We should allow overriding cluster settings and instance profile setting to accommodate novel settings.
+
+### Challenges
+- We cannot support multiple HMS
+- Using an external HMS to persist UCX tables will break functionality for a second workspace using UCX
+- We should consider using a pattern similar to our integration testing to rename the target database to allow persisting from multiple workspaces. For example WS1 --> UCX_ABC, WS2 --> UCX_DEF.
+
+### Suggested flow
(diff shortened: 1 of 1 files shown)


docs/external_hms.md
```diff
@@ -0,0 +1,37 @@
+# External HMS Integration
+### TL;DR
+The UCX toolkit by default relies on the internal workspace HMS as a source for tables and views.
+<br/>Many DB users utilize an external HMS instead of the Workspace HMS provided by DB.
+<br/>A popular external HMS is Amazon Glue.
+<br/>This document describes the considerations for UCX integration with external HMS.
+
+### Current Considerations
+- Integration with external HMS is set up on individual clusters.
+- Theoretically we can integrate separate clusters in a workspace with different HMS repositories.
```
nfx, line 10: what do we need for it practically? please describe

docs/external_hms.md
```diff
@@ -0,0 +1,37 @@
...
+- In reality most customers use a single (internal or external) HMS within a workspace.
+- When migrating from an external HMS we have to consider that it is used by more than one workspace.
+- Integration with external HMS has to be set on all DB Warehouses together.
```
nfx, line 13: can Ext-HMS WH access workspace-local HMS?

docs/external_hms.md
```diff
@@ -0,0 +1,37 @@
...
+- HMS connectivity is set, usually, on cluster policy.
```
nfx, line 14:
```suggestion
- HMS connectivity is set, usually, on cluster policy. As well as global SQL Warehouse config
```
right?

docs/external_hms.md
```diff
@@ -0,0 +1,37 @@
...
+- Typically external HMS setup relies on:
+  - Spark Config
+  - Instance Profiles
+  - Init scripts
+
+### Design Decisions
+- We should set up a single HMS for UCX
```
nfx, line 21:
```suggestion
- Should we set up a single HMS for UCX?
```

docs/external_hms.md
```diff
@@ -0,0 +1,37 @@
...
+- We should suggest copying the setup from an existing Cluster/Cluster policy
```
nfx, line 22:
```suggestion
- Should we suggest copying the setup from an existing Cluster/Cluster policy?
```

docs/external_hms.md
```diff
@@ -0,0 +1,37 @@
...
+- We shouldn't override the set up for the DB Warehouses (that may break functionality)
+- We should allow overriding cluster settings and instance profile setting to accommodate novel settings.
+
+### Challenges
+- We cannot support multiple HMS
```
nfx, line 27: but we can store raw delta tables on DBFS

docs/external_hms.md
```diff
@@ -0,0 +1,37 @@
...
+- Using an external HMS to persist UCX tables will break functionality for a second workspace using UCX
```
nfx, line 28: please suggest couple of options on addressing this

nfx (changes requested):

docs/external_hms.md
```diff
@@ -0,0 +1,38 @@
...
+- Typically external HMS setup relies on:
+  - Spark Config
+  - Instance Profiles
+  - Init scripts
+
+### Design Decisions
+- We should set up a single HMS for UCX
+- We should suggest copying the setup from an existing Cluster/Cluster policy
+- We shouldn't override the set up for the DB Warehouses (that may break functionality)
+- We should allow overriding cluster settings and instance profile setting to accommodate novel settings.
```
nfx, line 24: This is requirement, not a design decision

docs/external_hms.md
```diff
@@ -0,0 +1,38 @@
...
+- We cannot support multiple HMS
+- Using an external HMS to persist UCX tables will break functionality for a second workspace using UCX
+- We should consider using a pattern similar to our integration testing to rename the target database to allow persisting from multiple workspaces. For example WS1 --> UCX_ABC, WS2 --> UCX_DEF.
+- With external HMS it is likely that some of the tables will not be accessible by some of the workspaces. We may need to migrate certain databases from certain workspaces.
+
+### Suggested flow
+1. Start the installer.
+2. The installer looks for use of external HMS by the workspace. We review cluster policies or DBSQL warehouses settings.
+3. We alert the user that an external HMS is set and request ask a YES/NO to set external HMS.
+4. We alert the user if they opted for external HMS and the DB Warehouses are not set for external HMS
```
nfx, line 36: Should be step3, not 4 :)

docs/external_hms.md
```diff
@@ -0,0 +1,38 @@
...
+  - Init scripts
+
+### Design Decisions
+- We should set up a single HMS for UCX
+- We should suggest copying the setup from an existing Cluster/Cluster policy
+- We shouldn't override the set up for the DB Warehouses (that may break functionality)
+- We should allow overriding cluster settings and instance profile setting to accommodate novel settings.
+
+### Challenges
+- We cannot support multiple HMS
```
HariGS-DB, line 27: Could we not set up one job task for each unique hms. Usually workspace dont use different HMS, but they likely to have one Ext HMS and one local HMS.

docs/external_hms.md
```diff
@@ -0,0 +1,38 @@
...
+- Using an external HMS to persist UCX tables will break functionality for a second workspace using UCX
+- We should consider using a pattern similar to our integration testing to rename the target database to allow persisting from multiple workspaces. For example WS1 --> UCX_ABC, WS2 --> UCX_DEF.
```
HariGS-DB, line 29: If multiple workspaces are sharing the same Ext HMS, we should do table crawl just once but should do multiple grants crawl per each workspace

docs/external_hms.md
pohlposition: For the inventory tables, if we use the Workspace ID in either the schema name "ucx-################" or as a column for uniqueness, then we don't have to worry about conflicts.

Likely easier to introduce in the schema name since our teardown scripts drop tables.

docs/external_hms.md
pohlposition: changing the code to only crawl the external HMS once is a premature optimization.  This operation only takes a few minutes right now.

docs/external_hms.md
pohlposition: Is the concern that a workspace may be connected to an external metatore on standard clusters, but the SQL Warehouse is not?  If so, is this common?

PR #481, opened 2023-10-20
Fixed issue with workspace listing process on None type object_type

HariGS-DB:
When running the assessment job, the tasks workspace listing, fails with the error "KeyError: 'object_type'"
the object_type property is not set for experiment objects and while listing it throws this error. Since scope of workspace listing is limited to Notebooks, Directories, Repos, Files, Libraries, suggesting a code change will will filter out cases where object_type is None.

### DIFF
--- a/src/databricks/labs/ucx/workspace_access/generic.py
+++ b/src/databricks/labs/ucx/workspace_access/generic.py
@@ -246,17 +246,15 @@ class WorkspaceListing(Listing, CrawlerBase):
 
         ws_listing = WorkspaceListing(self._ws, num_threads=self._num_threads, with_directories=False)
         for obj in ws_listing.walk(self._start_path):
-            if obj is None:
+            if obj is None or obj.object_type is None:
                 continue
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -246,7 +246,7 @@ def _crawl(self) -> list[WorkspaceObjectInfo]:
 
         ws_listing = WorkspaceListing(self._ws, num_threads=self._num_threads, with_directories=False)
         for obj in ws_listing.walk(self._start_path):
-            if obj is None:
+            if obj is None or obj.object_type is None:
```
nfx, line 249: When object type is none?

nfx:
Make PR title changelog ready

mwojtyczka:
This also solves another issue. Currently the tool trips on its self by listing its own folder. You get `object_type` is `None` in that case as well.

nfx (approved):

Issue #73, opened 2023-08-09
[BUG] Issues with Jobs (or other objects) that have no `OWNER`


renardeinside:
In some customer environments the permission application step is failing because of absence of the `OWNER` on the object.

Such situations can happen when  user who is owner of a group has been deleted.

Proposed fix - collect the problematic objects into a separate list, and then provide a warning message saying that these objects require an owner property prior to migration.

PR #485, opened 2023-10-20
Fixing assessment throwing an error when the owner of an object is empty

william-conti:
Fixes https://github.com/databrickslabs/ucx/issues/73

By default, all objects without owners will have their created_by column persisted as an empty string instead of null/None for the sake of backward compatibility.

It also adds Integration tests for the SQLBackend in order to harden the code


src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -150,7 +150,16 @@ def _crawl(self) -> list[GlobalInitScriptInfo]:
 
     def _assess_global_init_scripts(self, all_global_init_scripts):
         for gis in all_global_init_scripts:
-            global_init_script_info = GlobalInitScriptInfo(gis.script_id, gis.name, gis.created_by, gis.enabled, 1, "")
+            if not gis.created_by:
+                logger.warning(
+                    f"Script {gis.name} have Unknown creator, it means that the original creator has been deleted"
+                    f" and should be re-created"
+                )
+                global_init_script_info = GlobalInitScriptInfo(gis.script_id, gis.name, "", gis.enabled, 1, "")
```
nfx, line 158: Pattern for entire PR: don't surround creation of entire object with "if", but rather surround introduce a variable for the username.

Having two code paths for append may cause difficult bugs in the future

tests/unit/assessment/test_assessment.py
```diff
@@ -2514,3 +2518,134 @@ def test_list_all_pipeline_with_conf_spn_secret_avlb(mocker):
...
+            cluster_source=ClusterSource.JOB,
+        )
+    ]
+    ws = Mock()
+    mockbackend = MockBackend()
+    ws.jobs.list.return_value = sample_jobs
+    ws.clusters.list.return_value = sample_clusters
+    JobsCrawler(ws, mockbackend, "ucx").snapshot()
+    result = mockbackend.rows_written_for("hive_metastore.ucx.jobs", "append")
+    assert result == [JobInfo(job_id="536591785949415", job_name=None, creator="", success=1, failures="[]")]
```
nfx, line 2563: Can we do None for creator name everywhere, instead of empty strings? It translates into NULL, which has easier semantics

nfx (changes requested):
Where's the change for GenericPermissionsSupport? :)

tests/unit/assessment/test_assessment.py
william-conti: I've went with this pattern at first ... But it didn't worked for most of the objects because classes needs to be changed from this:

```
class JobInfo:
    job_id: str
    job_name: str
    creator: str
    success: int
    failures: str
```

to this

```
class JobInfo:
    job_id: str
    job_name: str
    success: int
    failures: str
    creator: str: None

```
Which is a breaking change as it changes column order ...

william-conti: Even using a class like this:

class JobInfo:
    job_id: str
    job_name: str
    creator: str | None
    success: int
    failures: str

is breaking because creator goes from NOT NULL to NULLABLE, but it won't require re-creation of existing ucx deployment if there are no objects with unknown owners

nfx: > Which is a breaking change as it changes column order ...

our insert statement generation is aware of the actual column order

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -36,29 +36,29 @@
...
 
 
 @dataclass
 class ClusterInfo:
-    cluster_id: str
-    cluster_name: str
-    creator: str
     success: int
     failures: str
+    cluster_id: str = None
```
nfx, line 50: cluster, job, pipeline IDs - they cannot be done. it makes no sense ;)

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -150,7 +150,19 @@ def _crawl(self) -> list[GlobalInitScriptInfo]:
...
+                    f"Script {gis.name} have Unknown creator, it means that the original creator has been deleted"
+                    f" and should be re-created"
+                )
+            global_init_script_info = GlobalInitScriptInfo(
+                script_id=gis.script_id,
+                script_name=gis.name,
+                created_by=gis.created_by,
+                enabled=gis.enabled,
+                success=1,
+                failures="",
```
nfx, line 164:
```suggestion
                failures="[]",
```

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -384,7 +396,19 @@ def _crawl(self) -> list[PipelineInfo]:
...
+                logger.warning(
+                    f"Pipeline {pipeline.name} have Unknown creator, it means that the original creator "
+                    f"has been deleted and should be re-created"
+                )
+            pipeline_info = PipelineInfo(
+                pipeline_id=pipeline.pipeline_id,
+                pipeline_name=pipeline.name,
+                creator_name=pipeline.creator_user_name,
+                success=1,
+                failures="",
```
nfx, line 409:
```suggestion
                failures="[]",
```

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -417,7 +441,18 @@ def _assess_clusters(self, all_clusters):
...
+                logger.warning(
+                    f"Cluster {cluster.cluster_id} have Unknown creator, it means that the original creator "
+                    f"has been deleted and should be re-created"
+                )
+            cluster_info = ClusterInfo(
+                cluster_id=cluster.cluster_id,
+                cluster_name=cluster.cluster_name,
+                creator=cluster.creator_user_name,
+                success=1,
+                failures="",
```
nfx, line 454:
```suggestion
                failures="[]",
```

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -505,7 +540,19 @@ def _assess_jobs(self, all_jobs: list[BaseJob], all_clusters_by_id) -> list[JobI
...
+                    f"Job {job.job_id} have Unknown creator, it means that the original creator has been deleted "
+                    f"and should be re-created"
+                )
+
+            job_details[job.job_id] = JobInfo(
+                job_id=str(job.job_id),
+                job_name=job.settings.name,
+                creator=job.creator_user_name,
+                success=1,
+                failures="",
```
nfx, line 554:
```suggestion
                failures="[]",
```

nfx (changes requested):
integration tests must pass

src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -129,15 +130,31 @@ def _applier_task(self, object_type: str, object_id: str, acl: list[iam.AccessCo
...
     @rate_limited(max_requests=100)
     def _crawler_task(self, object_type: str, object_id: str) -> Permissions | None:
+        objects_with_owner_permission = ["jobs", "pipeline"]
+
         permissions = self._safe_get_permissions(object_type, object_id)
         if not permissions:
+            logger.warning(f"Object {object_type} {object_id} doesn't have any permissions")
+            return None
+        if not self._object_have_owner(permissions) and object_type in objects_with_owner_permission:
+            logger.warning(
```
nfx, line 140: can you add a unit test for this? :)

tests/unit/assessment/test_assessment.py
```diff
@@ -2664,3 +2668,134 @@ def test_list_all_pipeline_with_conf_spn_secret_avlb(mocker):
     assert result_set[0].get("application_id") == "Hello, World!"
     assert result_set[0].get("tenant_id") == "directory_12345"
     assert result_set[0].get("storage_account") == "newstorageacct"
+
+
+def test_job_crawler_with_no_owner_should_have_empty_creator_name():
```
nfx, line 2673: this better could be an integration test, but fine :)

tests/unit/workspace_access/test_generic.py
```diff
@@ -541,3 +548,54 @@ def test_workspace_snapshot():
...
+            Listing(ws.pipelines.list, "pipeline_id", "pipelines"),
+        ],
+    )
+    executables = list(sup.get_crawler_tasks())
+    tasks = [task() for task in executables]
+    assert len(tasks) == 1
+    assert tasks[0] == Permissions(
+        object_id="1234",
+        object_type="clusters",
+        raw='{"access_control_list": [{"all_permissions": ['
```
nfx, line 598: don't test for serialized JSON, because the order of fields may change between invocations.

nfx (changes requested):
could you add a unit test for the warning by codecov? :) 
image

nfx (approved):

Issue #499, opened 2023-10-25
Mitigate issue that Table ACL grants are not atomic


mwojtyczka:
There is a bug in the Grants API for the Table ACLs. It does not support concurrent grant/revoke operations.

> TableAcl grant/revoke operations are not atomic. When granting the permissions, the service would first get all existing permissions, append with the new permissions, and set the full list in the database. If there are concurrent grant requests, both requests might succeed and emit the audit logs, but what actually happens could be that the new permission list from one request overrides the other one, causing permission loss.

We won't hit the issue as long as we grant permissions for one securable to one user in one grant/statement. e.g., 
GRANT SELECT, CREATE, MODIFY ON table my_table TO `some-user`

Possible solutions:
* Remove threading from Table ACLs 
* Fold actions belonging to the same principal, object type and id into one grant statement (better)

[ES ticket for reference.](https://databricks.atlassian.net/browse/ES-908737)

nfx:
Make sure to add this as a comment in the code

mwojtyczka:
Fixed by [PR 512](https://github.com/databrickslabs/ucx/pull/512)

PR #506, opened 2023-10-25
Execute Table ACL Grants serially to mitigate platform limitations

mwojtyczka:
* Added support for applying permissions serially
* Execute grants for Table ACLs serially, fix for issue [499](https://github.com/databrickslabs/ucx/issues/499)

### DIFF
--- a/src/databricks/labs/ucx/workspace_access/base.py
+++ b/src/databricks/labs/ucx/workspace_access/base.py
@@ -38,3 +38,7 @@ class AclSupport:
     @abstractmethod
     def object_types(self) -> set[str]:
         """This method returns a set of strings, that represent object types that are applicable by this instance."""
+
+    @abstractmethod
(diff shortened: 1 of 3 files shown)


nfx (changes requested):
adding abstract method to a base class just to workaround a bug in the platform is not acceptable way to go. modify only the query construction in TableAclSupport class.  https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/workspace_access/tacl.py#L34-L43

see https://docs.databricks.com/en/sql/language-manual/security-grant.html &

mwojtyczka:
Will be fixed in PR [513](https://github.com/databrickslabs/ucx/pull/512) instead.

mwojtyczka:
Will be fixed in PR https://github.com/databrickslabs/ucx/pull/512 instead.

Issue #489, opened 2023-10-23
Add retry to redash permissions


mwojtyczka:
All applier tasks should have retry. There are likely other applier types that need a retry so all of them should be checked and retry added.

pohlposition:
What does this have to do with "redash"?

Is the ask for certain tasks within a workflow (which one?) to be able to be "retried"?

I don't understand the scenario.  This seems like a nice-to-have and not a bug.

PR #507, opened 2023-10-26
implement _safe_* calls for applier_task in redash, scim, generic

tnguyen-db:
* Resolves https://github.com/databrickslabs/ucx/issues/489
* Removes for loop in applier_task in redash, scim, generic
* Sets the right rate for each
* Implements _safe_* calls

### DIFF
--- a/src/databricks/labs/ucx/workspace_access/generic.py
+++ b/src/databricks/labs/ucx/workspace_access/generic.py
@@ -3,6 +3,7 @@ import json
 import logging
 from collections.abc import Callable, Iterator
 from dataclasses import dataclass
+from datetime import timedelta
 from functools import partial
 from typing import Optional
 
@@ -59,9 +60,12 @@ class Listing:
 
 
 class GenericPermissionsSupport(AclSupport):
-    def __init__(self, ws: WorkspaceClient, listings: list[Listing]):
+    def __init__(
+        self, ws: WorkspaceClient, listings: list[Listing], verify_timeout: timedelta | None = timedelta(minutes=20)
+    ):
         self._ws = ws
         self._listings = listings
+        self._verify_timeout = verify_timeout
 
     def get_crawler_tasks(self):
         for listing in self._listings:
@@ -107,22 +111,31 @@ class GenericPermissionsSupport(AclSupport):
                 )
         return results
 
-    @rate_limited(max_requests=30)
-    def _applier_task(self, object_type: str, object_id: str, acl: list[iam.AccessControlRequest]):
-        self._safe_update_permissions(object_type, object_id, acl=acl)
-
-        remote_permission = self._safe_get_permissions(object_type, object_id)
+    def _inflight_check(self, object_type: str, object_id: str, acl: list[iam.AccessControlRequest]):
+        # in-flight check for the applied permissions
+        # the api might be inconsistent, therefore we need to check that the permissions were applied
+        set_retry_on_value_error = retried(on=[RetryableError], timeout=self._verify_timeout)
+        set_retried_check = set_retry_on_value_error(self._safe_get_permissions)
+        remote_permission = set_retried_check(object_type, object_id)
         remote_permission_as_request = self._response_to_request(remote_permission.access_control_list)
         if all(elem in remote_permission_as_request for elem in acl):
             return True
+        else:
+            msg = f"""Couldn't apply appropriate permission for object type {object_type} with id {object_id}
+                acl to be applied={acl}
+                acl found in the object={remote_permission_as_request}
+                """
+            raise ValueError(msg)
 
-        logger.warning(
-            f"""Couldn't apply appropriate permission for object type {object_type} with id {object_id}
-        acl to be applied={acl}
-        acl found in the object={remote_permission_as_request}
-        """
-        )
-        return False
+    @rate_limited(max_requests=30)
+    def _applier_task(self, object_type: str, object_id: str, acl: list[iam.AccessControlRequest]):
+        update_retry_on_value_error = retried(on=[RetryableError], timeout=self._verify_timeout)
+        update_retried_check = update_retry_on_value_error(self._safe_update_permissions)
+        update_retried_check(object_type, object_id, acl)
+
+        retry_on_value_error = retried(on=[ValueError], timeout=self._verify_timeout)
+        retried_check = retry_on_value_error(self._inflight_check)
+        return retried_check(object_id, object_id, acl)
 
     @rate_limited(max_requests=100)
     def _crawler_task(self, object_type: str, object_id: str) -> Permissions | None:
@@ -170,7 +183,6 @@ class GenericPermissionsSupport(AclSupport):
         return acl.service_principal_name
 
     # TODO remove after ES-892977 is fixed
-    @retried(on=[RetryableError])
     def _safe_get_permissions(self, object_type: str, object_id: str) -> iam.ObjectPermissions | None:
         try:
             return self._ws.permissions.get(object_type, object_id)
@@ -186,7 +198,6 @@ class GenericPermissionsSupport(AclSupport):
             else:
                 raise RetryableError() from e
 
-    @retried(on=[RetryableError])
     def _safe_update_permissions(
         self, object_type: str, object_id: str, acl: list[iam.AccessControlRequest]
     ) -> iam.ObjectPermissions | None:
--- a/src/databricks/labs/ucx/workspace_access/generic.py
+++ b/src/databricks/labs/ucx/workspace_access/generic.py
@@ -200,7 +200,6 @@ class GenericPermissionsSupport(AclSupport):
                 "PERMISSION_DENIED",
                 "FEATURE_DISABLED",
                 "RESOURCE_DOES_NOT_EXIST",
-                "INTERNAL_SERVER_ERROR",
             ]:
                 logger.warning(f"Could not update permissions for {object_type} {object_id} due to {e.error_code}")
                 return None
--- a/src/databricks/labs/ucx/workspace_access/generic.py
+++ b/src/databricks/labs/ucx/workspace_access/generic.py
@@ -136,7 +136,7 @@ class GenericPermissionsSupport(AclSupport):
 
         retry_on_value_error = retried(on=[ValueError, RetryableError], timeout=self._verify_timeout)
         retried_check = retry_on_value_error(self._inflight_check)
-        return retried_check(object_id, object_id, acl)
+        return retried_check(object_type, object_id, acl)
 
     @rate_limited(max_requests=100)
     def _crawler_task(self, object_type: str, object_id: str) -> Permissions | None:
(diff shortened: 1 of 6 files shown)


larsgeorge-db (commented):
Please explain the change to me. I do not understand how the loop and the "retries on errors" achieve the same. More inline...

src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -110,21 +109,19 @@ def _response_to_request(
 
     @rate_limited(max_requests=30)
     def _applier_task(self, object_type: str, object_id: str, acl: list[iam.AccessControlRequest]):
-        for _i in range(0, 3):
```
larsgeorge-db, line 113: So this used a loop to apply the same (full set, mind you) permissions until they are all available on the object. There is no error handling here (that is, no `try/except`).

src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -189,6 +186,26 @@ def _safe_get_permissions(self, object_type: str, object_id: str) -> iam.ObjectP
             else:
                 raise RetryableError() from e
 
+    @retried(on=[RetryableError])
+    def _safe_update_permissions(
```
larsgeorge-db, line 190: This handles errors now, but how does this ensure all permissions have been applied successfully?

larsgeorge-db:
Also, this adds a bunch of new code paths that the test now covers less. Please add more coverage in the tests.

src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -110,21 +109,19 @@ def _response_to_request(
 
     @rate_limited(max_requests=30)
     def _applier_task(self, object_type: str, object_id: str, acl: list[iam.AccessControlRequest]):
```
nfx, line 111: If you remove the loop here, you must add "@retried(on=[this], timeout=timedelta(minutes=1))", preferably you make time delta configurable in the unit tests, otherwise they'll run slower. 


At this moment this change is a regression and introduces more bugs than it fixes

src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -189,6 +186,26 @@ def _safe_get_permissions(self, object_type: str, object_id: str) -> iam.ObjectP
             else:
                 raise RetryableError() from e
 
+    @retried(on=[RetryableError])
```
nfx, line 189: No timeout configured. See how it's configurable for unit testing in other places

nfx (changes requested):
This PR introduces regressions, please address all the comments

src/databricks/labs/ucx/workspace_access/generic.py
Retried: Leave me alone

src/databricks/labs/ucx/workspace_access/generic.py
tnguyen-db: If no error is returned, can we assume the permissions have been applied?

src/databricks/labs/ucx/workspace_access/generic.py
tnguyen-db: Does it make sense to add `@retried` on top if it's already present in the 2 function calls?

src/databricks/labs/ucx/workspace_access/generic.py
tnguyen-db: Which timeout value do you suggest?
I haven't found examples on how to configure it in the unit tests.
I was rather thinking of mocking `@retried` in the unit tests, wdyt?

tnguyen-db:
@larsgeorge-db 
The main problem this PR was trying to solve was to introduce retries for the dash permissions. 
[self._ws.dbsql_permissions.set](https://github.com/databrickslabs/ucx/blob/50301a6884d77bd244820c4f30a6a66038d03bce/src/databricks/labs/ucx/workspace_access/redash.py#L111) only had the 3 attempts in the loop which did not take into consideration the types of errors. 
The solution attempts to have a better control on the retry mechanism by lowering them at the level of the API call.

I've then generalized the solution for `generic` and `scim`

src/databricks/labs/ucx/workspace_access/generic.py
nfx: @tnguyen-db see https://github.com/databrickslabs/ucx/blob/bc7f29e0727c7b38d53f7c43e194fbe653bf46d6/src/databricks/labs/ucx/workspace_access/secrets.py#L92C1-L97

src/databricks/labs/ucx/workspace_access/generic.py
nfx: @tnguyen-db did you look at this one?
https://github.com/databrickslabs/ucx/blob/bc7f29e0727c7b38d53f7c43e194fbe653bf46d6/src/databricks/labs/ucx/workspace_access/secrets.py#L92-L97

src/databricks/labs/ucx/workspace_access/redash.py
```diff
@@ -101,30 +105,41 @@ def _crawler_task(self, object_id: str, object_type: sql.ObjectTypePlural) -> Pe
...
-            remote_permission = self._safe_get_dbsql_permissions(object_type, object_id)
-            if all(elem in remote_permission.access_control_list for elem in acl):
-                return True
+        set_retry_on_value_error = retried(on=[RetryableError], timeout=self._verify_timeout)
+        set_retried_check = set_retry_on_value_error(self._safe_set_permissions)
+        set_retried_check(object_type, object_id, acl)
+
+        retry_on_value_error = retried(on=[ValueError], timeout=self._verify_timeout)
+        retried_check = retry_on_value_error(self._inflight_check)
+        return retried_check(object_id, object_id, acl)
```
tnguyen-db, line 138: @nfx is it the kind of logic you expect?

tests/unit/workspace_access/test_redash.py
```diff
@@ -180,13 +183,15 @@ def test_applier_task_should_return_false_if_permission_are_not_up_to_date():
...
-    assert not result
+    sup = RedashPermissionsSupport(ws=ws, listings=[], verify_timeout=timedelta(seconds=1))
+    with pytest.raises(TimeoutError) as e:
+        result = sup._applier_task(
+            sql.ObjectTypePlural.QUERIES,
+            "test",
+            [sql.AccessControl(group_name="group_1", permission_level=sql.PermissionLevel.CAN_RUN)],
+        )
+    assert "Timed out after" in str(e.value)
+
```
tnguyen-db, line 194: @nfx same question as above but for the test

src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -60,9 +60,12 @@ def __iter__(self):
 
 
 class GenericPermissionsSupport(AclSupport):
-    def __init__(self, ws: WorkspaceClient, listings: list[Listing]):
+    def __init__(
+        self, ws: WorkspaceClient, listings: list[Listing], verify_timeout: timedelta | None = timedelta(minutes=20)
```
nfx, line 64:
```suggestion
        self, ws: WorkspaceClient, listings: list[Listing], verify_timeout: timedelta | None = timedelta(minutes=1)
```
20 minutes is too much, 1 has to be enough

src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -108,24 +111,31 @@ def _response_to_request(
                 )
         return results
 
+    def _inflight_check(self, object_type: str, object_id: str, acl: list[iam.AccessControlRequest]):
```
nfx, line 114: you have to retry the inflight check, not within the inflight check

src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -189,6 +198,25 @@ def _safe_get_permissions(self, object_type: str, object_id: str) -> iam.ObjectP
...
+                "INVALID_PARAMETER_VALUE",
+                "UNAUTHORIZED",
+                "PERMISSION_DENIED",
+                "FEATURE_DISABLED",
+                "RESOURCE_DOES_NOT_EXIST",
+            ]:
+                logger.warning(f"Could not update permissions for {object_type} {object_id} due to {e.error_code}")
+                return None
+            else:
+                raise RetryableError() from e
```
nfx, line 218: what error codes are we retrying then?... please add a comment, otherwise it's not clear

src/databricks/labs/ucx/workspace_access/generic.py
```diff
@@ -108,24 +111,31 @@ def _response_to_request(
                 )
         return results
 
+    def _inflight_check(self, object_type: str, object_id: str, acl: list[iam.AccessControlRequest]):
+        # in-flight check for the applied permissions
+        # the api might be inconsistent, therefore we need to check that the permissions were applied
+        retry_on_value_error = retried(on=[RetryableError], timeout=self._verify_timeout)
+        retried_check = retry_on_value_error(self._safe_get_permissions)
+        remote_permission = retried_check(object_type, object_id)
+        remote_permission_as_request = self._response_to_request(remote_permission.access_control_list)
```
nfx, line 120: bug: self._safe_get_permissions can return None and this will crash

nfx (changes requested):

src/databricks/labs/ucx/workspace_access/generic.py
larsgeorge-db: Apparently not? Please consult with @renardeinside, as there are seemingly cases where the function returns without error, but the permission is not available? I am just making the point that before it looped over presence of permissions, not that it runs without exceptions.

tnguyen-db: Yes, indeed, there is now `_inflight_check` to check if the permissions have been applied as expected in `_applier_task` for each type

mwojtyczka: There is an edge case in platform issue where grants are not available even though they return 200. [ ES ticket](https://databricks.atlassian.net/browse/ES-908737). This [PR](https://github.com/databrickslabs/ucx/pull/512) should solve this.
In general, when the API returns 200 the permission should be applied. If it's not then it should be reported as platform bug via ES ticket. What are the scenarios that we see the permissions are not being applied even though there is not exception?

tnguyen-db: Apparently, [the API can be inconsistent](https://github.com/databrickslabs/ucx/blob/50301a6884d77bd244820c4f30a6a66038d03bce/src/databricks/labs/ucx/workspace_access/secrets.py#L84-L85), I don't know if the problem has been raised already to the platform

mwojtyczka: Got it, we should fix that but this should also be reported to the platform.

nfx (approved):
lgtm

nfx:
Closing this as merged

PR #512, opened 2023-10-26
Breaking change: Mitigate permissions loss in Table ACLs by folding grants belonging to the same principal, object id and object type together

mwojtyczka:
Applying grants that belong to the same principle, object id, and object type concurrently is not supported in legacy Table ACLs currently:

> TableAcl grant/revoke operations are not atomic. When granting the permissions, the service would first get all existing permissions, append with the new permissions, and set the full list in the database. If there are concurrent grant requests, both requests might succeed and emit the audit logs, but what actually happens could be that the new permission list from one request overrides the other one, causing permission loss.

This PR will:
* Resolve issue [499](https://github.com/databrickslabs/ucx/issues/499). 
* Improve performance of migrating Table ACLs permissions
* Added tests covering objects ownership
* Increased test coverage around Table ACLs

**Example**

Instead of executing grants in separate transactions:
`GRANT SELECT ON TABLE hive_metastore.db_a.table_a TO group_a`
`GRANT MODIFY ON TABLE hive_metastore.db_a.table_a TO group_a`

Folding and executing in one statement:
`GRANT SELECT, MODIFY ON TABLE hive_metastore.db_a.table_a TO group_a`

### DIFF
--- a/src/databricks/labs/ucx/workspace_access/tacl.py
+++ b/src/databricks/labs/ucx/workspace_access/tacl.py
@@ -42,7 +42,7 @@ class TableAclSupport(AclSupport):
             key = (grant.principal, grant.this_type_and_key())
             folded_actions[key].add(grant.action_type)
 
-            # use one of the grant objects for all actions per principal, object type and id
+            # use one of the grants with actions folded per principal, object type and id
             grant_dict = dataclasses.asdict(grant)
             grant_dict["action_type"] = ", ".join(sorted(folded_actions[key]))
             grant_folded_actions[key] = grant_dict
--- a/src/databricks/labs/ucx/workspace_access/tacl.py
+++ b/src/databricks/labs/ucx/workspace_access/tacl.py
@@ -1,5 +1,6 @@
 import collections
 import dataclasses
+import functools
 import json
 from collections.abc import Callable, Iterator
 from functools import partial
@@ -36,21 +37,43 @@ class TableAclSupport(AclSupport):
         # * GRANT MODIFY ON TABLE hive_metastore.db_a.table_a TO group_a
         # will be folded and executed in one statement/transaction:
         # * GRANT SELECT, MODIFY ON TABLE hive_metastore.db_a.table_a TO group_a
+
         folded_actions = collections.defaultdict(set)
-        grant_folded_actions = {}
         for grant in self._grants_crawler.snapshot():
             key = (grant.principal, grant.this_type_and_key())
             folded_actions[key].add(grant.action_type)
 
-            # use one of the grants with actions folded per principal, object type and id
-            grant_dict = dataclasses.asdict(grant)
-            grant_dict["action_type"] = ", ".join(sorted(folded_actions[key]))
-            grant_folded_actions[key] = grant_dict
+        def inner(object_type: str, object_id: str, grant: Grant) -> Permissions:
+            return Permissions(object_type=object_type, object_id=object_id, raw=json.dumps(dataclasses.asdict(grant)))
 
-        for (_principal, (object_type, object_id)), grant in grant_folded_actions.items():
-            yield lambda ot=object_type, oi=object_id, g=grant: Permissions(
-                object_type=ot, object_id=oi, raw=json.dumps(g)
-            )
+        for (principal, (object_type, object_id)), actions in folded_actions.items():
+            grant = self._from_reduced(object_type, object_id, principal, ", ".join(sorted(actions)))
+            yield functools.partial(inner, object_type=object_type, object_id=object_id, grant=grant)
+
+    def _from_reduced(self, object_type: str, object_id: str, principal: str, action_type: str):
+        match object_type:
+            case "TABLE":
+                catalog, database, table = object_id.split(".")
+                return Grant(
+                    principal=principal, action_type=action_type, catalog=catalog, database=database, table=table
+                )
+            case "VIEW":
+                catalog, database, view = object_id.split(".")
+                return Grant(
+                    principal=principal, action_type=action_type, catalog=catalog, database=database, view=view
+                )
+            case "DATABASE":
+                catalog, database = object_id.split(".")
+                return Grant(principal=principal, action_type=action_type, catalog=catalog, database=database)
+            case "CATALOG":
+                catalog = object_id
+                return Grant(principal=principal, action_type=action_type, catalog=catalog)
+            case "ANONYMOUS FUNCTION":
+                catalog = object_id
+                return Grant(principal=principal, action_type=action_type, catalog=catalog, anonymous_function=True)
+            case "ANY FILE":
+                catalog = object_id
+                return Grant(principal=principal, action_type=action_type, catalog=catalog, any_file=True)
 
     def object_types(self) -> set[str]:
         return {"TABLE", "DATABASE", "VIEW", "CATALOG", "ANONYMOUS FUNCTION", "ANY FILE"}
--- a/src/databricks/labs/ucx/workspace_access/tacl.py
+++ b/src/databricks/labs/ucx/workspace_access/tacl.py
@@ -1,6 +1,5 @@
 import collections
 import dataclasses
-import functools
 import json
 from collections.abc import Callable, Iterator
 from functools import partial
@@ -37,19 +36,21 @@ class TableAclSupport(AclSupport):
         # * GRANT MODIFY ON TABLE hive_metastore.db_a.table_a TO group_a
         # will be folded and executed in one statement/transaction:
         # * GRANT SELECT, MODIFY ON TABLE hive_metastore.db_a.table_a TO group_a
-        folded = collections.defaultdict(lambda: {"action_type": set()})
+        folded_actions = collections.defaultdict(set)
+        grant_folded_actions = {}
         for grant in self._grants_crawler.snapshot():
             key = (grant.principal, grant.this_type_and_key())
-            folded[key]["action_type"].add(grant.action_type)
+            folded_actions[key].add(grant.action_type)
+
+            # use one of the grant objects for all actions per principal, object type and id
             grant_dict = dataclasses.asdict(grant)
-            grant_dict["action_type"] = ", ".join(sorted(folded[key]["action_type"]))
-            folded[key]["grant_folded"] = grant_dict
+            grant_dict["action_type"] = ", ".join(sorted(folded_actions[key]))
+            grant_folded_actions[key] = grant_dict
 
-        def inner(obj_type: str, obj_key: str, grant_folded: dict) -> Permissions:
-            return Permissions(object_type=obj_type, object_id=obj_key, raw=json.dumps(grant_folded))
-
-        for (_principal, (object_type, object_key)), grant_data in folded.items():
-            yield functools.partial(inner, object_type, object_key, grant_data["grant_folded"])
+        for (_principal, (object_type, object_id)), grant in grant_folded_actions.items():
+            yield lambda ot=object_type, oi=object_id, g=grant: Permissions(
+                object_type=ot, object_id=oi, raw=json.dumps(g)
+            )
 
     def object_types(self) -> set[str]:
         return {"TABLE", "DATABASE", "VIEW", "CATALOG", "ANONYMOUS FUNCTION", "ANY FILE"}
--- a/src/databricks/labs/ucx/workspace_access/tacl.py
+++ b/src/databricks/labs/ucx/workspace_access/tacl.py
@@ -22,8 +22,21 @@ class TableAclSupport(AclSupport):
         self._sql_backend = sql_backend
 
     def get_crawler_tasks(self) -> Iterator[Callable[..., Permissions | None]]:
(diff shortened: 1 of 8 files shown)


src/databricks/labs/ucx/workspace_access/tacl.py
```diff
@@ -21,12 +22,34 @@ def __init__(self, grants_crawler: GrantsCrawler, sql_backend: SqlBackend):
...
+        #
+        # Below optimization mitigates the issue by folding all action types (grants)
+        # for the same principal, object_id and object_type into one grant with comma separated list of action types.
+        #
+        # For example, the following table grants:
+        # * GRANT SELECT ON TABLE hive_metastore.db_a.table_a TO group_a
+        # * GRANT MODIFY ON TABLE hive_metastore.db_a.table_a TO group_a
+        # will be folded and executed in one statement/transaction:
+        # * GRANT SELECT, MODIFY ON TABLE hive_metastore.db_a.table_a TO group_a
+        folded = collections.defaultdict(lambda: {"action_type": set()})
```
nfx, line 40:
```suggestion
        folded = collections.defaultdict(set)
```
Why do we need more than a set?

src/databricks/labs/ucx/workspace_access/tacl.py
```diff
@@ -21,12 +22,34 @@ def __init__(self, grants_crawler: GrantsCrawler, sql_backend: SqlBackend):
...
         for grant in self._grants_crawler.snapshot():
-            yield functools.partial(inner, grant)
+            key = (grant.principal, grant.this_type_and_key())
+            folded[key]["action_type"].add(grant.action_type)
+            grant_dict = dataclasses.asdict(grant)
+            grant_dict["action_type"] = ", ".join(sorted(folded[key]["action_type"]))
+            folded[key]["grant_folded"] = grant_dict
```
nfx, line 46: Why do we need this?

src/databricks/labs/ucx/workspace_access/tacl.py
```diff
@@ -21,12 +22,34 @@ def __init__(self, grants_crawler: GrantsCrawler, sql_backend: SqlBackend):
...
+
+        def inner(obj_type: str, obj_key: str, grant_folded: dict) -> Permissions:
+            return Permissions(object_type=obj_type, object_id=obj_key, raw=json.dumps(grant_folded))
+
+        for (_principal, (object_type, object_key)), grant_data in folded.items():
+            yield functools.partial(inner, object_type, object_key, grant_data["grant_folded"])
```
nfx, line 53: Why you can't yield lambda?

tests/integration/workspace_access/test_tacl.py
```diff
@@ -58,8 +59,25 @@ def test_recover_permissions_from_grants(
...
+    actual_raw_permissions = next(
+        obj.raw
+        for obj in permissions_list
+        if obj.object_id == dummy_table.full_name
+        and obj.object_type == "TABLE"
+        and json.loads(obj.raw)["principal"] == ws_group.display_name
+    )
+
+    assert (
+        f'{{"principal": "{ws_group.display_name}", "action_type": "MODIFY, SELECT", '
```
nfx, line 79: Don't assert on JSON string, assert on fields. Keys will be serialized in different order each time and flake the test

tests/unit/workspace_access/test_tacl.py
```diff
@@ -30,6 +30,56 @@ def test_tacl_crawler():
...
+    table_acl_support = TableAclSupport(grants_crawler, sql_backend)
+
+    crawler_tasks = table_acl_support.get_crawler_tasks()
+
+    permissions = next(crawler_tasks)()
+
+    assert "TABLE" == permissions.object_type
+    assert "catalog_a.database_b.table_c" == permissions.object_id
+    assert (
+        '{"principal": "foo@example.com", "action_type": "MODIFY, SELECT", '
```
nfx, line 57: Don't assert on JSON string

nfx (changes requested):

src/databricks/labs/ucx/workspace_access/tacl.py
mwojtyczka: we need to fold the action types and emit the just one grant for each key with the updated action type

src/databricks/labs/ucx/workspace_access/tacl.py
mwojtyczka: Class 'set' does not define '__getitem__', so the '[]' operator cannot be used on its instances. I need that to manipulate the action types below.

src/databricks/labs/ucx/workspace_access/tacl.py
mwojtyczka: actually, i tried that before, but it did not work. Did not have time to investigate so left it as it was before with partial.

tests/integration/workspace_access/test_tacl.py
mwojtyczka: will correct

tests/unit/workspace_access/test_tacl.py
mwojtyczka: will correct

src/databricks/labs/ucx/workspace_access/tacl.py
mwojtyczka: simplified

tests/integration/workspace_access/test_tacl.py
mwojtyczka: corrected

tests/unit/workspace_access/test_tacl.py
mwojtyczka: corrected

src/databricks/labs/ucx/workspace_access/tacl.py
```diff
@@ -21,12 +21,36 @@ def __init__(self, grants_crawler: GrantsCrawler, sql_backend: SqlBackend):
...
+        grant_folded_actions = {}
         for grant in self._grants_crawler.snapshot():
-            yield functools.partial(inner, grant)
+            key = (grant.principal, grant.this_type_and_key())
+            folded_actions[key].add(grant.action_type)
+
+            # use one of the grants with actions folded per principal, object type and id
+            grant_dict = dataclasses.asdict(grant)
+            grant_dict["action_type"] = ", ".join(sorted(folded_actions[key]))
+            grant_folded_actions[key] = grant_dict
```
nfx, line 48: this is a risk of corrupting data, can't accept this PR

nfx (changes requested):

nfx:
do reconstruction of grants, not the overwrite

```
Index: src/databricks/labs/ucx/workspace_access/tacl.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/databricks/labs/ucx/workspace_access/tacl.py b/src/databricks/labs/ucx/workspace_access/tacl.py
--- a/src/databricks/labs/ucx/workspace_access/tacl.py	(revision 45bae4a880095ce35719cc01c27c05f1786ac711)
+++ b/src/databricks/labs/ucx/workspace_access/tacl.py	(date 1698304643087)
@@ -1,5 +1,5 @@
+import collections
 import dataclasses
-import functools
 import json
 from collections.abc import Callable, Iterator
 from functools import partial
@@ -21,18 +21,37 @@
         self._sql_backend = sql_backend
 
     def get_crawler_tasks(self) -> Iterator[Callable[..., Permissions | None]]:
-        def inner(grant: Grant) -> Permissions:
-            object_type, object_key = grant.this_type_and_key()
-            return Permissions(object_type=object_type, object_id=object_key, raw=json.dumps(dataclasses.asdict(grant)))
-
+        folded = collections.defaultdict(set)
         for grant in self._grants_crawler.snapshot():
-            yield functools.partial(inner, grant)
+            folded[(grant.principal, grant.this_type_and_key())].add(grant.action_type)
+        for (principal, (object_type, key)), actions in folded.items():
+            yield lambda: Permissions(
+                object_type=object_type,
+                object_id=f'{principal}/{key}',
+                raw=json.dumps(sorted(actions)))
 
     def object_types(self) -> set[str]:
         return {"TABLE", "DATABASE", "VIEW", "CATALOG", "ANONYMOUS FUNCTION", "ANY FILE"}
 
+    def _from_reduced(self, item: Permissions):
+        actions = ', '.join(json.loads(item.raw))
+        principal, object_id = item.object_id
+        raw = {'principal': principal, 'action_type': actions}
+        match item.object_type:
+            case 'TABLE':
+                return raw | {'table': object_id}
+            case 'VIEW':
+                return raw | {'view': object_id}
+            case 'DATABASE':
+                return raw | {'database': object_id}
+            case 'ANONYMOUS FUNCTION':
+                return raw | {'anonymous_function': object_id}
+            case 'CATALOG':
+                return raw | {'catalog': object_id}
+
     def get_apply_task(self, item: Permissions, migration_state: GroupMigrationState, destination: Destination):
-        grant = Grant(**json.loads(item.raw))
+        raw = self._from_reduced(item)
+        grant = Grant(**raw)
         target_principal = migration_state.get_target_principal(grant.principal, destination)
         if target_principal is None:
             # this is a grant for user, service principal, or irrelevant group
```

mwojtyczka:
> do reconstruction of grants, not the overwrite
> 
> ```
> Index: src/databricks/labs/ucx/workspace_access/tacl.py
> IDEA additional info:
> Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
> <+>UTF-8
> ===================================================================
> diff --git a/src/databricks/labs/ucx/workspace_access/tacl.py b/src/databricks/labs/ucx/workspace_access/tacl.py
> --- a/src/databricks/labs/ucx/workspace_access/tacl.py	(revision 45bae4a880095ce35719cc01c27c05f1786ac711)
> +++ b/src/databricks/labs/ucx/workspace_access/tacl.py	(date 1698304643087)
> @@ -1,5 +1,5 @@
> +import collections
>  import dataclasses
> -import functools
>  import json
>  from collections.abc import Callable, Iterator
>  from functools import partial
> @@ -21,18 +21,37 @@
>          self._sql_backend = sql_backend
>  
>      def get_crawler_tasks(self) -> Iterator[Callable[..., Permissions | None]]:
> -        def inner(grant: Grant) -> Permissions:
> -            object_type, object_key = grant.this_type_and_key()
> -            return Permissions(object_type=object_type, object_id=object_key, raw=json.dumps(dataclasses.asdict(grant)))
> -
> +        folded = collections.defaultdict(set)
>          for grant in self._grants_crawler.snapshot():
> -            yield functools.partial(inner, grant)
> +            folded[(grant.principal, grant.this_type_and_key())].add(grant.action_type)
> +        for (principal, (object_type, key)), actions in folded.items():
> +            yield lambda: Permissions(
> +                object_type=object_type,
> +                object_id=f'{principal}/{key}',
> +                raw=json.dumps(sorted(actions)))
>  
>      def object_types(self) -> set[str]:
>          return {"TABLE", "DATABASE", "VIEW", "CATALOG", "ANONYMOUS FUNCTION", "ANY FILE"}
>  
> +    def _from_reduced(self, item: Permissions):
> +        actions = ', '.join(json.loads(item.raw))
> +        principal, object_id = item.object_id
> +        raw = {'principal': principal, 'action_type': actions}
> +        match item.object_type:
> +            case 'TABLE':
> +                return raw | {'table': object_id}
> +            case 'VIEW':
> +                return raw | {'view': object_id}
> +            case 'DATABASE':
> +                return raw | {'database': object_id}
> +            case 'ANONYMOUS FUNCTION':
> +                return raw | {'anonymous_function': object_id}
> +            case 'CATALOG':
> +                return raw | {'catalog': object_id}
> +
>      def get_apply_task(self, item: Permissions, migration_state: GroupMigrationState, destination: Destination):
> -        grant = Grant(**json.loads(item.raw))
> +        raw = self._from_reduced(item)
> +        grant = Grant(**raw)
>          target_principal = migration_state.get_target_principal(grant.principal, destination)
>          if target_principal is None:
>              # this is a grant for user, service principal, or irrelevant group
> ```

Maybe I am wrong here but IMO this implementation proposal does not make sense for several reasons:
- The raw permissions is incomplete. It only contains actions. This may introduce a lot of bugs because the consumers may not be aware that the permissions object is incomplete. It will also break a lot of existing tests.
- object id will contain both the object and principal. Why an object id should ever contain information about a principal? it does not fit logically. It will also break a lot of existing tests.
- The permissions is reconstructed wrongly. A lot of fields will be missing in the raw field. Again this will break a lot of existing tests. 

IMO the crawler should return complete permissions to avoid issues with consumers of the method that may not be aware of the limitation that the raw object is incomplete. A better approach would be to create a complete permissions by creating a Grant based on the information at hand. I pushed an updated version for doing exactly that.

src/databricks/labs/ucx/workspace_access/tacl.py
mwojtyczka: corrected

nfx (approved):

PR #514, opened 2023-10-27
Fixed handling of keys with and without values.

### LINKED ISSUE #491: install.sh fails with the following error "ValueError: not enough values to unpack"
image

larsgeorge-db:
Fixes #491


FastLee (approved):

FastLee:
Update the branch

PR #518, opened 2023-10-27
Mitigate permissions loss in Table ACLs by running appliers with single thread

mwojtyczka:
Applying grants that belong to the same principle, object id, and object type concurrently is not supported in legacy Table ACLs currently:

> TableAcl grant/revoke operations are not atomic. When granting the permissions, the service would first get all existing permissions, append with the new permissions, and set the full list in the database. If there are concurrent grant requests, both requests might succeed and emit the audit logs, but what actually happens could be that the new permission list from one request overrides the other one, causing permission loss.

This PR will:
* Resolve issue https://github.com/databrickslabs/ucx/issues/499.

### DIFF
--- a/src/databricks/labs/ucx/framework/parallel.py
+++ b/src/databricks/labs/ucx/framework/parallel.py
@@ -27,8 +27,9 @@ class Threads(Generic[Result]):
         self._default_log_every = 100
 
     @classmethod
-    def gather(cls, name: str, tasks: list[Callable[..., Result]], num_threads: int = None) \
-            -> (list[Result], list[Exception]):
+    def gather(
+        cls, name: str, tasks: list[Callable[..., Result]], num_threads: int | None = None
+    ) -> (list[Result], list[Exception]):
         if num_threads is None:
             num_threads = os.cpu_count() * 2
             if num_threads < MIN_THREADS:
--- a/src/databricks/labs/ucx/framework/parallel.py
+++ b/src/databricks/labs/ucx/framework/parallel.py
@@ -27,8 +27,10 @@ class Threads(Generic[Result]):
         self._default_log_every = 100
 
     @classmethod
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/framework/parallel.py
```diff
@@ -27,8 +27,10 @@ def __init__(self, name, tasks: list[Callable[..., Result]], num_threads: int):
         self._default_log_every = 100
 
     @classmethod
-    def gather(cls, name: str, tasks: list[Callable[..., Result]]) -> (list[Result], list[Exception]):
-        num_threads = os.cpu_count() * 2
+    def gather(cls, name: str, tasks: list[Callable[..., Result]], num_threads: int = None) \
+            -> (list[Result], list[Exception]):
+        if num_threads is None:
+            num_threads = os.cpu_count() * 2
```
nfx, line 33: next the following two lines into this if statement - otherwise the value would always be 8

src/databricks/labs/ucx/workspace_access/manager.py
```diff
@@ -118,7 +118,9 @@ def apply_group_permissions(self, migration_state: GroupMigrationState, destinat
             applier_tasks.extend(tasks_for_support)
 
         logger.info(f"Starting to apply permissions on {destination} groups. Total tasks: {len(applier_tasks)}")
-        _, errors = Threads.gather(f"apply {destination} group permissions", applier_tasks)
+        # run with 1 thread to temp bugfix TableACL grants issue in the platform
+        # see: https://databricks.atlassian.net/browse/ES-908737
+        _, errors = Threads.gather(f"apply {destination} group permissions", applier_tasks, num_threads=1)
```
nfx, line 123: get it from configuration, not hardcoded value

nfx (changes requested):

nfx (approved):

PR #532, opened 2023-10-30
Improve error message for not null constraints violated

william-conti:
Currently there are a lot of customers facing an issue looking like this:

`pyspark.errors.exceptions.base.PySparkValueError: [CANNOT_BE_NONE] Argument `obj` can not be None.`

Basically it's always due to the fact that the code tries to insert a None value in a field that is NOT NULL. 
It's very hard to debug and understand what's going on there for customers and we should improve the error message here


nfx (approved):
Lgtm

mwojtyczka (approved):

Issue #531, opened 2023-10-30
Show databricks host(s) being used by install.sh.


dmoore247:
While running install.sh I notice that it doesn't provide feedback as to where the install is going. This is critical piece of information to ensure the environment isn't configured incorrectly.

dmoore247:
If this is an upgrade, one does receive a warning with the environment, however for a fresh install, no such info is provided.

nfx:
Thank you for the feature request! Currently, the team operates in a limited capacity, carefully prioritizing, and we cannot provide a timeline to implement this feature. Please make a Pull Request if you'd like to see this feature sooner, and we'll guide you through the journey.

nfx:
Done

PR #536, opened 2023-10-31
Added handling for workspaces that have write protected DBFS.

### LINKED ISSUE #493: Install fails to write to dbfs:/ Workspace FS isn't enough to run the jobs
At customer site, `dbfs:/` is blocked due to bucket policies. Writing the the /Workspace/Users/<email>/.ucx works.
The installer fails on the `dbfs:/` write. Jobs don't run off of the workspace FS.

We're blocked on the installer, trying to manually create the job from JSON and manually load the `.whl` file into an interactive cluster as a work around.

dmoore247:
This may close #493 and #531.

Implemented cluster overrides in installer. User is prompted for cluster ids for a legacy unassigned cluster and a Legacy TACL cluster, then these clusters are used for jobs.
The reason for using interactive clusters is to have compute where the UCX admin has manually loaded the python wheel file as the the wheel file can't be read from Workspace FS and used in python tasks with the current platform release.

### DIFF
--- a/src/databricks/labs/ucx/config.py
+++ b/src/databricks/labs/ucx/config.py
@@ -198,36 +198,36 @@ class AccountConfig(_Config["AccountConfig"]):
 
 @dataclass
 class WorkspaceConfig(_Config["WorkspaceConfig"]):
-    inventory_database: str
-    groups: GroupsConfig
-    instance_pool_id: str = None
-    warehouse_id: str = None
-    connect: ConnectConfig | None = None
-    num_threads: int | None = 10
-    log_level: str | None = "INFO"
+    connect: ConnectConfig = None
     database_to_catalog_mapping: dict[str, str] = None
     default_catalog: str = "ucx_default"
-
-    # Starting path for notebooks and directories crawler
-    workspace_start_path: str = "/"
+    groups: GroupsConfig = None
+    instance_pool_id: str = None
     instance_profile: str = None
+    inventory_database: str = None
+    log_level: str = "INFO"
+    num_threads: int = 10
+    override_clusters: dict[Any, Any] = None
     spark_conf: dict[str, str] = None
+    warehouse_id: str = None
+    workspace_start_path: str = "/"  # Starting path for notebooks and directories crawler
 
     @classmethod
     def from_dict(cls, raw: dict):
         cls._verify_version(raw)
         return cls(
-            inventory_database=raw.get("inventory_database"),
+            database_to_catalog_mapping=raw.get("database_to_catalog_mapping", None),
+            default_catalog=raw.get("default_catalog", "main"),
             groups=GroupsConfig.from_dict(raw.get("groups", {})),
             connect=ConnectConfig.from_dict(raw.get("connect", {})),
             instance_pool_id=raw.get("instance_pool_id", None),
-            warehouse_id=raw.get("warehouse_id", None),
-            num_threads=raw.get("num_threads", 10),
-            log_level=raw.get("log_level", "INFO"),
-            database_to_catalog_mapping=raw.get("database_to_catalog_mapping", None),
             instance_profile=raw.get("instance_profile", None),
+            inventory_database=raw.get("inventory_database"),
+            log_level=raw.get("log_level", "INFO"),
+            num_threads=raw.get("num_threads", 10),
+            override_clusters=raw.get("override_clusters", None),
             spark_conf=raw.get("spark_conf", None),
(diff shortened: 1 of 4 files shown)


dmoore247:
Still running install level checks, does look promising.
This PR does thus far create DAGs w/o task level libraries, stores cluster IDs in the config.yml file. If the cluster IDs do change, the Workflow DAG would have to be updated (manually) or by re-install

src/databricks/labs/ucx/config.py
```diff
@@ -198,36 +198,36 @@ def to_account_client(self) -> AccountClient:
...
+    groups: GroupsConfig = None
+    instance_pool_id: str = None
     instance_profile: str = None
+    inventory_database: str = None
+    log_level: str = "INFO"
+    num_threads: int = 10
+    override_clusters: dict[Any, Any] = None
     spark_conf: dict[str, str] = None
+    warehouse_id: str = None
+    workspace_start_path: str = "/"  # Starting path for notebooks and directories crawler
```
nfx, line 213: Please don't reorder the fields in this pr - it's hard to review what is the change

src/databricks/labs/ucx/install.py
```diff
@@ -288,6 +288,15 @@ def warehouse_type(_):
         else:
             groups_config_args["auto"] = True
 
+        # Checking for override clusters
+        if os.environ.get("DATABRICKS_CLUSTER_ID", None) and os.environ.get(
```
nfx, line 292: Ask for clusters via self._question, not the env vars

src/databricks/labs/ucx/install.py
```diff
@@ -486,9 +496,12 @@ def _upload_wheel(self) -> str:
...
             with local_wheel.open("rb") as f:
-                self._ws.dbfs.mkdirs(remote_dirname)
-                logger.info(f"Uploading wheel to dbfs:{remote_wheel}")
-                self._ws.dbfs.upload(remote_wheel, f, overwrite=True)
+                try:
+                    self._ws.dbfs.mkdirs(remote_dirname)
+                    logger.info(f"Uploading wheel to dbfs:{remote_wheel}")
+                    self._ws.dbfs.upload(remote_wheel, f, overwrite=True)
+                except DatabricksError as err:
+                    logger.warning(f"Uploading wheel files to DBFS failed, perhaps DBFS is protected. {err}")
```
nfx, line 504: Swallow the exception only if user is explicitly requesting for a write-protected dbfs handling

src/databricks/labs/ucx/install.py
```diff
@@ -534,13 +548,15 @@ def _upload_wheel_runner(self, remote_wheel: str):
...
     def _apply_cluster_overrides(settings: dict[str, any], overrides: dict[str, str], wheel_runner: str) -> dict:
+        logger.info("Applying cluster overrides")
         settings["job_clusters"] = [_ for _ in settings["job_clusters"] if _.job_cluster_key not in overrides]
         for job_task in settings["tasks"]:
             if job_task.job_cluster_key is None:
                 continue
             if job_task.job_cluster_key in overrides:
                 job_task.existing_cluster_id = overrides[job_task.job_cluster_key]
                 job_task.job_cluster_key = None
+                job_task.libraries = None  # Job libraries not loading from Workspace FS
```
nfx, line 559: Do it only when special dbfs handling is requested for

nfx (changes requested):

src/databricks/labs/ucx/install.py
dmoore247: That's the guidance I needed.

src/databricks/labs/ucx/install.py
dmoore247: @nfx There are no semantics to ask for write protected DBFS. Maybe we can recommend cluster overrides when this fails. I'm not clear on the data structure to know if the overrides are available at the time of this upload. I'll take another look.

src/databricks/labs/ucx/config.py
dmoore247: Will keep that in mind for the future. I dealing with a tricky mess at the time where adding one more property was causing an error.

src/databricks/labs/ucx/install.py
dmoore247: Addressed this logic. First time through, user will run into ERROR and then instructed to supply cluster overrides.

dmoore247:
@nfx Abandoning this PR in favor of this re-implement https://github.com/databrickslabs/ucx/pull/610

Issue #601, opened 2023-11-18
crawl cluster breaks on custom runtime value


dmoore247:
Environment: Pulled from main (0.6.0+PR#600), aws workspace, UCX v0.6.1+120231118114616
Crawler code should handle custom runtimes

```
ValueError: invalid literal for int() with base 10: 'custom:snapshot__14'
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File ~/.ipykernel/1615/command--1-1539221514:18
     15 entry = [ep for ep in metadata.distribution("databricks_labs_ucx").entry_points if ep.name == "runtime"]
     16 if entry:
     17   # Load and execute the entrypoint, assumes no parameters
---> 18   entry[0].load()()
     19 else:
     20   import databricks_labs_ucx

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/runtime.py:299, in main(*argv)
    297 if len(argv) == 0:
    298     argv = sys.argv
--> 299 trigger(*argv)

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/framework/tasks.py:149, in trigger(*argv)
    146         f.write(f"[run #{workflow_run_id}](/#job/{job_id}/run/{workflow_run_id})\n")
    148 try:
--> 149     current_task.fn(cfg)
    150 except BaseException as error:
    151     log_file_for_cli = str(log_file).lstrip("/Workspace")

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/runtime.py:114, in assess_clusters(cfg)
    112 ws = WorkspaceClient(config=cfg.to_databricks_config())
    113 crawler = ClustersCrawler(ws, RuntimeBackend(), cfg.inventory_database)
--> 114 crawler.snapshot()

File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/databricks/labs/ucx/assessment/crawlers.py:512, in ClustersCrawler.snapshot(self)
    511 def snapshot(self) -> list[ClusterInfo]:
--> 512     return self._snapshot(self._try_fetch, self._crawl)

[...]

PR #602, opened 2023-11-19
Fixed `crawl_cluster` failure over custom runtimes

dmoore247:
May fix https://github.com/databrickslabs/ucx/issues/601 
Test for `custom`, `dlt` and add exception handling around casting to int.
Add unit tests based on observation of internal workspaces.

### DIFF
--- a/src/databricks/labs/ucx/assessment/crawlers.py
+++ b/src/databricks/labs/ucx/assessment/crawlers.py
@@ -125,13 +125,23 @@ def spark_version_compatibility(spark_version: str) -> str:
     first_comp_custom_x = 2
     dbr_version_components = spark_version.split("-")
     first_components = dbr_version_components[0].split(".")
+    if "custom" in spark_version:
+        # custom runtime
+        return "unsupported"
+    if "dlt" in spark_version:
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -125,13 +125,23 @@ def spark_version_compatibility(spark_version: str) -> str:
     first_comp_custom_x = 2
     dbr_version_components = spark_version.split("-")
     first_components = dbr_version_components[0].split(".")
+    if "custom" in spark_version:
```
alexott, line 128: should it be better to do `spark_version.starts_with("custom:")` instead?

src/databricks/labs/ucx/assessment/crawlers.py
```diff
@@ -125,13 +125,23 @@ def spark_version_compatibility(spark_version: str) -> str:
...
+        # custom runtime
+        return "unsupported"
+    if "dlt" in spark_version:
```
alexott, line 131: same here

alexott (commented):
looks good, just some comments

nfx (approved):

PR #582, opened 2023-11-16
fix/issue_545: Include hostname in jdbc:/

pritishpai:
The UCX tool fetches the 'location' from the table properties while for an external location with jdbc, the host information is stored in 'storage properties'.

Add a field for Table class and fetch Storage Properties from the table. Using the host field from Storage properties, create the full jdbc url

PR #621, opened 2023-11-22
Fixed External Location widget to return hostname for `jdbc:` locations

pritishpai:
The UCX tool fetches the 'location' from the table properties while for an external location with jdbc, the host information is stored in 'storage properties'.

Add a field for Table class and fetch Storage Properties from the table. Using the host field from Storage properties, create the full jdbc url

Older PR closed due to Git issues: https://github.com/databrickslabs/ucx/pull/582


william-conti (approved):
lgtm, integration tests should pass

PR #461, opened 2023-10-16
Added task cloud flags to generate custom cloud assessment workflows

dmoore247:
May close #460, is needed before #459 is finalized.

PR #630, opened 2023-11-27
Added task filtering by cloud

dmoore247:
This will filter out Azure only tasks when on AWS cloud and similarly for all clouds and task types.
This will be useful for cloud specific assessment types such as AWS Instance profiles, Azure service principals etc.

### DIFF
--- a/src/databricks/labs/ucx/framework/tasks.py
+++ b/src/databricks/labs/ucx/framework/tasks.py
@@ -5,6 +5,8 @@ from functools import wraps
 from logging.handlers import TimedRotatingFileHandler
 from pathlib import Path
 
+from databricks.sdk.core import Config
+
 from databricks.labs.ucx.__about__ import __version__
 from databricks.labs.ucx.config import WorkspaceConfig
 from databricks.labs.ucx.framework.logger import _install
@@ -24,6 +26,22 @@ class Task:
     notebook: str = None
     dashboard: str = None
 
+    # if all are False then task is for all clouds
+    is_aws: bool = True
+    is_azure: bool = True
+    is_gcp: bool = True
+
+    def cloud_compatible(self, config: Config) -> bool:
+        """Test compatibility between config and task"""
+        return any(
+            [
+                config.is_aws == self.is_aws,
+                config.is_azure == self.is_azure,
+                config.is_gcp == self.is_gcp,
+                (self.is_aws is False and self.is_azure is False and self.is_gcp is False),
+            ]
+        )
+
 
 @staticmethod
 def _remove_extra_indentation(doc: str) -> str:
@@ -37,7 +55,17 @@ def _remove_extra_indentation(doc: str) -> str:
     return "\n".join(stripped)
 
 
-def task(workflow, *, depends_on=None, job_cluster="main", notebook: str | None = None, dashboard: str | None = None):
+def task(
+    workflow,
+    *,
+    depends_on=None,
+    job_cluster="main",
+    notebook: str | None = None,
+    dashboard: str | None = None,
+    is_azure: bool = True,
+    is_aws: bool = True,
+    is_gcp: bool = True,
+):
     def decorator(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
@@ -79,6 +107,9 @@ def task(workflow, *, depends_on=None, job_cluster="main", notebook: str | None 
             job_cluster=job_cluster,
             notebook=notebook,
             dashboard=dashboard,
+            is_aws=is_aws,
+            is_azure=is_azure,
+            is_gcp=is_gcp,
         )
 
         return wrapper
--- a/src/databricks/labs/ucx/runtime.py
+++ b/src/databricks/labs/ucx/runtime.py
@@ -130,7 +130,7 @@ def assess_pipelines(cfg: WorkspaceConfig):
     crawler.snapshot()
 
 
-@task("assessment")
+@task("assessment", is_azure=True, is_aws=False, is_gcp=False)
 def assess_azure_service_principals(cfg: WorkspaceConfig):
     """This module scans through all the clusters configurations, cluster policies, job cluster configurations,
     Pipeline configurations, Warehouse configuration and identifies all the Azure Service Principals who has been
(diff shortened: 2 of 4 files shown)


src/databricks/labs/ucx/runtime.py
```diff
@@ -130,7 +130,7 @@ def assess_pipelines(cfg: WorkspaceConfig):
     crawler.snapshot()
 
 
-@task("assessment")
+@task("assessment", is_azure=True, is_aws=False, is_gcp=False)
```
william-conti, line 133: Can we have is_azure=True only ? It allows for less boilerplate and easier readability

william-conti (changes requested):

src/databricks/labs/ucx/runtime.py
dmoore247: @william-conti That was considered in the first implementation.. that led to more complex logic evaluation, a bug and made it difficult to evaluate correctness. This is only needed on cloud specific tasks (1 current, 1 planned).

william-conti: Why ? IMO you need to modify cloud_compatible method to compare the current task key and compare it with the conf and done

william-conti (approved):

src/databricks/labs/ucx/framework/tasks.py
```diff
@@ -24,6 +26,22 @@ class Task:
     notebook: str = None
     dashboard: str = None
 
+    # if all are False then task is for all clouds
+    is_aws: bool = True
+    is_azure: bool = True
+    is_gcp: bool = True
```
FastLee, line 32: That's not a clean implementation. Capture the the cloud from the config class.

src/databricks/labs/ucx/framework/tasks.py
```diff
@@ -24,6 +26,22 @@ class Task:
...
+
+    def cloud_compatible(self, config: Config) -> bool:
```
FastLee, line 34: Same as above

src/databricks/labs/ucx/framework/tasks.py
```diff
@@ -79,6 +107,9 @@ def wrapper(*args, **kwargs):
             job_cluster=job_cluster,
             notebook=notebook,
             dashboard=dashboard,
+            is_aws=is_aws,
+            is_azure=is_azure,
+            is_gcp=is_gcp,
```
FastLee, line 112: Same as above

src/databricks/labs/ucx/runtime.py
FastLee, line 133: Use a list to support multiple clouds

src/databricks/labs/ucx/framework/tasks.py
```diff
@@ -37,7 +55,17 @@ def _remove_extra_indentation(doc: str) -> str:
...
+def task(
+    workflow,
+    *,
+    depends_on=None,
+    job_cluster="main",
+    notebook: str | None = None,
+    dashboard: str | None = None,
+    is_azure: bool = True,
+    is_aws: bool = True,
+    is_gcp: bool = True,
```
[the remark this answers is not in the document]
nfx: +1, let's add `cloud` keyword argument and do the logic based on that. having three boolean-true-by-default options will cause more issues in the future.

src/databricks/labs/ucx/framework/tasks.py
nfx: We'll be getting a Cloud enum from SDK soon - see
- https://github.com/databricks/databricks-sdk-py/pull/390
- https://github.com/databricks/databricks-sdk-go/pull/700

src/databricks/labs/ucx/framework/tasks.py
```diff
@@ -24,6 +26,22 @@ class Task:
...
+    def cloud_compatible(self, config: Config) -> bool:
+        """Test compatibility between config and task"""
+        return any(
+            [
+                config.is_aws == self.is_aws,
+                config.is_azure == self.is_azure,
+                config.is_gcp == self.is_gcp,
+                (self.is_aws is False and self.is_azure is False and self.is_gcp is False),
+            ]
+        )
```
nfx, line 43: this logic is brittle. don't use `any()`. use clear if statements. 

```
if self.cloud is None: return True
if self.cloud.lower() == 'aws' and cfg.is_aws: return True
if self.cloud.lower() == 'azure' and cfg.is_azure: return True
if self.cloud.lower() == 'gcp' and cfg.is_gcp: return True
return False
```

src/databricks/labs/ucx/runtime.py
nfx: fully agree with @william-conti . i'd suggest a string instead of boolean flags - that how we do it in go/openapi/spec, at least ;)

nfx (changes requested):
this logic is brittle. don't use `any()`. use clear if statements. 

```
if self.cloud is None: return True
if self.cloud.lower() == 'aws' and cfg.is_aws: return True
if self.cloud.lower() == 'azure' and cfg.is_azure: return True
if self.cloud.lower() == 'gcp' and cfg.is_gcp: return True
return False
```

We use the same approach in go/openapi/spec.

src/databricks/labs/ucx/runtime.py
dmoore247: @nfx I can't find the reference on the page.

dmoore247:
@nfx are we to block on another product update https://github.com/databrickslabs/ucx/pull/630#discussion_r1407634744 ?

Boolean properties were fine here in databricks-sdk...
image

And there was no objection to bools mentioned here: https://github.com/databrickslabs/ucx/pull/461#discussion_r1361693222

There are multiple ways of handling this simple signal variable and too many time varying opinions (@william-conti @FastLee @nfx) on the structure as evidenced from above [enums, bools, strings, arrays] and the multiplicity [one or many]

I think we lost sight of the goal that we have just two tasks at this time that are cloud specific.

I'm closing this PR, someone else will have to pick this up.

dmoore247:
See above

nfx:
@dmoore247 we're not blocking on other product update :) we're asking you to build the logic that scales

PR #632, opened 2023-11-27
Introduce `StaticTablesCrawler` for integration tests

### LINKED ISSUE #617: Add `StaticTablesCrawler` with `snapshot()` only returning the subset of tables specified in the init and use it in as many places possible to reduce the runtimes of the tests
the active part of `TablesCrawler` is done by the `tables.scala` notebook. `TablesCrawler` is heavily used in the integration tests to fetch the table metadata. We need to reduce the frequent operation of metadata fetching within tests to just returning a static list.

nfx:
Reduces removes 20 second of runtime on average from the integration tests involving `TablesCrawler`.

Fix #617

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -11,7 +11,11 @@ from databricks.sdk import AccountClient
 from databricks.sdk.core import Config
 from databricks.sdk.errors import NotFound
 from databricks.sdk.retries import retried
+from databricks.sdk.service.catalog import TableInfo
 
+from databricks.labs.ucx.framework.crawlers import SqlBackend
+from databricks.labs.ucx.hive_metastore import TablesCrawler
+from databricks.labs.ucx.hive_metastore.tables import Table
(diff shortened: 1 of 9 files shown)


tests/integration/conftest.py
```diff
@@ -104,3 +107,23 @@ def inner():
         return ws_group, acc_group
 
     return inner
+
+
+class StaticTablesCrawler(TablesCrawler):
```
william-conti, line 112: I'm confused, why do we introduce a specific Table Crawler instead of modifying the existing one ? This way, we'll be able to run UCX on a very small scope, running the assessment + group migration + data migration for one table / catalog

william-conti (changes requested):

tests/integration/conftest.py
nfx: it's specifically added for integration tests, following the same interface, so that it can be swapped

PR #635, opened 2023-11-27
Combined group migration tests

### LINKED ISSUE #616: Test `test_set_owner_permission` is effectively a duplicate of `test_replace_workspace_groups_with_account_groups` and has to be merged into one due to long runtimes
We need to reduce the number of tests that take more than 60 seconds to run.

Make this test more low-level

FastLee:
Closes #616


nfx:
already handled by https://github.com/databrickslabs/ucx/pull/636

PR #655, opened 2023-11-30
Added cluster policy selector in the installer

mirhendi:
With this PR, we can specify a cluster policy and create job clusters based on the specified policy


mirhendi:
fixes issue #656

src/databricks/labs/ucx/install.py
```diff
@@ -419,6 +419,12 @@ def warehouse_type(_):
                     )
                     instance_profile, spark_conf_dict = self._get_ext_hms_conf_from_policy(cluster_policy)
 
+        use_policy = self._question("Do you want to follow a policy to create clusters", default="no")
+        if use_policy != 'no':
+            custom_cluster_policy_id = self._question("Provide a cluster policy id")
```
nfx, line 424: Can you use _choices()?..

mirhendi: Added 👍

nfx (approved):

mirhendi:
Fixed formats and added unit test, please re-execute the workflow.

src/databricks/labs/ucx/install.py
```diff
@@ -482,6 +482,13 @@ def warehouse_type(_):
                     )
                     instance_profile, spark_conf_dict = self._get_ext_hms_conf_from_policy(cluster_policy)
 
+        use_policy = self._question("Do you want to follow a policy to create clusters?", default="no")
```
dmoore247, line 485: Too much code in this method, and module. Please extend with a component class. (see `configure.py`)

src/databricks/labs/ucx/install.py
```diff
@@ -971,6 +986,15 @@ def _get_cluster_policies_with_external_hive_metastores(self):
                     yield policy
                     break
 
+    def _check_policy_has_instance_pool(self, policy_id):
```
dmoore247, line 989: Too much code in this module, please extend with another module.

mirhendi: Same here, the logic is pretty simple, not sure how to break it

src/databricks/labs/ucx/install.py
```diff
@@ -971,6 +986,15 @@ def _get_cluster_policies_with_external_hive_metastores(self):
...
+        policy = self._ws.cluster_policies.get(policy_id=policy_id)
+        def_json = json.loads(policy.definition)
+        instance_pool = def_json.get("instance_pool_id")
+        if instance_pool is not None:
```
nfx, line 993:
```suggestion
        return instance_pool is not None
```
And if-else becomes redundant 😉

nfx (approved):

PR #659, opened 2023-12-01
Add retry on workspace listing

william-conti:
It may happend that we run into backend issues when listing workspace, goal of this PR is to catch these error and retry on them.

### DIFF
--- a/src/databricks/labs/ucx/workspace_access/listing.py
+++ b/src/databricks/labs/ucx/workspace_access/listing.py
@@ -5,7 +5,8 @@ from concurrent.futures import FIRST_COMPLETED, ThreadPoolExecutor, wait
 from itertools import groupby
 
 from databricks.sdk import WorkspaceClient
-from databricks.sdk.errors import NotFound
+from databricks.sdk.errors import NotFound, InternalError, DatabricksError
+from databricks.sdk.retries import retried
 from databricks.sdk.service.workspace import ObjectInfo, ObjectType
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/workspace_access/listing.py
```diff
@@ -40,9 +43,18 @@ def _progress_report(self, _):
                 f" rps: {rps:.3f}/sec"
             )
 
+    def _try_list_workspace(self, path: str) -> Iterator[ObjectType]:
+        try:
+            return self._ws.workspace.list(path=path, recursive=False)
+        except InternalError as e:
+            logger.warning(f"Had an InternalError when listing path {path}, retrying")
```
nfx, line 50: we're already doing debug logging on SDK level: https://github.com/databricks/databricks-sdk-py/blame/main/databricks/sdk/retries.py#L49, no need to double-log here

src/databricks/labs/ucx/workspace_access/listing.py
```diff
@@ -60,6 +72,9 @@ def _list_and_analyze(self, obj: ObjectInfo) -> (list[ObjectInfo], list[ObjectIn
         except NotFound:
             # See https://github.com/databrickslabs/ucx/issues/230
             logger.warning(f"{obj.path} is not listable. Ignoring")
+        except TimeoutError as e:
+            logger.warning(f"Could not list {obj.path} due to backend error. Ignoring")
+            logger.warning(e)
```
nfx, line 77:
```suggestion
            logger.warning(f"Could not list {obj.path} due to backend error: {e}")
```
don't do two log lines for one event.

nfx (changes requested):

nfx (approved):

Issue #638, opened 2023-11-27
Avoid getting members with SCIM list calls


xinjiezhen-db:
members as a group attribute can NOT be paginated as part of SCIM core RFCs. Since this is potentially an unbounded request if the members list is ever desired it'd be slightly better to retrieve that as part of a GET call instead of any list calls to reduce the response size as much as possible.

e.g.
https://github.com/databrickslabs/group-flow/blob/7819c08e0d990f223b3bafb641743b64689f96a7/account/databricks.go#L16

better to call list with "attributes=id,displayName" first, and loop through the list to get members.

andrascsillag-db:
thanks for the input @xinjiezhen-db 
Are there other group attributes that are not paginated?
So the most efficient way to get these attributes then from your perspective is to call list without these attributes and then call get in a loop?
cc @william-conti

xinjiezhen-db:
> thanks for the input @xinjiezhen-db Are there other group attributes that are not paginated? So the most efficient way to get these attributes then from your perspective is to call list without these attributes and then call get in a loop? cc @william-conti

No attributes can be paginated per core SCIM RFCs. So it's more about whether a multi value attribute can get too large.

For account SCIM `members` are the biggest problem, `groups` can cause problems too especially for large accounts. Workspace SCIM is a different beast and there are `entitlements` and `roles` but that's for another day.

And yes, it would be slightly better to iterate through the list and point get Groups with those attributes to reduce response size and server load, especially for large accounts.

nfx:
Done

PR #660, opened 2023-12-01
fixing groups list with large attributes

### LINKED ISSUE #625: Add pagination when listing workspace groups
In groups.py, there are multiple calls made on group listing that returns multiple attributes ("id,displayName,meta,members,roles,entitlements" as a minimum)

This call often raises a timeout when there is large amount of groups in the Workspace, this is because each calls returns the entire groups and a lot of attributes.

TODO:

- Identify all ws.groups.list calls that are possibly unsafe
- Add count and start_index to each call to ws.groups.list ([doc](https://databricks-sdk-py.readthedocs.io/en/latest/workspace/groups.html#GroupsAPI.list))
- The number should be a balance between reliability and performance (ie:50 as a starting point)
- Ideally, this number can be modified in the configuration file
- Add an integration test by provisioning 1000 groups

andrascsillag-db:
**Summary:**

This PR intends to fix the timeout on `groups.list` api that happens when large number of groups are fetched with attributes that can potentially get very large like `members`. 

Initially, the idea was to add pagination to the `groups.list `call but that won't help when the attributes get large, so the approach is to call groups.list with limited attributes first and then call groups.get for each group in a loop as suggested by  @xinjiezhen-db in #638. This will most probably slow down the whole list process but make it more reliable.

Fixes #625 and #638 

**Changes:**
- modified the `_list_workspace_groups` method to have a condition based on the attributes required
- added the rate limited `_safe_get_group` for when iterating over groups is required in `_list_workspace_groups`
- added unit tests

**Tests:**
- unit tests
- performance tested locally

### DIFF
--- a/src/databricks/labs/ucx/workspace_access/groups.py
+++ b/src/databricks/labs/ucx/workspace_access/groups.py
@@ -82,11 +82,13 @@ class GroupManager(CrawlerBase):
         inventory_database: str,
         include_group_names: list[str] | None = None,
         renamed_group_prefix: str = "ucx-renamed-",
+        verify_timeout: timedelta | None = timedelta(minutes=1),
     ):
         super().__init__(sql_backend, "hive_metastore", inventory_database, "groups", MigratedGroup)
         self._ws = ws
         self._include_group_names = include_group_names
         self._renamed_group_prefix = renamed_group_prefix
+        self._verify_timeout = verify_timeout
 
     def snapshot(self) -> list[MigratedGroup]:
         return self._snapshot(self._fetcher, self._crawler)
@@ -218,10 +220,13 @@ class GroupManager(CrawlerBase):
         attributes = scim_attributes.split(",")
         if "members" in attributes:
             attributes.remove("members")
-            for g in self._ws.groups.list(attributes="id,displayName,meta"):
+            for g in self._ws.groups.list(attributes=",".join(attributes)):
                 if self._is_group_out_of_scope(g, resource_type):
                     continue
-                group_with_all_attributes = self._get_group_with_retries(g.id)
+                # group_with_all_attributes = self._get_group_with_retries(g.id)
+                set_retry_on_value_error = retried(on=[InternalError], timeout=self._verify_timeout)
+                set_retried_check = set_retry_on_value_error(self._get_group_with_retries)
+                group_with_all_attributes = set_retried_check(g.id)
                 results.append(group_with_all_attributes)
         else:
             for g in self._ws.groups.list(attributes=scim_attributes):
@@ -231,9 +236,8 @@ class GroupManager(CrawlerBase):
         logger.info(f"Found {len(results)} {resource_type}")
         return results
 
-    @retried(on=[InternalError], timeout=timedelta(minutes=2))
     @rate_limited(max_requests=255, burst_period_seconds=60)
-    def _get_group_with_retries(self, group_id: str) -> iam.Group | None:
+    def _get_group_with_retries(self, group_id: str) -> iam.Group:
         return self._ws.groups.get(group_id)
 
     def _list_account_groups(self, scim_attributes: str) -> list[iam.Group]:
--- a/src/databricks/labs/ucx/workspace_access/groups.py
+++ b/src/databricks/labs/ucx/workspace_access/groups.py
@@ -3,6 +3,7 @@ import json
 import logging
 import typing
 from dataclasses import dataclass
+from datetime import timedelta
 
 from databricks.sdk import WorkspaceClient
 from databricks.sdk.core import DatabricksError
@@ -202,7 +203,9 @@ class GroupManager(CrawlerBase):
         return by_name
 
     def _is_group_out_of_scope(self, group: iam.Group, resource_type: str) -> bool:
-        if group.display_name in self._SYSTEM_GROUPS or group.meta.resource_type != resource_type:
+        if group.display_name in self._SYSTEM_GROUPS:
+            return True
+        if group.meta.resource_type != resource_type:
             return True
         return False
 
@@ -212,7 +215,9 @@ class GroupManager(CrawlerBase):
         # these attributes can get too large causing the api to timeout
         # so we're fetching groups without these attributes first
         # and then calling get on each of them to fetch all attributes
-        if "members" in scim_attributes or "roles" in scim_attributes or "entitlements" in scim_attributes:
+        attributes = scim_attributes.split(",")
+        if "members" in attributes:
+            attributes.remove("members")
             for g in self._ws.groups.list(attributes="id,displayName,meta"):
                 if self._is_group_out_of_scope(g, resource_type):
                     continue
@@ -226,7 +231,7 @@ class GroupManager(CrawlerBase):
         logger.info(f"Found {len(results)} {resource_type}")
         return results
 
-    @retried(on=[InternalError])
+    @retried(on=[InternalError], timeout=timedelta(minutes=2))
     @rate_limited(max_requests=255, burst_period_seconds=60)
     def _get_group_with_retries(self, group_id: str) -> iam.Group | None:
         return self._ws.groups.get(group_id)
--- a/src/databricks/labs/ucx/workspace_access/groups.py
+++ b/src/databricks/labs/ucx/workspace_access/groups.py
@@ -6,6 +6,7 @@ from dataclasses import dataclass
 
 from databricks.sdk import WorkspaceClient
 from databricks.sdk.core import DatabricksError
+from databricks.sdk.errors import InternalError
 from databricks.sdk.retries import retried
 from databricks.sdk.service import iam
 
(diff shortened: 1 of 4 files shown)


william-conti (changes requested):
missing scim.py

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -203,14 +204,34 @@ def _account_groups_in_account(self) -> dict[str, str]:
...
+            for g in self._ws.groups.list(attributes="id,displayName,meta"):
+                if g.display_name in self._SYSTEM_GROUPS:
+                    continue
+                if g.meta.resource_type != resource_type:
+                    continue
+                g = self._safe_get_group(g.id)
+                results.append(g)
+        else:
+            for g in self._ws.groups.list(attributes=scim_attributes):
+                if g.display_name in self._SYSTEM_GROUPS:
```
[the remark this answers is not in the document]
andrascsillag-db: fixed

andrascsillag-db:
thanks for the review comments. I fixed them, can you please recheck?

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -200,18 +201,36 @@ def _account_groups_in_account(self) -> dict[str, str]:
             by_name[g.display_name] = g.id
         return by_name
 
+    def _is_group_out_of_scope(self, group: iam.Group, resource_type: str) -> bool:
+        if group.display_name in self._SYSTEM_GROUPS or group.meta.resource_type != resource_type:
```
nfx, line 205: please make two `if` conditions - it'll be easier to support in long term.

src/databricks/labs/ucx/workspace_access/scim.py
```diff
@@ -36,7 +36,7 @@ def get_crawler_tasks(self):
     # TODO remove after ES-892977 is fixed
     @retried(on=[DatabricksError])
     def _get_groups(self):
-        return list(self._ws.groups.list(attributes="id,displayName,roles,entitlements"))
+        return list(self._list_workspace_groups(scim_attributes="id,displayName,roles,entitlements"))
```
nfx, line 39:
```suggestion
        return list(self._list_workspace_groups("id,displayName,roles,entitlements"))
```
single keyed argument doesn't need to be explicit

src/databricks/labs/ucx/workspace_access/scim.py
```diff
@@ -108,3 +108,24 @@ def _safe_get_group(self, group_id: str) -> Group | None:
         except NotFound:
             logger.warning(f"removed on backend: {group_id}")
             return None
+
+    @retried(on=[InternalError])
```
nfx, line 112: we can't afford unbounded retries. add a timeout, either in annotation or in configuration - see example in https://github.com/databrickslabs/ucx/pull/659/files

tests/unit/workspace_access/test_groups.py
```diff
@@ -132,25 +133,29 @@ def test_snapshot_should_consider_groups_defined_in_conf():
...
-            members=None,
-            external_id="11",
-            roles=None,
-            entitlements=None,
-        )
-    ]
+
+    gm = GroupManager(backend, wsclient, inventory_database="inv", include_group_names=["de"])
+
+    with patch.object(gm, "_list_workspace_groups", return_value=[group1, group2]):
```
nfx, line 144: don't mock out the code written in this repo - mock out the code that this repo depends upon.

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -203,14 +204,34 @@ def _account_groups_in_account(self) -> dict[str, str]:
...
-        for g in self._ws.groups.list(attributes=scim_attributes):
-            if g.display_name in self._SYSTEM_GROUPS:
-                continue
-            if g.meta.resource_type != resource_type:
-                continue
-            results.append(g)
+        # these attributes can get too large causing the api to timeout
+        # so we're fetching groups without these attributes first 
+        # and then calling get on each of them to fetch all attributes
+        if "members" in scim_attributes or "roles" in scim_attributes or "entitlements" in scim_attributes:
```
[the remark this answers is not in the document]
nfx: I'm not convinced we need to do the `roles` and `entitlement` here. keep only `members`.

src/databricks/labs/ucx/workspace_access/scim.py
```diff
@@ -108,3 +108,24 @@ def _safe_get_group(self, group_id: str) -> Group | None:
...
+
+    def _list_workspace_groups(self, scim_attributes: str) -> list[iam.Group]:
+        results = []
+        logger.info(f"Listing workspace groups with {scim_attributes}...")
+        # these attributes can get too large causing the api to timeout
+        # so we're fetching groups without these attributes first
+        # and then calling get on each of them to fetch all attributes
+        if "members" in scim_attributes or "roles" in scim_attributes or "entitlements" in scim_attributes:
+            for g in self._ws.groups.list(attributes="id,displayName,meta"):
+                group_with_all_attributes = self._get_group_with_retries(g.id)
```
nfx, line 125:
```suggestion
        attributes = scim_attributes.split(",")
        if "members" in attributes:
            attributes.remove("members") # i don't remember the syntax, but do this
            for g in self._ws.groups.list(attributes="id"):
                group_with_all_attributes = self._get_group_with_retries(g.id, scim_attributes)
```

1) don't operate on strings
2) if we predefined a list of attributes in the calling code, then your refactoring has to keep that logic.

src/databricks/labs/ucx/workspace_access/scim.py
```diff
@@ -108,3 +108,24 @@ def _safe_get_group(self, group_id: str) -> Group | None:
...
+                results.append(group_with_all_attributes)
+        else:
+            for g in self._ws.groups.list(attributes=scim_attributes):
+                results.append(g)
```
nfx, line 129:
```suggestion
            results = self._ws.groups.list(attributes=scim_attributes)
```
for loop is redundant.

make sure to cover this branch with unit tests.

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -200,18 +201,36 @@ def _account_groups_in_account(self) -> dict[str, str]:
...
+                results.append(group_with_all_attributes)
+        else:
+            for g in self._ws.groups.list(attributes=scim_attributes):
+                if self._is_group_out_of_scope(g, resource_type):
+                    continue
+                results.append(g)
         logger.info(f"Found {len(results)} {resource_type}")
         return results
 
+    @retried(on=[InternalError])
```
nfx, line 229: add timeout bounds to retries.

nfx (changes requested):

src/databricks/labs/ucx/workspace_access/scim.py
andrascsillag-db: Reverted all `scim.py` changes because if only `members` attribute is handled differently, then `scim.py` is not impacted by the change.

src/databricks/labs/ucx/workspace_access/scim.py
andrascsillag-db: Reverted all scim.py changes because if only members attribute is handled differently, then scim.py is not impacted by the change.

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -200,18 +202,40 @@ def _account_groups_in_account(self) -> dict[str, str]:
...
-            if g.meta.resource_type != resource_type:
-                continue
-            results.append(g)
+        # these attributes can get too large causing the api to timeout
+        # so we're fetching groups without these attributes first
+        # and then calling get on each of them to fetch all attributes
+        attributes = scim_attributes.split(",")
+        if "members" in attributes:
+            attributes.remove("members")
+            for g in self._ws.groups.list(attributes="id,displayName,meta"):
```
nfx, line 221:
```suggestion
            for g in self._ws.groups.list(attributes=",".join(attributes)):
```
otherwise it's a bug

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -200,18 +202,40 @@ def _account_groups_in_account(self) -> dict[str, str]:
...
+                results.append(group_with_all_attributes)
+        else:
+            for g in self._ws.groups.list(attributes=scim_attributes):
+                if self._is_group_out_of_scope(g, resource_type):
+                    continue
+                results.append(g)
         logger.info(f"Found {len(results)} {resource_type}")
         return results
 
+    @retried(on=[InternalError], timeout=timedelta(minutes=2))
```
nfx, line 234: does this increase the runtime of unit tests? if it does, make it configurable for unit tests.

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -200,18 +202,40 @@ def _account_groups_in_account(self) -> dict[str, str]:
...
+    @rate_limited(max_requests=255, burst_period_seconds=60)
+    def _get_group_with_retries(self, group_id: str) -> iam.Group | None:
```
nfx, line 236:
```suggestion
    def _get_group_with_retries(self, group_id: str) -> iam.Group:
```

it can't have `None` and we're already throwing `NotFound`.

nfx (changes requested):
few fixes remaining. please make sure integration tests do pass

src/databricks/labs/ucx/workspace_access/groups.py
andrascsillag-db: fixed

PR #661, opened 2023-12-02
Changed cluster security mode from NONE to LEGACY_SINGLE_USER, as `crawl_tables` was failing when run on non-UC Workspace in No Isolation mode with unable to access the config file

HariGS-DB:
This PR contains 3 changes:
1. updating logger name in uninstall.py to databricks.labs.ucx.install so that it installs the install logs when running ucx uninstall
2. made fix to the test_installation.py test_jobs_with_no_inventory_database. if there multiple grants for a principal, it might come in different order resulting in assert test falsly failing. added sort dict to sort the dict values before comparing.
3. changing the cluster security mode from DataSecurityMode.NONE to DataSecurityMode. LEGACY_SINGLE_USER. The scala code is failing when run on Non UC Workspace in No Isolation mode with unable to access the config file. Changing the SecurityMode fixes it.

### DIFF
--- a/tests/integration/test_installation.py
+++ b/tests/integration/test_installation.py
@@ -210,8 +210,8 @@ def test_jobs_with_no_inventory_database(
             assert sort_dict(schema_a_grants) == sort_dict(src_schema_a_grants)
             assert sort_dict(schema_b_grants) == sort_dict(src_schema_b_grants)
             assert (
-                    sort_dict(schema_default_grants)[ws_group_c.display_name]
-                    == sort_dict(src_schema_default_grants)[ws_group_c.display_name]
+                sort_dict(schema_default_grants)[ws_group_c.display_name]
+                == sort_dict(src_schema_default_grants)[ws_group_c.display_name]
             )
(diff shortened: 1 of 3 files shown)


HariGS-DB:
[TestResults.pdf](https://github.com/databrickslabs/ucx/files/13537365/TestResults.pdf)

HariGS-DB:
https://docs.google.com/document/d/1yL9i8ti0h0J4Be2O6GbEGKgRUTwygrF4sw5Dy0gU3os/edit

tests/integration/test_installation.py
```diff
@@ -192,6 +192,10 @@ def validate_permissions():
         @retried(on=[AssertionError], timeout=timedelta(minutes=2))
         def validate_tacl():
             logger.info("validating tacl")
+
+            def sort_dict(obj: dict):
+                return {k: sorted(v) for k, v in obj.items()}
```
nfx, line 197: It's a set, it doesn't require sorting

PR #679, opened 2023-12-06
Added `databricks labs ucx installations` command

nfx:
Detects concurrent installations of UCX by multiple users


pritishpai (commented):
LGTM. Just needs test cov.

pritishpai (approved):

PR #681, opened 2023-12-07
Added filters to run only cloud specific task

### LINKED ISSUE #460: Installation and assessment need to differentiate between clouds and cloud specific tasks
Currently there is no differentiation in the:
- installation
- assessment job
- dashboard
for tasks that are cloud specific. Add logic to include only applicable tasks.

image

prajin-29:
may close #460

### DIFF
--- a/src/databricks/labs/ucx/framework/tasks.py
+++ b/src/databricks/labs/ucx/framework/tasks.py
@@ -5,6 +5,8 @@ from functools import wraps
 from logging.handlers import TimedRotatingFileHandler
 from pathlib import Path
 
+from databricks.sdk.core import Config
+
 from databricks.labs.ucx.__about__ import __version__
 from databricks.labs.ucx.config import WorkspaceConfig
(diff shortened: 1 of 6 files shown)


src/databricks/labs/ucx/framework/tasks.py
```diff
@@ -23,6 +25,21 @@ class Task:
     job_cluster: str = "main"
     notebook: str = None
     dashboard: str = None
+    # if all are False then task is for all clouds
+    is_aws: bool = False
+    is_azure: bool = False
+    is_gcp: bool = False
```
nfx, line 31: use only one `cloud: str` field

nfx (changes requested):

src/databricks/labs/ucx/framework/tasks.py
prajin-29: updated this part to cloud:str

nfx (approved):

PR #705, opened 2023-12-19
Filtering out inventory database from loading into tables and filtering out the same from grant detail view

### LINKED ISSUE #665: Assessment job/dashboard reports on tables it creates during the assessment process
Appears that 13 tables get created in a schema called 'yes during the assessment job.

Currently, UCX will assess the readiness of these tables to upgrade to UC, which doesn't really make sense.

Screenshot 2023-12-04 at 11 07 05 AM

prajin-29:
Filtering out inventory database from loading into tables and filtering out the same from grant detail view.
may close #665


src/databricks/labs/ucx/hive_metastore/tables.scala
```diff
@@ -79,7 +79,7 @@ def getInventoryDatabase(): String={
 }
 
 val inventoryDatabase = getInventoryDatabase()
-var df = metadataForAllTables(spark.sharedState.externalCatalog.listDatabases(), failures)
+var df = metadataForAllTables(spark.sharedState.externalCatalog.listDatabases().filter(_ != s"$inventoryDatabase"), failures)
```
nfx, line 82:
```suggestion
var df = metadataForAllTables(spark.sharedState.externalCatalog.listDatabases().filter(_ != inventoryDatabase), failures)
```
check if this works

src/databricks/labs/ucx/queries/views/grant_detail.sql
```diff
@@ -29,4 +29,4 @@ SELECT
     catalog,
     database,
     table
-FROM $inventory.grants
\ No newline at end of file
+FROM $inventory.grants where database != split("$inventory",'[.]')[1]
```
nfx, line 32: are you sure this is the right logic? $inventory won't have dots

prajin-29: inventory comes with "hive_metastore.database_name" here. I tested the same.

src/databricks/labs/ucx/hive_metastore/tables.scala
prajin-29: Tested the same and the table was not having the inventory database.Working as expected

nfx: why do you need to interpolate a string within another string? `inventoryDatabase` is a string already.

nfx (approved):

PR #725, opened 2023-12-22
Allow specifying a group list when group match options are used

### LINKED ISSUE #721: Allow group mapping and group list selection on the installer
The current options on the installer allows you to select table mapping or apply a list of groups.
These options shouldn't be mutually exclusive.
You should be able to apply mapping AND select a list of groups to be migrated.

FastLee:
closes #721 

Added option to select a group list when one of the group matching option is used.
Added unit tests.


tests/unit/test_install.py
```diff
@@ -1002,3 +1002,232 @@ def test_uninstall_no_config_file(ws, mocker):
     ws.workspace.download = lambda _: io.BytesIO(config_bytes)
     ws.workspace.get_status.side_effect = NotFound(...)
     install.uninstall()
+
+
+def test_save_config_with_prefix(ws):
```
nfx, line 1007: The logic is in the ConfigureGroups class, move tests there. It doesn't belong to installer, which is one level above configure groups logic

WorkspaceInstaller does too much already and it's difficult to support unit tests that create a config file.

tests/unit/workspace_access/test_groups.py
```diff
@@ -686,15 +686,95 @@ def test_snapshot_with_group_matched_by_external_id():
...
+
+
+def test_configure_match():
+    cg = ConfigureGroups(
+        MockPrompts(
+            {
+                "Backup prefix": "",
+                r"Choose how to map the workspace groups.*": "5",  # partial match
+                r".*workspace group$": r"\[(#+)\]",
+                r".*account group$": r"\((#+)\)",
```
nfx, line 773: This will break if we add color decoration to prompts, and we won't be able to detect the end of a string

nfx (changes requested):

tests/unit/test_install.py
FastLee: Done

nfx (approved):

nfx (approved):

PR #735, opened 2023-12-23
Verify metastore exists in current workspace

### LINKED ISSUE #104: Verify that metastore has been configured on account

pritishpai:
Resolved #104

### DIFF
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -90,20 +90,6 @@ def create_table_mapping():
     webbrowser.open(f"{ws.config.host}/#workspace{path}")
 
 
-def verify_metastore():
-    """
-    Scans and verifies if a metastore exists for a metastore
-    :param cfg:
-    :return:
-    """
-    ws = WorkspaceClient()
-    current_metastore = ws.metastores.current()
-    if current_metastore:
-        print(f"Metastore {current_metastore.default_catalog_name} with {current_metastore.metastore_id} exists")
-    else:
-        print("Metastore does not exist")
-
-
 def validate_external_locations():
     ws = WorkspaceClient()
     prompts = Prompts()
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -92,7 +92,7 @@ def create_table_mapping():
 
 def verify_metastore():
     """
-    Scans and verifies if a metastore esists for a metostore
+    Scans and verifies if a metastore exists for a metastore
     :param cfg:
     :return:
     """
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -2,12 +2,8 @@ import json
 import logging
 import sys
 import webbrowser
-from typing import Optional
 
 from databricks.sdk import WorkspaceClient
-
-from databricks.sdk.errors import NotFound
-from databricks.sdk.service.catalog import MetastoreAssignment
 
 from databricks.labs.ucx.account import AccountWorkspaces, WorkspaceInfo
 from databricks.labs.ucx.config import AccountConfig, ConnectConfig
@@ -18,7 +14,6 @@ from databricks.labs.ucx.hive_metastore.mapping import TableMapping
 from databricks.labs.ucx.hive_metastore.table_migrate import TablesMigrate
 from databricks.labs.ucx.install import WorkspaceInstaller
 from databricks.labs.ucx.installer import InstallationManager
-from tests.unit.test_install import ws
 
 logger = logging.getLogger("databricks.labs.ucx")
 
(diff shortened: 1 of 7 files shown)


src/databricks/labs/ucx/cli.py
```diff
@@ -89,6 +89,20 @@ def create_table_mapping():
     webbrowser.open(f"{ws.config.host}/#workspace{path}")
 
 
+def verify_metastore():
+    """
+    Scans and verifies if a metastore exists for a metastore
+    :param cfg:
+    :return:
+    """
+    ws = WorkspaceClient()
```
nfx, line 98: please extract it as a `VerifyHasMetastore` utility class, that raises a readable exception if metastore is not assigned and call it from:

1. https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/runtime.py#L237-L238
2. https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/hive_metastore/tables.py#L207-L208

there's no need for a CLI command

nfx (changes requested):

src/databricks/labs/ucx/runtime.py
```diff
@@ -238,6 +242,12 @@ def rename_workspace_local_groups(cfg: WorkspaceConfig):
     """Renames workspace local groups by adding `ucx-renamed-` prefix."""
     sql_backend = RuntimeBackend()
     ws = WorkspaceClient(config=cfg.to_databricks_config())
+    verify_has_metastore = VerifyHasMetastore(ws)
+    try:
+        if verify_has_metastore.verify_metastore():
+            logger.info("Metastore exists in the workspace")
+    except MetastoreNotFoundError as e:
```
nfx, line 249: Don't catch it - task has to fail with this message

src/databricks/labs/ucx/workspace_access/verification.py
```diff
@@ -79,3 +79,43 @@ def verify_roles_and_entitlements(self, migration_state: MigrationState, target:
...
+            self.metastore_id = current_metastore.metastore_id
+            self.workspace_id = current_metastore.workspace_id
+            return True
+
+        return False
+
+
+class MetastoreNotFoundError(Exception):
+    def __init__(self, message="Metastore not found in the workspace"):
+        self.message = message
```
nfx, line 120: It's not necessary - exceptions can be turned to strings easily

src/databricks/labs/ucx/workspace_access/verification.py
```diff
@@ -79,3 +79,43 @@ def verify_roles_and_entitlements(self, migration_state: MigrationState, target:
...
+        Verifies if a metastore exists for a metastore
+        :param cfg:
+        :return:
+        """
+        metastore_exists = self.check_metastore_existence()
+        if not metastore_exists:
+            raise MetastoreNotFoundError()
+        return True
+
+    def check_metastore_existence(self) -> bool:
```
nfx, line 102: It's not used


Remove this method, it increases the surface area, that we don't need now

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -211,6 +212,7 @@ def __init__(
         self._database_to_catalog_mapping = database_to_catalog_mapping
         self._default_catalog = self._init_default_catalog(default_catalog)
         self._seen_tables: dict[str, str] = {}
+        self._verify_has_metastore = VerifyHasMetastore(ws)
```
nfx, line 215: Why it's not called?

nfx (changes requested):

src/databricks/labs/ucx/workspace_access/verification.py
pritishpai: It is used in the verify_metastore method and also unit tested (if that is what you meant by surface area)

src/databricks/labs/ucx/hive_metastore/tables.py
pritishpai: Done!

src/databricks/labs/ucx/workspace_access/verification.py
nfx: it's not used anywhere else

nfx (approved):

src/databricks/labs/ucx/workspace_access/verification.py
pritishpai: Removing it with the fix for failing integration tests which was because of not handling the non-UC env permissions error.

src/databricks/labs/ucx/workspace_access/verification.py
```diff
@@ -1,10 +1,16 @@
...
 from databricks.sdk import WorkspaceClient
+from databricks.sdk.errors import PermissionDenied
 
 from databricks.labs.ucx.workspace_access.groups import MigrationState
 from databricks.labs.ucx.workspace_access.secrets import SecretScopesSupport
 
+logging.basicConfig()
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
```
nfx, line 13:
```suggestion
```

PR #745, opened 2024-01-02
Address Table mapping options and migration types including DBFS root tables.

### LINKED ISSUE #332: [FEATURE]: Connect `_migrate_table_create_ctas` to `migrate-tables` workflow
most of the code is done in:
- https://github.com/databrickslabs/ucx/pull/1434

Diff:
https://github.com/databrickslabs/ucx/blob/9cb9dbf05ba26d6905a53ebbf964f25ce1bddc09/src/databricks/labs/ucx/hive_metastore/table_migrate.py#L286-L299

Tasks:
- [x] Copy this table using CTAS
- [ ] Copy the Metadata with `CREATE TABLE ... LIKE ...`
- [ ] Skip tables as needed based on size threshold or an exception list
- [x] Update the source table with `upgraded_to` property
- [ ] Link it to the workflow as task

FastLee:
Closes #332


nfx:
Please create two separate pull requests - one with moving the files and other with new features, otherwise it's impossible to review.

Issue #761, opened 2024-01-08
Reflect account group to workspace step fails with ResourceDoesNotExist


mwojtyczka:
See screenshot
![Screenshot 2024-01-08 at 10 33 34](https://github.com/databrickslabs/ucx/assets/15330825/23af4a87-6ab9-40ee-b07b-56d9a5b92e64)

PR #762, opened 2024-01-08
Improve error handling by not failing group rename step if a group was removed from account before reflecting it to workspace

mwojtyczka:
- Fix isssue [761](https://github.com/databrickslabs/ucx/issues/761)

In rare cases, it can happen that a group is removed from the account before we manage to reflect it in the workspace. It's usually a couple of minutes between listing the groups and reflecting it. This PR will log such cases without failing the tool.

### DIFF
--- a/src/databricks/labs/ucx/workspace_access/groups.py
+++ b/src/databricks/labs/ucx/workspace_access/groups.py
@@ -18,7 +18,6 @@ from databricks.sdk.errors.mapping import (
     InternalError,
     NotFound,
     ResourceConflict,
-    ResourceDoesNotExist,
 )
 from databricks.sdk.retries import retried
 from databricks.sdk.service import iam
--- a/src/databricks/labs/ucx/workspace_access/groups.py
+++ b/src/databricks/labs/ucx/workspace_access/groups.py
@@ -508,7 +508,7 @@ class GroupManager(CrawlerBase[MigratedGroup]):
         except BadRequest:
             # already exists
             return True
-        except ResourceDoesNotExist:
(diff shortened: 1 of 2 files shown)


tests/unit/workspace_access/test_groups.py
```diff
@@ -380,6 +380,26 @@ def do_side_effect(*args, **kwargs):
...
+                "Resources": [g.as_dict() for g in [account_group1]],
+            }
+        if method == "PUT":
+            raise ResourceDoesNotExist(
+                "The group has been removed from the Databricks account after getting the group "
+                "and before reflecting it to the workspace."
+            )
+
+    wsclient.api_client.do.side_effect = reflect_account_side_effect
+    GroupManager(backend, wsclient, inventory_database="inv").reflect_account_groups_on_workspace()
```
william-conti, line 400: Please add some assertions there, something like:

wsclient.api_client.do.assert_called_with(xxx)

william-conti (changes requested):
LGTM, just a minor change on the test

william-conti (approved):

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -507,6 +508,10 @@ def _reflect_account_group_to_workspace(self, account_group_id: str):
         except BadRequest:
             # already exists
             return True
+        except ResourceDoesNotExist:
```
nfx, line 511:
```suggestion
        except NotFound:
```
use wider exception

nfx (approved):

Issue #649, opened 2023-11-28
Catch groups that doesn't have the same membership on Workspace and Account level


william-conti:
In groups.py, when we find a corresponding group in the Account console we do not check if the number of Users is the same.
We should log the Account groups that have the same name as a Workspace group, but a different membership

nfx:
this is a duplicate of some of the issues, though not remembering which ones. can you add test cases?

Issue #17, opened 2023-06-05
Add ability to create account groups with members from the workspace group with members


HariGS-DB:

zpappa:
Can you please explain this in detail?

HariGS-DB:
I think this refers to adding a group in account if it doesnt exist but exist in workspace , its covered in scope of issue 343?

Issue #629, opened 2023-11-27
`make dev` is failing with `pyenv` installed


dmoore247:
a fresh git clone as of 10 minutes ago.
a fresh pyenv

```shell
% make dev
if ! command -v hatch &> /dev/null; then pip install hatch ; fi
hatch env create
pyenv: hatch: command not found

The `hatch' command exists in these Python versions:
  3.10.10/envs/zzz
  zzz

Note: See 'pyenv help global' for tips on allowing both
      python2 and python3 to be found.
make: *** [.venv/bin/python] Error 127
```

dmoore247:
manually installing `pip install hatch` then `make test` will run tests until they stall... (hang)...
```
[gw0] PASSED tests/unit/test_install.py::test_replace_clusters_for_integration_tests 
tests/unit/test_install.py::test_run_workflow_creates_proper_failure 
[gw0] PASSED tests/unit/test_install.py::test_run_workflow_creates_proper_failure 
tests/unit/test_install.py::test_install_database_happy 
[gw0] PASSED tests/unit/test_install.py::test_install_database_happy 
tests/unit/test_install.py::test_install_database_unhappy 
[gw0] PASSED tests/unit/test_install.py::test_install_database_unhappy 
```

PR #763, opened 2024-01-08
Add a command to create account level groups if they do not exist

william-conti:
Attempt to fix 
- https://github.com/databrickslabs/ucx/issues/17
- https://github.com/databrickslabs/ucx/issues/649

Adds a command to create groups at account level by crawling all workspaces configured in the account and in scope of the migration

This pull request adds several new methods to the `account.py` file in the `databricks/labs/ucx` directory. The main method added is `create_account_level_groups`, which crawls all workspaces in an account and creates account-level groups if a workspace-local group is not present in the account. The method `get_valid_workspaces_groups` is added to retrieve a dictionary of all valid workspace groups, while `has_not_same_members` checks if two groups have the same members. The method `get_account_groups` retrieves a dictionary of all account groups.

Regarding the tests, the `test_account.py` file has been updated to include new tests for the `create_account_level_groups` method. The test `test_create_acc_groups_should_create_acc_group_if_no_group_found` verifies that an account-level group is created if no group with the same name is found. The test `test_create_acc_groups_should_filter_groups_in_other_workspaces` checks that the method filters groups present in other workspaces and only creates groups that are not present in the account.

Additionally, the `cli.py` file has been updated to include a new command, `create_account_level_groups`, which uploads workspace config to all workspaces in the account where ucx is installed.


src/databricks/labs/ucx/account.py
```diff
@@ -88,6 +88,42 @@ def sync_workspace_info(self):
...
+
+        workspace_clients = self.workspace_clients()
+        all_workspace_groups = {}
+        for client in workspace_clients:
+            ws_groups_ids = client.groups.list(attributes="id")
+            for grp_id in ws_groups_ids:
+                full_workpace_group = client.groups.get(grp_id.id)
+                if full_workpace_group.display_name in all_workspace_groups:
+                    if len(all_workspace_groups[full_workpace_group.display_name]) == full_workpace_group.members:
+                        logger.debug(f"Group {full_workpace_group.display_name} already found in another workspace")
```
nfx, line 109: 6-tab levels of nesting is too much and hard to support/read. please refactor

src/databricks/labs/ucx/account.py
```diff
@@ -88,6 +88,42 @@ def sync_workspace_info(self):
...
+        """
+        acc_groups = {}
+        ac_grp_ids = self._ac.groups.list(attributes="id")
+        for acc_grp_id in ac_grp_ids:
+            full_account_group = self._ac.groups.get(acc_grp_id.id)
+            acc_groups[full_account_group.display_name] = full_account_group.members
+
+        workspace_clients = self.workspace_clients()
+        all_workspace_groups = {}
+        for client in workspace_clients:
```
nfx, line 103: inline workspace_clients, it's used only once

src/databricks/labs/ucx/account.py
```diff
@@ -88,6 +88,42 @@ def sync_workspace_info(self):
                 path = f"{installation.path}/{self.SYNC_FILE_NAME}"
                 ws.workspace.upload(path, info, overwrite=True, format=ImportFormat.AUTO)
 
+    def create_account_level_groups(self):
+        """
+        Crawl all workspaces, and create account level groups if a WS local group is not present in the account
+        """
+        acc_groups = {}
+        ac_grp_ids = self._ac.groups.list(attributes="id")
```
nfx, line 96: don't use abbreviations, at least for groups - it's two more characters and makes it more consistent

src/databricks/labs/ucx/account.py
```diff
@@ -88,6 +88,42 @@ def sync_workspace_info(self):
...
+            acc_groups[full_account_group.display_name] = full_account_group.members
+
+        workspace_clients = self.workspace_clients()
+        all_workspace_groups = {}
+        for client in workspace_clients:
+            ws_groups_ids = client.groups.list(attributes="id")
+            for grp_id in ws_groups_ids:
+                full_workpace_group = client.groups.get(grp_id.id)
+                if full_workpace_group.display_name in all_workspace_groups:
+                    if len(all_workspace_groups[full_workpace_group.display_name]) == full_workpace_group.members:
```
nfx, line 108: using len for comparison is a faulty assumption - compare based on usernames.

src/databricks/labs/ucx/account.py
```diff
@@ -88,6 +88,42 @@ def sync_workspace_info(self):
...
+            full_account_group = self._ac.groups.get(acc_grp_id.id)
+            acc_groups[full_account_group.display_name] = full_account_group.members
+
+        workspace_clients = self.workspace_clients()
+        all_workspace_groups = {}
+        for client in workspace_clients:
+            ws_groups_ids = client.groups.list(attributes="id")
+            for grp_id in ws_groups_ids:
+                full_workpace_group = client.groups.get(grp_id.id)
+                if full_workpace_group.display_name in all_workspace_groups:
```
nfx, line 107: what if group A exists in workspace B and C, but has different members?

nfx (changes requested):
logic is incorrect and has hidden bugs

src/databricks/labs/ucx/account.py
william-conti: IMO the account group should be the single source of truth. 

If acc group has 3 members, ws B group has 10 members and ws C group has 5 members, we'll log two times that the group isn't consistent at all.

src/databricks/labs/ucx/account.py
```diff
@@ -88,6 +89,59 @@ def sync_workspace_info(self):
...
+        all_valid_workspace_groups = self.get_valid_workspaces_groups()
+
+        for group_name, group in all_valid_workspace_groups.items():
+            if group_name in acc_groups:
+                logger.info(f"Group {group_name} already exist in the account, ignoring")
+            else:
+                self._ac.groups.create(display_name=group_name, members=group.members)
+                logger.info(f"Group {group_name} created in the account")
+
+    def get_valid_workspaces_groups(self) -> dict[str, Group]:
```
nfx, line 107: make method private

src/databricks/labs/ucx/account.py
```diff
@@ -88,6 +89,59 @@ def sync_workspace_info(self):
...
+                        inconsistent_groups.append(group_name)
+                        all_workspaces_groups.pop(group_name)
+                    else:
+                        logger.info(f"Workspace group {group_name} already found, ignoring")
+                else:
+                    logger.info(f"Found new group {group_name}")
+                    all_workspaces_groups[group_name] = full_workspace_group
+        return all_workspaces_groups
+
+    def has_not_same_members(self, group_1: Group, group_2: Group) -> bool:
```
nfx, line 133: make private

src/databricks/labs/ucx/account.py
```diff
@@ -88,6 +89,59 @@ def sync_workspace_info(self):
...
+        ws_members_set = set([m.display for m in group_1.members] if group_1.members else [])
+        ws_members_set_2 = set([m.display for m in group_2.members] if group_2.members else [])
+        return bool((ws_members_set - ws_members_set_2).union(ws_members_set_2 - ws_members_set))
+
+    def get_account_groups(self) -> dict[str | None, list[ComplexValue] | None]:
```
nfx, line 138: make private

nfx (changes requested):
mostly lgtm, please add more unit tests

prajin-29:
Should we add integration test to it?

src/databricks/labs/ucx/account.py
```diff
@@ -88,6 +89,59 @@ def sync_workspace_info(self):
...
+                logger.info(f"Group {group_name} created in the account")
+
+    def get_valid_workspaces_groups(self) -> dict[str, Group]:
+        all_workspaces_groups: dict[str, Group] = {}
+        inconsistent_groups: list[str] = []
+        for client in self.workspace_clients():
+            ws_group_ids = client.groups.list(attributes="id")
+            for group_id in ws_group_ids:
+                full_workspace_group = client.groups.get(group_id.id)
+                group_name = full_workspace_group.display_name
```
nfx, line 114: If two groups in two different workspaces have the same name, but different members, add workspace name as a prefix and create group with those members on the account

tests/unit/test_account.py
```diff
@@ -98,3 +98,98 @@ def test_manual_workspace_info(mocker):
...
+    acc_client.config = account_config.to_databricks_config()
+
+    account_config.to_account_client = lambda: acc_client
+    account_config.include_workspace_names = ["foo"]
+
+    acc_client.workspaces.list.return_value = [
+        Workspace(workspace_name="foo", workspace_id=123, workspace_status_message="Running", deployment_name="abc")
+    ]
+
+    mock1 = MagicMock()
```
nfx, line 118: Use create_autospec(WorkspaceClient)

tests/unit/test_account.py
```diff
@@ -98,3 +98,98 @@ def test_manual_workspace_info(mocker):
...
+
+    account_config.to_account_client = lambda: acc_client
+    account_config.include_workspace_names = ["foo", "bar"]
+
+    acc_client.workspaces.list.return_value = [
+        Workspace(workspace_name="foo", workspace_id=123, workspace_status_message="Running", deployment_name="abc"),
+        Workspace(workspace_name="bar", workspace_id=456, workspace_status_message="Running", deployment_name="def"),
+    ]
+
+    mock1 = MagicMock()
```
nfx, line 160: Create autospec

nfx (changes requested):
This PR won't solve the #629. It's a different use-case

src/databricks/labs/ucx/account.py
```diff
@@ -87,6 +88,57 @@ def sync_workspace_info(self):
...
+        """
+        acc_groups = self._get_account_groups()
+
+        all_valid_workspace_groups = self._get_valid_workspaces_groups()
+
+        for group_name, group in all_valid_workspace_groups.items():
+            if group_name in acc_groups:
+                logger.info(f"Group {group_name} already exist in the account, ignoring")
+            else:
+                self._ac.groups.create(display_name=group_name, members=group.members)
```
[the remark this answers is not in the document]
william-conti: Thanks for the review @xinjiezhen-db

src/databricks/labs/ucx/account.py
```diff
@@ -6,6 +6,8 @@
 from databricks.labs.blueprint.tui import Prompts
 from databricks.sdk import AccountClient, WorkspaceClient
 from databricks.sdk.errors import NotFound
+from databricks.sdk.service import iam
```
nfx, line 9: Looks redundant - see the following line

src/databricks/labs/ucx/account.py
```diff
@@ -69,6 +71,77 @@ def sync_workspace_info(self):
...
                 installation.save(workspaces, filename=self.SYNC_FILE_NAME)
 
+    def create_account_level_groups(self):
+        acc_groups = self._get_account_groups()
+        all_valid_workspace_groups = self._get_valid_workspaces_groups()
+
+        for group_name, valid_group in all_valid_workspace_groups.items():
+            if group_name in acc_groups:
+                logger.info(f"Group {group_name} already exist in the account, ignoring")
+            else:
```
nfx, line 81: Replace else with continue and reduce nesting

src/databricks/labs/ucx/account.py
```diff
@@ -69,6 +71,77 @@ def sync_workspace_info(self):
...
+                    self._add_members_to_acc_group(acc_group.id, group_name, valid_group)
+
+                logger.info(f"Group {group_name} created in the account")
+
+    def _add_members_to_acc_group(self, acc_group_id: str, group_name: str, valid_group: Group):
+        for chunk in self._chunks(valid_group.members, 20):
+            logger.debug(f"Adding 20 members to acc group {group_name}")
+            self._ac.groups.patch(
+                acc_group_id,
+                operations=[iam.Patch(op=iam.PatchOp.ADD, path="members", value=[x.as_dict() for x in chunk])],
```
nfx, line 94: Would love to see a demo of this 😉

src/databricks/labs/ucx/account.py
```diff
@@ -69,6 +71,77 @@ def sync_workspace_info(self):
...
+                group_name = full_workspace_group.display_name
+
+                if group_name in all_workspaces_groups:
+                    if self._has_not_same_members(all_workspaces_groups[group_name], full_workspace_group):
+                        logger.warning(
+                            f"Group {group_name} does not have same amount of members "
+                            f"in workspace {client.config.host}, it will be created with account "
+                            f"name {client.config.host}_{group_name}"
+                        )
+                        all_workspaces_groups[f"{client.config.host}_{group_name}"] = full_workspace_group
```
nfx, line 121: Can you use workspace name instead of the hostname? See account.py

I also think that you should prompts.confirm() that with the user

src/databricks/labs/ucx/account.py
```diff
@@ -69,6 +71,77 @@ def sync_workspace_info(self):
...
+                    else:
```
nfx, line 122: If not...: continue

src/databricks/labs/ucx/account.py
```diff
@@ -69,6 +71,77 @@ def sync_workspace_info(self):
...
+                        logger.info(f"Workspace group {group_name} already found, ignoring")
+                else:
```
nfx, line 124: Unreadable, reduce nesting, if not ...: continue

src/databricks/labs/ucx/account.py
```diff
@@ -69,6 +71,77 @@ def sync_workspace_info(self):
...
+                    logger.info(f"Found new group {group_name}")
+                    if not group_name:
+                        continue
+                    all_workspaces_groups[group_name] = full_workspace_group
+        return all_workspaces_groups
+
+    def _has_not_same_members(self, group_1: Group, group_2: Group) -> bool:
+        ws_members_set = set([m.display for m in group_1.members] if group_1.members else [])
```
nfx, line 132:
```suggestion
        ws_members_set_1 = set([m.display for m in group_1.members] if group_1.members else [])
```
Be consistent 😉

tests/unit/test_account.py
```diff
@@ -68,3 +69,201 @@ def test_manual_workspace_info(mocker):
...
+    acc_client.groups.create.assert_called_with(
+        display_name="de",
+    )
+    acc_client.groups.patch.assert_called_with(
+        "12",
+        operations=[
+            iam.Patch(
+                op=iam.PatchOp.ADD,
+                path='members',
+                value=[{'display': 'test-user-1', 'value': '20'}, {'display': 'test-user-2', 'value': '21'}],
```
nfx, line 110: Isn't just value enough there?

nfx (changes requested):
Almost there

src/databricks/labs/ucx/account.py
william-conti: We should make a separate to PR to handle multiple scenarios, let's create a default way to handle this first. Workspace name is indeed much better

src/databricks/labs/ucx/account.py
william-conti: What do you mean there ? This code actually comes from test_performance.py

tests/unit/test_account.py
william-conti: Test will fail, have to specify all parameters ...

src/databricks/labs/ucx/account.py
```diff
@@ -69,6 +70,84 @@ def sync_workspace_info(self):
...
+            acc_group = self._ac.groups.create(display_name=group_name)
+
+            if len(acc_group.members) > 0:
+                self._add_members_to_acc_group(acc_group.id, group_name, valid_group)
+
+            logger.info(f"Group {group_name} created in the account")
+
+    def _add_members_to_acc_group(self, acc_group_id: str, group_name: str, valid_group: Group):
+        for chunk in self._chunks(valid_group.members, 20):
+            logger.debug(f"Adding 20 members to acc group {group_name}")
```
nfx, line 91: Nit: what about the last chunk, which has less than 20 members?

src/databricks/labs/ucx/account.py
```diff
@@ -69,6 +70,84 @@ def sync_workspace_info(self):
...
+
+                full_workspace_group = client.groups.get(group_id.id)
+                group_name = full_workspace_group.display_name
+
+                if group_name in all_workspaces_groups:
+                    if self._has_same_members(all_workspaces_groups[group_name], full_workspace_group):
+                        logger.info(f"Workspace group {group_name} already found, ignoring")
+                        continue
+
+                    logger.warning(
```
nfx, line 121:
```suggestion
                    if prompts.confirm(
```
It has to be a conscious choice of account admin to do so

labs.yml
```diff
@@ -118,3 +118,6 @@ commands:
     flags:
       - name: aws-profile
         description: AWS Profile to use for authentication
+
+  - name: create-account-groups
```
nfx, line 122: Missing "is_account: true"

nfx (changes requested):
Did you run this command through Databricks CLI already? 😉 can you record a short demo? Eg you can create an additional workspace in labs-aws for this purpose and delete it after.

One last comment is about prompt-confirm

src/databricks/labs/ucx/account.py
william-conti: Much better indeed

src/databricks/labs/ucx/account.py
```diff
@@ -69,6 +70,84 @@ def sync_workspace_info(self):
...
+    def _add_members_to_acc_group(self, acc_group_id: str, group_name: str, valid_group: Group):
+        for chunk in self._chunks(valid_group.members, 20):
+            logger.debug(f"Adding 20 members to acc group {group_name}")
+            self._ac.groups.patch(
+                acc_group_id,
+                operations=[Patch(op=PatchOp.ADD, path="members", value=[x.as_dict() for x in chunk])],
+                schemas=[PatchSchema.URN_IETF_PARAMS_SCIM_API_MESSAGES_2_0_PATCH_OP],
+            )
+
+    def _chunks(self, lst, n):
```
FastLee, line 98: Consider adding to blueprint

FastLee (changes requested):
Adding mapping functionality is important specially if we have duplicate group names in different workspaces.

nfx:
@william-conti rebase

nfx (approved):
❤️

PR #764, opened 2024-01-08
Added baseline for getting Azure Resource Role Assignments

nfx:
This pull request adds a new command `save-azure-storage-accounts` to the UCX CLI tool. This command identifies all storage accounts used by tables in a HMS metastore, identifies the corresponding service principals and their permissions on each storage account, and saves the data in the CSV file on workspace. The new method `AzureResourcePermissions.save_spn_permissions` performs this functionality.

Additionally, the pull request adds the `AzureResources` class, which is used to fetch information about Azure resources such as subscriptions, storage accounts, and containers. The `AzureRoleAssignment` and `Principal` classes are also added to represent role assignments and their associated principals.

Changes related to tests include:

* Test cases are added for the new `save-azure-storage-accounts` command, including tests for valid and invalid subscription IDs, as well as tests for cases where there are no external tables, no Azure storage accounts, or no valid Azure storage accounts.
* The test case for `test_azure_spn_info_without_secret` is updated to use the `create_autospec` function to create a mock `WorkspaceClient` object instead of creating a mock object manually.
* The `test_move` function is updated to use the `patch` decorator to patch the `TableMove.move_tables` method instead of using a mock object.
* The `test_save_azure_storage_accounts_no_ucx` test case is added to test the behavior when UCX is not installed.
* The `test_save_azure_storage_accounts_not_azure` test case is added to test the behavior when the workspace is not on Azure.
* The `test_save_azure_storage_accounts_no_azure_cli` test case is added to test the behavior when the Azure CLI authentication method is not used.
* The `test_save_azure_storage_accounts_no_subscription_id` test case is added to test the behavior when the subscription ID is not provided.


src/databricks/labs/ucx/assessment/azure.py
```diff
@@ -33,6 +40,27 @@ class AzureServicePrincipalInfo:
...
+@dataclass
+class AzureStorageAccount:
+    name: str
+    resource_id: str
+    subscription_id: str
+
+
+@dataclass
+class AzureStorageSpnPermissionMapping:
+    storage_acct_name: str
```
nfx, line 59: What about containers? What about ACLs on paths? It doesn't have to be done in this PR, but make data structure scale for changes.

Do we want to store this in SQL or CSV? Csv is editable via workspace, in case we want to decouple from az cli runs

Don't use abbreviations.

src/databricks/labs/ucx/assessment/azure.py
```diff
@@ -247,3 +275,113 @@ def snapshot(self) -> Iterable[AzureServicePrincipalInfo]:
...
+
+    return inner
+
+
+class AzureResourcePermissions(CrawlerBase[AzureStorageSpnPermissionMapping]):
+    def __init__(self, ws: WorkspaceClient, lc: ExternalLocations, sql_backend: SqlBackend, inventory_schema: str):
+        super().__init__(
+            sql_backend, "hive_metastore", inventory_schema, "azure_storage_accounts", AzureStorageSpnPermissionMapping
+        )
+        self._rm_host = ws.config.arm_environment.resource_manager_endpoint
```
nfx, line 296: Please separate the class that calls azure API from class that compares data with external locations. And probably we want to the mapping we get from azure api into csv, like we do for table mapping

nfx (commented):
Add tests

Issue #767, opened 2024-01-09
User and backup group permissions lost for redash objects after applying group permissions


mwojtyczka:
We apply group permissions for redash objects using POST (there is no PATCH operation available). That means we are replacing all the permissions. Currently, we only collect and apply the permissions for the account groups. Existing permissions for users, backup groups, and other workspace-local groups that don't have corresponding account groups are lost.

nfx:
@mwojtyczka redash has no PUT

 we collect all permissions, aa far as I remember. We just apply the group permissions 


Thank you for the feature request! Currently, the team operates in a limited capacity, carefully prioritizing, and we cannot provide a timeline to implement this feature. Please make a Pull Request if you'd like to see this feature sooner, and we'll guide you through the journey.

mwojtyczka:
Yes, we collect all groups but we don't collect users. Making PR for this.

PR #770, opened 2024-01-10
Fix issue with missing users and temp groups after workspace-local groups migration and skip table when crawling table size if it does not exist anymore

mwojtyczka:
- Fix for issue: [767](https://github.com/databrickslabs/ucx/issues/767)
- Added new unit and integration tests, improved existing assertions
- Improve get_table_size crawling to skip table if it does not exist anymore, instead of failing the task. This increases the stability of assessment job and integration tests.


src/databricks/labs/ucx/hive_metastore/table_size.py
```diff
@@ -55,12 +60,20 @@ def _try_load(self) -> Iterable[TableSize]:
...
         """
         return self._snapshot(partial(self._try_load), partial(self._crawl))
 
-    def get_table_size(self, table_full_name: str) -> int:
+    def _get_safe_table_size(self, table_full_name: str) -> int | None:
         logger.debug(f"Evaluating {table_full_name} table size.")
-        return self._spark._jsparkSession.table(table_full_name).queryExecution().analyzed().stats().sizeInBytes()
+        try:
+            return self._spark._jsparkSession.table(table_full_name).queryExecution().analyzed().stats().sizeInBytes()
+        except NotFound as nf:
```
nfx, line 74: Spark doesn't yet throw the NotFound, catching Exception is okay here

tests/integration/conftest.py
```diff
@@ -100,7 +100,7 @@ def inner(workspace_group_name=None, account_group_name=None):
         members = [user.id]
         ws_group = make_group(display_name=workspace_group_name, members=members, entitlements=["allow-cluster-create"])
         acc_group = make_acc_group(display_name=account_group_name, members=members)
-        return ws_group, acc_group
+        return ws_group, acc_group, user
```
nfx, line 103: This change doesn't belong here - you see this by the amount of result ignores in tests. If you need a user for one of the tests - don't use this fixture there and write a lower-level code.

tests/integration/workspace_access/test_groups.py
```diff
@@ -175,7 +178,9 @@ def test_group_name_change_substitute(ws, sql_backend, inventory_schema, make_uc
     random_elem = f"{make_random(4)}"
     ws_display_name = f"ucx_engineering_{random_elem}"
     acct_display_name = f"ucx_eng_{random_elem}"
-    ws_group, accnt_group = make_ucx_group(workspace_group_name=ws_display_name, account_group_name=acct_display_name)
+    ws_group, accnt_group, _ = make_ucx_group(
```
nfx, line 181: Revert these

tests/integration/workspace_access/test_groups.py
```diff
@@ -207,7 +212,7 @@ def test_group_matching_names(ws, sql_backend, inventory_schema, make_ucx_group,
 
 
 # average runtime is 100 seconds
-@retried(on=[NotFound], timeout=timedelta(minutes=3))
+@retried(on=[NotFound], timeout=timedelta(minutes=15))
```
nfx, line 215: Don't modify this test, create a new one just for redash queries. It should be in test_redash.py, same as for generic/scim/etc

Mindlessly modifying unrelated tests has previously led them being unmaintainable and i had to spend days rewriting them

tests/integration/workspace_access/test_groups.py
```diff
@@ -216,56 +221,99 @@ def test_replace_workspace_groups_with_account_groups(
...
+
+    permission_manager = PermissionManager(
+        sql_backend, inventory_schema, [generic_permissions, redash_permissions, tacl]
+    )
 
     permission_manager.inventorize_permissions()
 
     @retried(on=[AssertionError], timeout=timedelta(seconds=30))
-    def assert_table_has_two_permissions():
+    def check_inventory_table_permissions():
```
nfx, line 286: Don't modify existing tests! Create new ones.

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/table_size.py
mwojtyczka: we were frequently getting NotFound exceptions here in our integration tests since many tests can run in parallel and tables might get removed.

tests/integration/conftest.py
mwojtyczka: will correct

tests/integration/workspace_access/test_groups.py
mwojtyczka: will do

tests/integration/workspace_access/test_groups.py
mwojtyczka: will do

tests/integration/workspace_access/test_groups.py
mwojtyczka: done

tests/integration/workspace_access/test_groups.py
mwojtyczka: done

tests/integration/workspace_access/test_groups.py
mwojtyczka: done

tests/integration/conftest.py
mwojtyczka: done

src/databricks/labs/ucx/hive_metastore/table_size.py
mwojtyczka: Do you suggest any changes here?

mwojtyczka: corrected

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,115 @@
...
+
+    redash_permissions = RedashPermissionsSupport(
+        ws,
+        [redash.Listing(ws.queries.list, sql.ObjectTypePlural.QUERIES)],
+    )
+
+    permission_manager = PermissionManager(sql_backend, inventory_schema, [redash_permissions])
+
+    permission_manager.inventorize_permissions()
+    state = group_manager.get_migration_state()
```
nfx, line 53: this is too slow and too flaky, rewrite with apply tasks:

```
apply_tasks(
        redash_permissions,
        [
            MigratedGroup.partial_info(group_a, group_b),
        ],
    )
```

see other tests in the folder using the `apply_tasks`

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,115 @@
...
+    group_manager.rename_groups()
+    group_info = state.groups[0]
+
+    # Group information in Redash are cached for up to 10 minutes.
+    # After the group is renamed, the old name may still be returned by the dbsql permissions api.
+    # More details here: https://databricks.atlassian.net/browse/ES-992619
+    @retried(on=[AssertionError], timeout=timedelta(minutes=10))
+    def check_permissions_for_backup_group():
+        logger.info("check_permissions_for_backup_group()")
```
nfx, line 62: don't code this in test, but rather in `RedashPermissionsSupport`. It risks making the whole testing run slower

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,115 @@
...
+        logger.info("check_table_permissions_after_backup_delete()")
+
+        query_permissions = redash_permissions.load_as_dict(sql.ObjectTypePlural.QUERIES, query.id)
+        assert group_info.temporary_name not in query_permissions
+        assert group_info.name_in_account in query_permissions
+        assert user.display_name in query_permissions
+        assert sql.PermissionLevel.CAN_RUN == query_permissions[group_info.name_in_account]
+        assert sql.PermissionLevel.CAN_RUN == query_permissions[user.display_name]
+
+    check_permissions_after_backup_delete()
```
nfx, line 115: this test is too big and going to be flaky for everyone, staggering the merges. please simplify/rewrite it. Look at `test_permissions_for_secrets` as the example short/focused/quick test.

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,115 @@
...
+
+from databricks.labs.ucx.workspace_access import redash
+from databricks.labs.ucx.workspace_access.groups import GroupManager
+from databricks.labs.ucx.workspace_access.manager import PermissionManager
+from databricks.labs.ucx.workspace_access.redash import RedashPermissionsSupport
+
+logger = logging.getLogger(__name__)
+
+
+@retried(on=[NotFound], timeout=timedelta(minutes=13))
```
nfx, line 16: no, 13 minutes is too much. i don't believe this test needs more than 5 minutes of timeout, otherwise it'll slow down the entire merge pipeline

src/databricks/labs/ucx/hive_metastore/table_size.py
```diff
@@ -55,12 +58,19 @@ def _try_load(self) -> Iterable[TableSize]:
...
         Takes a snapshot of tables in the specified catalog and database.
+        Return None if the table cannot be found anymore.
 
         Returns:
             list[Table]: A list of Table objects representing the snapshot of tables.
         """
         return self._snapshot(partial(self._try_load), partial(self._crawl))
 
-    def get_table_size(self, table_full_name: str) -> int:
+    def _get_safe_table_size(self, table_full_name: str) -> int | None:
```
nfx, line 68: nit: are you getting "safe table" or are you "safely getting a table"? :)

tests/unit/workspace_access/test_migration_state.py
```diff
@@ -0,0 +1,19 @@
+from databricks.labs.ucx.workspace_access.groups import MigratedGroup, MigrationState
```
nfx, line 1: by the naming logic, this test should be in the `test_groups`, not the `test_migration_state`.

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/table_size.py
mwojtyczka: safely getting

mwojtyczka: reverted to the previous name

tests/unit/workspace_access/test_migration_state.py
mwojtyczka: moved to test_groups

src/databricks/labs/ucx/hive_metastore/table_size.py
mwojtyczka: redash permissions are cached for up to 10 mins. It is usually faster but worst case it can be 10 mins. So after the rename operation it can take 10 mins until we can see the renamed group in the permissions. On top we have to leave some time to actually perform the rest of the test.

tests/integration/workspace_access/test_redash.py
mwojtyczka: redash permissions are cached for up to 10 mins. It is usually faster but worst case it can be 10 mins. So after the rename operation it can take 10 mins until we can see the renamed group in the permissions. On top we have to leave some time to actually perform the rest of the test.

tests/integration/workspace_access/test_redash.py
mwojtyczka: this is not needed, i removed it

tests/integration/workspace_access/test_redash.py
mwojtyczka: added comments about this in the test

tests/integration/workspace_access/test_redash.py
mwojtyczka: This approach would crawl and then apply permissions. But we need to rename the groups in between. So it won't work.

tests/integration/workspace_access/test_redash.py
mwojtyczka: simplified a bit but we need to go thru the flow of crawling, then renaming and then applying permissions. Otherwise we cannot catch the issue we are trying to solve here.

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,88 @@
...
+    user = make_user()
+
+    query = make_query()
+    make_query_policy_permissions(
+        object_id=query.id,
+        permission_level=sql.PermissionLevel.CAN_RUN,
+        group_name=ws_group.display_name,
+        user_name=user.display_name,
+    )
+    logger.info(f"Query: {ws.config.host}/sql/editor/{query.id}")
```
nfx, line 44: this has to be logged from `make_query` fixture

tests/integration/test_installation.py
```diff
@@ -76,7 +76,7 @@ def test_job_failure_propagates_correct_error_message_and_logs(ws, sql_backend,
     assert len(workflow_run_logs) == 1
 
 
-@retried(on=[NotFound, InvalidParameterValue], timeout=timedelta(minutes=6))
+@retried(on=[NotFound, InvalidParameterValue], timeout=timedelta(minutes=12))
```
nfx, line 79: why do we need this change? increasing a timeout would just hide problems and make tests execute slower.

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,88 @@
...
+
+    group_manager = GroupManager(sql_backend, ws, inventory_schema, [ws_group.display_name], "ucx-temp-")
+    redash_permissions = RedashPermissionsSupport(
+        ws,
+        [redash.Listing(ws.queries.list, sql.ObjectTypePlural.QUERIES)],
+    )
+    permission_manager = PermissionManager(sql_backend, inventory_schema, [redash_permissions])
+
+    permission_manager.inventorize_permissions()
```
nfx, line 53: as i said before, don't call `inventorize_permissions` here, it's too slow.

use the `apply_tasks` instead

```
    apply_tasks(
        generic_permissions,
        [
            MigratedGroup.partial_info(group_a, group_b),
        ],
    )
```

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,88 @@
...
+    group_manager.rename_groups()
+    group_manager.reflect_account_groups_on_workspace()
+    state = group_manager.get_migration_state()
+    permission_manager.apply_group_permissions(state)
+
+    group_info = state.groups[0]
+
+    @retried(on=[AssertionError], timeout=timedelta(minutes=1))
```
nfx, line 61: don't do these `@retried(on=[AssertionError]` checks within integration tests - they're fragile. instead - improve the logic of `RedashPermissionsSupport` and make sure it gives back an already consistent result.

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,88 @@
...
+def test_redash_replace_workspace_groups_with_account_groups(
+    ws,
+    sql_backend,
+    inventory_schema,
+    make_ucx_group,
+    make_user,
+    make_query,
+    make_query_policy_permissions,
+):
+    """
```
nfx, line 29: remove comment, it makes no sense here

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,88 @@
...
+
+    Args:
+        sql_backend (object):
+    """
+    ws_group, _ = make_ucx_group()
```
nfx, line 34: use `apply_tasks` and two workspace groups, the same way as other tests in this directory are doing.

nfx (changes requested):
these integration tests are very brittle in the current shape and I'm not comfortable merging it in current state.

tests/integration/workspace_access/test_redash.py
mwojtyczka: removed

tests/integration/test_installation.py
mwojtyczka: the assessment jobs runs for more than 10 mins sometimes.

tests/integration/workspace_access/test_redash.py
mwojtyczka: but this will both crawl the permissions and apply them to the account. We need to do a rename in between.

tests/integration/workspace_access/test_redash.py
mwojtyczka: I guess we don't need these at all. Will remove

tests/integration/workspace_access/test_redash.py
mwojtyczka: removed

tests/integration/workspace_access/test_redash.py
nfx: why do we need to do a rename? you need group_a and group_b and verify if permissions copied from group a to group b.

we have other tests that check if group renaming works.

mwojtyczka: the bug happens when we rename the group and then read redash permissions. It does not happen if you don't do rename.

tests/integration/workspace_access/test_redash.py
mwojtyczka: updated

tests/integration/workspace_access/test_redash.py
mwojtyczka: updated

tests/integration/workspace_access/test_redash.py
mwojtyczka: updated

tests/integration/workspace_access/test_redash.py
mwojtyczka: updated

tests/integration/workspace_access/test_redash.py
mwojtyczka: updated

tests/integration/workspace_access/test_redash.py
mwojtyczka: updated

src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -271,6 +273,47 @@ def _path(ws, path):
     ]
 
 
+def _redash_permissions_mapping():
+    def _simple(_, object_id):
+        return object_id
+
+    return [
+        (
+            "query_policy",
```
nfx, line 282: Something completely slipped my mind: what is query policy?... did you mean just query?.. 

P.s. we have cluster policy and cluster

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,114 @@
...
+    def rename_group(group: iam.Group, new_group_name: str) -> iam.Group:
+        ws.groups.patch(ws_group.id, operations=[iam.Patch(iam.PatchOp.REPLACE, "displayName", new_group_name)])
+        group.display_name = new_group_name
+        return group
+
+    ws_new_group_name = ws_group.display_name + "_new"
+    permissions = rename_group_in_permissions_obj(ws_group.display_name, ws_new_group_name, permissions)
+    ws_group = rename_group(ws_group, ws_new_group_name)
+
+    group_to_migrate = MigratedGroup.partial_info(ws_group, ws_group)
```
nfx, line 108: This doesn't look correct - can you create two different groups, like all other similar tests? You're creating a single group and doing these weird things related to empty temporary name

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    Redash permissions are cached for up to 10 mins. See: https://databricks.atlassian.net/browse/ES-992619
+    Therefore, when a group is renamed, get redash permissions API can return the old group name for some time.
+    Note that the update/set API is strongly consistent and is not affected by this behaviour.
+    This test validates that Redash Permissions Support is able to apply permissions correctly after rename operation.
+    """
+    ws_group = make_group()
+
+    query = make_query()
+    make_query_policy_permissions(
```
nfx, line 76: What is query policy? Is it a new product?

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,114 @@
...
+    make_query,
+    make_query_policy_permissions,
+):
+    """
+    Redash permissions are cached for up to 10 mins. See: https://databricks.atlassian.net/browse/ES-992619
+    Therefore, when a group is renamed, get redash permissions API can return the old group name for some time.
+    Note that the update/set API is strongly consistent and is not affected by this behaviour.
+    This test validates that Redash Permissions Support is able to apply permissions correctly after rename operation.
+    """
+    ws_group = make_group()
```
nfx, line 73: This doesn't look correct - can you create two different groups, like all other similar tests? You're creating a single group and doing these weird things related to empty temporary name

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,114 @@
...
+        ws.groups.patch(ws_group.id, operations=[iam.Patch(iam.PatchOp.REPLACE, "displayName", new_group_name)])
+        group.display_name = new_group_name
+        return group
+
+    ws_new_group_name = ws_group.display_name + "_new"
+    permissions = rename_group_in_permissions_obj(ws_group.display_name, ws_new_group_name, permissions)
+    ws_group = rename_group(ws_group, ws_new_group_name)
+
+    group_to_migrate = MigratedGroup.partial_info(ws_group, ws_group)
+    group_to_migrate.temporary_name = None  # don't apply permissions for the temp group
```
nfx, line 109: This doesn't look correct - can you create two different groups, like all other similar tests? You're creating a single group and doing these weird things related to empty temporary name

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,114 @@
...
+
+    apply_tasks_appliers(redash_permissions, permissions, MigrationState([group_to_migrate]))
+
+    query_permissions = redash_permissions.load_as_dict(sql.ObjectTypePlural.QUERIES, query.id)
+    assert sql.PermissionLevel.CAN_EDIT == query_permissions[ws_group.display_name]
```
nfx, line 114: Assert permissions for both original and migrated groups

tests/integration/workspace_access/test_redash.py
```diff
@@ -0,0 +1,114 @@
...
+    query = make_query()
+    make_query_policy_permissions(
+        object_id=query.id,
+        permission_level=sql.PermissionLevel.CAN_EDIT,
+        group_name=ws_group.display_name,
+        user_name=user.display_name,
+    )
+
+    group_to_migrate = MigratedGroup.partial_info(ws_group, ws_group)
+    group_to_migrate.temporary_name = None  # don't apply permissions for the temp group
```
nfx, line 43: This doesn't look correct - can you create two different groups, like all other similar tests? You're creating a single group and doing these weird things related to empty temporary name

nfx (changes requested):

src/databricks/labs/ucx/mixins/fixtures.py
mwojtyczka: corrected

tests/integration/workspace_access/test_redash.py
mwojtyczka: corrected

tests/integration/workspace_access/test_redash.py
mwojtyczka: corrected

tests/integration/workspace_access/test_redash.py
mwojtyczka: corrected

tests/integration/workspace_access/test_redash.py
mwojtyczka: corrected

tests/integration/workspace_access/test_redash.py
mwojtyczka: corrected

tests/integration/workspace_access/test_redash.py
mwojtyczka: corrected

nfx (approved):

PR #772, opened 2024-01-11
Added `databricks labs ucx validate-groups-membership` command to validate groups to see if they have same membership across acount and workspace level

### LINKED ISSUE #649: Catch groups that doesn't have the same membership on Workspace and Account level

prajin-29:
This pull request introduces several changes related to managing Databricks workspace groups. Here's a summary of the changes:

1. A new command `validate-groups-membership` is added to the CLI tool in the file `cli.py`. This command validates the groups in the workspace and account levels to see if they have different memberships. It returns a table showing the differences between the groups.
2. A new class `GroupManager` is added to the file `groups.py` to manage the groups in the workspace and account levels. This class includes a new method `validate_group_membership` that compares the members of each group in the workspace and account levels and returns a list of groups that have different memberships.
3. A new fixture `make_ucx_group_with_diff_members` is added to the file `conftest.py` to create a pair of groups with different members.
4. A new test case `test_group_matching_names_with_diff_users` is added to the file `test_groups.py` to test the `validate_group_membership` method with a pair of groups that have different members.
5. A retries decorator is added to the test case `test_replace_workspace_groups_with_account_groups` to handle potential intermittent failures.

These changes improve the group management functionality of the tool and make it more robust and reliable. The new `validate-groups-membership` command provides valuable information about the groups in the workspace and account levels, and the new `GroupManager` class and test cases ensure that the group management functionality works as expected.

Closes #649

### DIFF
--- a/labs.yml
+++ b/labs.yml
@@ -80,3 +80,6 @@ commands:
         description: target catalog to migrate schema to
       - name: to-schema
         description: target schema to migrate tables to
+
+  - name: validate-groups
+    description: Validate the groups to see if the groups at account level and workspace level have different membership
--- a/labs.yml
+++ b/labs.yml
@@ -81,5 +81,9 @@ commands:
       - name: to-schema
         description: target schema to migrate tables to
 
-  - name: validate-groups
+  - name: validate-groups-membership
     description: Validate the groups to see if the groups at account level and workspace level have different membership
+    table_template: |-
+      Workflow_Group_Name\tAccount_Group_Name
+      {{range .}}{{.wf_group_name}}\t{{.ac_group_name}}
+      {{end}}
--- a/labs.yml
+++ b/labs.yml
@@ -90,6 +90,6 @@ commands:
   - name: validate-groups-membership
     description: Validate the groups to see if the groups at account level and workspace level have different membership
     table_template: |-
-      Workflow_Group_Name\tAccount_Group_Name
-      {{range .}}{{.wf_group_name}}\t{{.ac_group_name}}
+      Workflow Group Name\tAccount Group Name
+      {{range .}}{{.name_in_workspace}}\t{{.name_in_group}}
       {{end}}
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -155,6 +155,7 @@ def validate_groups_membership(w: WorkspaceClient):
     if not installation:
         logger.error(CANT_FIND_UCX_MSG)
         return None
+    logger.info("Validating Groups which are having different memberships between account and workspace level")
     group_manager = GroupManager(
         sql_backend=sql_backend,
         ws=w,
@@ -166,9 +167,7 @@ def validate_groups_membership(w: WorkspaceClient):
         account_group_regex=account_group_regex,
     )
     mismatch_groups = group_manager.validate_group_membership()
-
-    if mismatch_groups:
-        print(json.dumps(mismatch_groups))
+    print(json.dumps(mismatch_groups))
 
 
 @ucx.command
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -144,6 +144,9 @@ def validate_groups_membership(w: WorkspaceClient):
     """Validate the groups to see if the groups at account level and workspace level has different membership"""
     installation_manager = InstallationManager(w)
     installation = installation_manager.for_user(w.current_user.me())
+    if not installation:
+        logger.error(CANT_FIND_UCX_MSG)
+        return None
     warehouse_id = installation.config.warehouse_id
     inventory_database = installation.config.inventory_database
     renamed_group_prefix = installation.config.renamed_group_prefix
@@ -152,10 +155,7 @@ def validate_groups_membership(w: WorkspaceClient):
     account_group_regex = installation.config.account_group_regex
     include_group_names = installation.config.include_group_names
     sql_backend = StatementExecutionBackend(w, warehouse_id)
-    if not installation:
-        logger.error(CANT_FIND_UCX_MSG)
-        return None
-    logger.info("Validating Groups which are having different memberships between account and workspace level")
+    logger.info("Validating Groups which are having different memberships between account and workspace")
     group_manager = GroupManager(
         sql_backend=sql_backend,
         ws=w,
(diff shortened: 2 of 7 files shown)


src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -38,6 +38,7 @@ class MigratedGroup:
     entitlements: str | None = None
     external_id: str | None = None
     roles: str | None = None
+    is_same_membership: bool | None = None
```
nfx, line 41: no, you're not allowed to add this column, as it'll be a breaking change to database, which this small feature is not justifying

nfx (changes requested):
do this functionality as a CLI command. You can rely on `GroupManager`, but don't rely on `ucx.groups` table - e.g. it should be possible to run `databricks labs ucx validate-groups` before running the assessment workflow

src/databricks/labs/ucx/workspace_access/groups.py
prajin-29: Changed the logic to incorporate it with CLI.

prajin-29:
> do this functionality as a CLI command. You can rely on `GroupManager`, but don't rely on `ucx.groups` table - e.g. it should be possible to run `databricks labs ucx validate-groups` before running the assessment workflow

@nfx .Did this in CLI Command. Can you please validate the same.

tests/integration/conftest.py
```diff
@@ -105,6 +105,26 @@ def inner(workspace_group_name=None, account_group_name=None):
     return inner
 
 
+@pytest.fixture
+def make_ucx_group_with_diff_members(make_random, make_group, make_acc_group, make_user):
```
nfx, line 109: Don't add a fixture if it's used only once. It makes the code convoluted and harder to maintain. Rewrite by inlining the fixture code into that single test

tests/unit/workspace_access/test_groups.py
```diff
@@ -824,3 +824,62 @@ def test_state():
     assert not state.get_target_principal("invalid_group_name")
     assert not state.get_temp_principal("invalid_group_name")
     assert not state.is_in_scope("invalid_group_name")
+
+
+def test_validate_group_diff_membership():
+    backend = MockBackend()
+    wsclient = MagicMock()
```
nfx, line 831: Use "create_autospec(WorkspaceClient)"

tests/unit/workspace_access/test_groups.py
```diff
@@ -824,3 +824,62 @@ def test_state():
...
+    }
+    grp_membership = GroupManager(
+        backend, wsclient, inventory_database="inv", workspace_group_regex=r"\(([1-9]+)\)", account_group_regex="[1-9]+"
+    ).validate_group_membership()
+    assert grp_membership == [{"wf_group_name": "test_(1234)", "ac_group_name": "ac_test_1234"}]
+
+
+def test_validate_group_same_membership():
+    backend = MockBackend()
+    wsclient = MagicMock()
```
nfx, line 858: Create autospec

nfx (changes requested):

src/databricks/labs/ucx/cli.py
```diff
@@ -138,6 +139,37 @@ def repair_run(w: WorkspaceClient, step):
...
+        ws=w,
+        inventory_database=inventory_database,
+        include_group_names=include_group_names,
+        renamed_group_prefix=renamed_group_prefix,
+        workspace_group_regex=workspace_group_regex,
+        workspace_group_replace=workspace_group_replace,
+        account_group_regex=account_group_regex,
+    )
+    mismatch_groups = group_manager.validate_group_membership()
+    print(json.dumps(mismatch_groups))
```
william-conti, line 170: It should be better to return an error code when there's at least one entry in the mismatch_groups

src/databricks/labs/ucx/cli.py
```diff
@@ -138,6 +139,37 @@ def repair_run(w: WorkspaceClient, step):
     installer.repair_run(step)
 
 
+@ucx.command
+def validate_groups_membership(w: WorkspaceClient):
```
william-conti, line 143: What if I wan't to scope it for all workspaces in my account ?

william-conti (changes requested):
We should plan on enlarging this feature for all the workspaces in the account

src/databricks/labs/ucx/cli.py
nfx: It has to be an additional command

labs.yml
```diff
@@ -80,3 +80,10 @@ commands:
         description: target catalog to migrate schema to
       - name: to-schema
         description: target schema to migrate tables to
+
+  - name: validate-groups-membership
+    description: Validate the groups to see if the groups at account level and workspace level have different membership
+    table_template: |-
+      Workflow_Group_Name\tAccount_Group_Name
```
nfx, line 87: Use spaces for separating words, not underscores

labs.yml
```diff
@@ -80,3 +80,10 @@ commands:
...
+      {{range .}}{{.wf_group_name}}\t{{.ac_group_name}}
```
nfx, line 88:
```suggestion
      {{range .}}{{.name_in_workspace}}\t{{.name_in_account}}
```
Use conventional variable names

src/databricks/labs/ucx/cli.py
nfx: I'm not sure if CLI wrapper supports that just yet

nfx (changes requested):
1. Inline fixture that is not reused
2. Make variable names consistent 
3. Add additional account level command

tests/integration/conftest.py
prajin-29: Removed this fixture and rewrote it inline with test.

tests/unit/workspace_access/test_groups.py
prajin-29: Used create_autospec

labs.yml
prajin-29: Replaced _ with space

labs.yml
prajin-29: Renamed with the conventional names

nfx (approved):

PR #779, opened 2024-01-12
Mount Point crawler lists /Volume with four variations which is confusing

### LINKED ISSUE #622: Mount Point crawler lists /Volume with four variations which is confusing
In the Mount Point crawler, how do we end up with four spellings for Volume ?
I think we should clean this up in UCX and escalate to platform.
We can also enhance this view with a action/recommendation column.

/ | DatabricksRoot |  
-- | -- | --
/Volume | DbfsReserved |  
/Volumes | UnityCatalogVolumes |  
/databricks-datasets | databricks-datasets |  
/databricks-results | databricks-results |  
/databricks/mlflow-registry | databricks/mlflow-registry |  
/databricks/mlflow-tracking | databricks/mlflow-tracking |  
/volume | DbfsReserved |  
/volumes | DbfsReserved

dipankarkush-db:
This pull request includes changes to the `locations.py` file in the `databricks/labs/ucx/hive_metastore` directory and the addition of a new unit test in `test_locations.py`. The changes in `locations.py` involve modifying the `_deduplicate_mounts` method in the `Mounts` class. The method now checks if the name of the `Mount` object contains the word "volume", and if so, it standardizes the name to "/Volume". This is done to ensure that mounts with different casing for the word "volume" are treated as the same mount and not duplicated in the deduplicated list.

The new unit test in `test_locations.py` verifies that the `_deduplicate_mounts` method correctly deduplicates a list of `Mount` objects with different casing for the word "volume" in their names. The test creates a list of `Mount` objects with different casing for the word "volume" and passes it to the `_deduplicate_mounts` method. The test then asserts that the deduplicated list contains only one `Mount` object with the name "/Volume". This test ensures that the changes to the `_deduplicate_mounts` method work as intended and prevent duplication of mounts with different casing for the word "volume" in their names.

In summary, the changes in this pull request modify the `_deduplicate_mounts` method to standardize the name of mounts containing the word "volume" and add a unit test to ensure that the method correctly deduplicates a list of `Mount` objects with different casing for the word "volume" in their names.

### DIFF
--- a/tests/unit/hive_metastore/test_locations.py
+++ b/tests/unit/hive_metastore/test_locations.py
@@ -47,17 +47,24 @@ def test_list_mounts_should_return_a_deduped_list_of_mount_without_encryption_ty
 
 
 def test_list_mounts_should_return_a_deduped_list_of_mount_without_variable_volume_names():
-    mounts = [
-        Mount(name="/Volume", source="DbfsReserved"),
-        Mount(name="/Volumes", source="DbfsReserved"),
-        Mount(name="/volume", source="DbfsReserved"),
-        Mount(name="/volumes", source="DbfsReserved"),
-    ]
     client = MagicMock()
-    backend = MockBackend()
-    mount_list = Mounts(backend, client, "test")._deduplicate_mounts(mounts)
+    client.dbutils.fs.mounts.return_value = [
+        MountInfo("/Volume", "DbfsReserved", "info_1"),
+        MountInfo("/Volumes", "DbfsReserved", "info_2"),
+        MountInfo("/volume", "DbfsReserved", "info_3"),
+        MountInfo("/volumes", "DbfsReserved", "info_4"),
+        MountInfo("mp_1", "path_1", "info_1"),
+        MountInfo("mp_2", "path_2", "info_2"),
+        MountInfo("mp_2", "path_2", "info_2"),
+    ]
 
-    assert mount_list == [Mount("/Volume", "DbfsReserved")]
+    backend = MockBackend()
+    instance = Mounts(backend, client, "test")
+
+    instance.inventorize_mounts()
+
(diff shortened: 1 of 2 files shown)


tests/unit/hive_metastore/test_locations.py
```diff
@@ -46,6 +46,20 @@ def test_list_mounts_should_return_a_deduped_list_of_mount_without_encryption_ty
     assert expected == backend.rows_written_for("hive_metastore.test.mounts", "append")
 
 
+def test_list_mounts_should_return_a_deduped_list_of_mount_without_variable_volume_names():
+    mounts = [
+        Mount(name="/Volume", source="DbfsReserved"),
+        Mount(name="/Volumes", source="DbfsReserved"),
+        Mount(name="/volume", source="DbfsReserved"),
+        Mount(name="/volumes", source="DbfsReserved"),
```
nfx, line 54: Add normal mount and assert it appears

tests/unit/hive_metastore/test_locations.py
```diff
@@ -46,6 +46,20 @@ def test_list_mounts_should_return_a_deduped_list_of_mount_without_encryption_ty
     assert expected == backend.rows_written_for("hive_metastore.test.mounts", "append")
 
 
+def test_list_mounts_should_return_a_deduped_list_of_mount_without_variable_volume_names():
+    mounts = [
+        Mount(name="/Volume", source="DbfsReserved"),
+        Mount(name="/Volumes", source="DbfsReserved"),
```
nfx, line 52: Can we just filter out dbfs reserved mounts instead?

nfx (changes requested):

tests/unit/hive_metastore/test_locations.py
dipankarkush-db: Done!

tests/unit/hive_metastore/test_locations.py
dipankarkush-db: checked for DbfsReserved.

PR #780, opened 2024-01-14
Create instructions to install UCX v0.9.0 and greater from a Databricks cluster web terminal

dmoore247:
Stumbled upon how to install UCX from a web terminal. This will be useful for when the user does not have desktop access to install dependencies etc.

image

### DIFF
--- a/docs/cluster_install.md
+++ b/docs/cluster_install.md
@@ -11,7 +11,7 @@ apt install python3.10-venv
 unset DATABRICKS_RUNTIME_VERSION
 ```
 
-Download and install databricks cli
+Download and install Databricks cli
 
 ```shell
 curl -fsSL https://raw.githubusercontent.com/databricks/setup-cli/main/install.sh | sh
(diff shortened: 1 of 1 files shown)


dmoore247:
I need to format the readme better... working...

docs/cluster_install.md
```diff
@@ -0,0 +1,65 @@
...
+Verify Databricks cli:
+
+```shell
+which databricks
+databricks --version
+```
+
+Configure Databricks cli to authenticate to correct workspace, add host name and PAT token:
+```shell
+databricks configure
```
nfx, line 35: This is security hole, as this configuration is accessible to all users on the same cluster. No way we're recommending this

nfx (changes requested):

docs/cluster_install.md
nfx: cc @hamza-db

dmoore247: We can specify a single user cluster and denying access to the cluster.

dmoore247: @nfx any alternatives you'd suggest?

dmoore247: Ok, Environmental variable is session based, not available to other shells.

nfx: users by mistake run it on a shared cluster and then it's root - everyone, they leak admin credentials for all users. no.

nfx:
this PR has security windows wide open, not willing to be on customer calls related to breached PAT tokens on shared clusters, that were not recommended explicitly. let's wait for a proper implementation on platform side.

PR #782, opened 2024-01-15
Added debug_truncate_bytes parameter to the config and extend the log truncation limit

mwojtyczka:
By default, the UCX is truncating API requests and responses in the log files to 96 bytes. This makes it very hard to debug any issues. This PR adds the debug_truncate_bytes parameter in the config and increases the truncation limit.


nfx (approved):
Lgtm

PR #816, opened 2024-01-19
Improve validate groups membership cli command

mwojtyczka:
## Changes
* Corrected the fields returned by the cli. Currently the cli does not display the group names correctly.
* Added group members count

### Functionality 

- [x] modified existing cli command: `databricks labs ucx validate-groups-membership`

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/workspace_access/groups.py
+++ b/src/databricks/labs/ucx/workspace_access/groups.py
@@ -424,7 +424,9 @@ class GroupManager(CrawlerBase[MigratedGroup]):
             mismatch_group.append(
                 {
                     "wf_group_name": groups.name_in_workspace,
-                    "ac_group_name": groups.name_in_account,
+                    "wf_group_members_count": len(ws_members_set),
+                    "acc_group_name": groups.name_in_account,
+                    "acc_group_members_count": len(ac_members_set),
                 }
             )
         if not mismatch_group:
--- a/labs.yml
+++ b/labs.yml
@@ -90,6 +90,6 @@ commands:
   - name: validate-groups-membership
     description: Validate the groups to see if the groups at account level and workspace level have different membership
     table_template: |-
-      Workflow Group Name\tAccount Group Name
-      {{range .}}{{.name_in_workspace}}\t{{.name_in_group}}
+      Workspace Group Name\tMembers Count\tAccount Group Name\tMembers Count
+      {{range .}}{{.wf_group_name}}\t{{.wf_group_members_count}}\t{{.acc_group_name}}\t{{.acc_group_members_count}}
       {{end}}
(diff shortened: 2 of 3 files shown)


src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -408,23 +408,25 @@ def _crawler(self) -> Iterable[MigratedGroup]:
 
     def validate_group_membership(self) -> list[dict]:
         workspace_groups_in_workspace = self._workspace_groups_in_workspace()
-        account_groups_in_account = self._account_groups_in_account()
+        account_groups_in_account = self._account_groups_in_account("id,displayName,externalId,members")
```
william-conti, line 411: Be careful when fetching unbounded attributes like "members" we had timeout problems with the Workspace API in the past because of this. We may run into the same issue there

william-conti: Additionally, I don't think you need the externalId field

william-conti (changes requested):
Looks great, only minor change that is debatable

src/databricks/labs/ucx/workspace_access/groups.py
mwojtyczka: corrected. I pull the members now with a separate Get. It's much faster and we don't risk a timeout

william-conti (approved):

nfx (approved):

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -409,19 +409,24 @@ def validate_group_membership(self) -> list[dict]:
...
-            ws_members_set = set([m.get("display") for m in json.loads(groups.members)] if groups.members else [])
-            account_group = account_groups_in_account[groups.name_in_account]
-            ac_members_set = set(
-                [a.as_dict().get("display") for a in account_group.members] if account_group.members else []
-            )
-            set_diff = (ws_members_set - ac_members_set).union(ac_members_set - ws_members_set)
+        retry_on_internal_error = retried(on=[InternalError], timeout=self._verify_timeout)
+        get_account_group = retry_on_internal_error(self._get_account_group)
+        for ws_group in migrated_groups:
+            # Users with the same display name but different email will be deduplicated!
```
nfx, line 415: Isn't that a bug?🐜

nfx (approved):

src/databricks/labs/ucx/workspace_access/groups.py
mwojtyczka: Comparing by email address would be better. But it's a limitation of the scim groups API. It only returns display name of the user. Email address is not present. We could fetch user info separately but a user can theoretically have multiple email addresses so this might not help.

nfx (approved):
Lgtm

PR #836, opened 2024-01-24
Fixed escaping SQL object names

### LINKED ISSUE #833: [BUG]: Function crawl_grants blows up when databases have special characters

### Current Behavior

The task blows up with the following exception:
[INVALID_IDENTIFIER] The identifier canopy-test-db is invalid. Please, consider quoting it with back-quotes as `canopy-test-db`.(line 1, pos 46)
== SQL ==
SHOW USER FUNCTIONS FROM hive_metastore.canopy-test-db;

### Expected Behavior

The the catalog.schema should be enclosed in `` to handle special characters in names.  Also the schema shema should be skipped and task should continue to other databases (not blow up)

### Steps To Reproduce

1.  Create a database with dashes in hive_metastore, for example canopy-test-db

### Cloud

AWS

### Operating System

Linux

### Version

latest via Databricks CLI

### Relevant log output

```shell
[INVALID_IDENTIFIER] The identifier canopy-test-db is invalid. Please, consider quoting it with back-quotes as `canopy-test-db`.(line 1, pos 46)
== SQL ==
SHOW USER FUNCTIONS FROM hive_metastore.canopy-test-db;


Below is the function where this exception comes from:
Its in this function:  " def _crawl(self) -> Iterable[Udf]:
        """Crawls and lists udfs within the specified catalog and database."""
        tasks = []
        catalog = "hive_metastore"
        # need to set the current catalog otherwise "SHOW USER FUNCTIONS FROM" is raising error:
        # "target schema <database> is not in the current catalog"
        self._exec(f"USE CATALOG {catalog};")
        for (database,) in self._all_databases():
            logger.debug(f"[{catalog}.{database}] listing udfs")
            for (udf,) in self._fetch(f"SHOW USER FUNCTIONS FROM {catalog}.{database};"):
                if udf.startswith(f"{catalog}.{database}"):
                    udf_name = udf[udf.rfind(".") + 1 :]  # remove catalog and database info from the name
                    tasks.append(partial(self._describe, catalog, database, udf_name))
        catalog_tables, errors = Threads.gather(f"listing udfs in {catalog}", tasks)
        if len(errors) > 0:
            logger.error(f"Detected {len(errors)} while scanning udfs in {catalog}")
        return catalog_tables"
```

larsgeorge-db:
## Changes

Implements handling of special characters in SQL names. Optionally escapes the catalog, schema, or table names with backticks for SQL queries.

### Linked issues

Resolves #833

### Tests

- [x] added unit tests

### DIFF
--- a/src/databricks/labs/ucx/framework/crawlers.py
+++ b/src/databricks/labs/ucx/framework/crawlers.py
@@ -2,7 +2,6 @@ import dataclasses
 import logging
 import os
 import pkgutil
-import string
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable, Iterator, Sequence
 from types import UnionType
@@ -20,7 +19,6 @@ from databricks.sdk.errors import (
 from databricks.labs.ucx.mixins.sql import Row, StatementExecutionExt
 
 logger = logging.getLogger(__name__)
-allowed_object_chars = set(string.ascii_letters + string.digits + '_')
 
 
 class DataclassInstance(Protocol):
@@ -232,28 +230,6 @@ class CrawlerBase(Generic[Result]):
             str: The full table name.
         """
         return f"{self._catalog}.{self._schema}.{self._table}"
-
-    @staticmethod
-    def escape(path: str, optional: bool | None = True) -> str:
-        """
-        Escapes the path components to make them SQL safe.
-
-        Args:
-            path (str): The dot-separated path of a catalog object.
-            optional (bool): If `True` then do not escape if no special characters are present.
-
-        Returns:
-             str: The path with all parts escaped in backticks.
-        """
-        parts = path.split(".")
-        escaped = []
-        for part in parts:
-            if not part.startswith("`") and not part.endswith("`"):
-                part = part.strip("`")
-                if not optional or not set(part) <= allowed_object_chars:
-                    part = f"`{part}`"
-            escaped.append(part)
-        return ".".join(escaped)
 
     @staticmethod
     def _valid(name: str) -> str:
--- a/src/databricks/labs/ucx/framework/crawlers.py
+++ b/src/databricks/labs/ucx/framework/crawlers.py
@@ -2,6 +2,7 @@ import dataclasses
 import logging
 import os
 import pkgutil
+import string
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable, Iterator, Sequence
 from types import UnionType
@@ -19,6 +20,7 @@ from databricks.sdk.errors import (
 from databricks.labs.ucx.mixins.sql import Row, StatementExecutionExt
 
 logger = logging.getLogger(__name__)
+allowed_object_chars = set(string.ascii_letters + string.digits + '_')
 
 
 class DataclassInstance(Protocol):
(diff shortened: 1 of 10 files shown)


tests/unit/framework/test_crawlers.py
```diff
@@ -51,6 +51,28 @@ def test_full_name():
...
+    assert "a.`b-c`.d" == cb.escape("a.b-c.d")
+    assert "a.b.`c-d`" == cb.escape("a.b.c-d")
+    # test with escaping enforced
+    assert "`a`" == cb.escape("a", False)
+    assert "`a`.`b`" == cb.escape("a.b", False)
+    assert "`a`.`b`.`c`" == cb.escape("a.b.c", False)
+    assert "`a-b`.`c`.`d`" == cb.escape("a-b.c.d", False)
+    assert "`a`.`b-c`.`d`" == cb.escape("a.b-c.d", False)
+    assert "`a`.`b`.`c-d`" == cb.escape("a.b.c-d", False)
+
```
dmoore247, line 74: Amazing!!!
Please consider testing unicode, kanji, embedded space characters.

This documentation isn't much help... https://docs.databricks.com/en/sql/language-manual/sql-ref-names.html#names

dmoore247 (approved):
Amazing work!
Please consider testing unicode, kanji (Multi-language support, and embedded space chars).

src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -231,6 +233,28 @@ def _full_name(self) -> str:
         """
         return f"{self._catalog}.{self._schema}.{self._table}"
 
+    @staticmethod
```
nfx, line 236: create a function of it and call it `escape_sql_identifier`, it doesn't really belong to this class, as it turns out from number of `CrawlerBase.` prefixes

nfx (changes requested):

dmoore247:
@larsgeorge-db can you test for `mytable version1.0` ?

daniel-martinez-db:
I have a pattern like this:
my_catalog.my-database.mytable_namev1.0 which should be escaped as 
`my_catalog`.`my-database`.`mytable_name_v1.0`
Can that test be included please?

larsgeorge-db:
> I have a pattern like this: my_catalog.my-database.mytable_namev1.0 which should be escaped as `my_catalog`.`my-database`.`mytable_name_v1.0` Can that test be included please?

How would you handle this: `cat.alog.schema.table`? Is your naming assuming only the last part can have dots in it? Please clarify.

src/databricks/labs/ucx/framework/crawlers.py
larsgeorge-db: Yeah, that makes sense, I was about to ask that. Where do we put functions? A new Python source in the root of the project? Or some `utils` subdir?

daniel-martinez-db:
> > I have a pattern like this: my_catalog.my-database.mytable_namev1.0 which should be escaped as `my_catalog`.`my-database`.`mytable_name_v1.0` Can that test be included please?
> 
> How would you handle this: `cat.alog.schema.table`? Is your naming assuming only the last part can have dots in it? Please clarify. Also, see the above linked Databricks Naming convention, it states dots are NOT allowed in names.

Thanks. The patterns I'm working with only has dots in the table name, although I think we can escape all of them (maybe not in the test, but in the actual code). It's true that DB does not allow for dots, but other catalogs do; I'm facing this issue as I'm crawling an external hive metastore which allows for dots. Thank you!

nfx (approved):

PR #841, opened 2024-01-25
Added verification of group permissions

andrascsillag-db:
## Changes
Added `verify` method to all `AclSupport` subclasses. Existing `_inflight_check` and similar names were renamed to _verify for consistency. Expose a method to validate groups permissions and added cli command.

### Functionality 

- [x] added new CLI command (`validate-groups-permissions`) to run verification of groups permissions

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


src/databricks/labs/ucx/workspace_access/tacl.py
```diff
@@ -86,15 +88,47 @@ def get_apply_task(self, item: Permissions, migration_state: MigrationState):
             return None
         target_grant = dataclasses.replace(grant, principal=target_principal)
         # this has to be executed on tacl cluster, otherwise - use SQLExecutionAPI backend & Warehouse
-        return partial(self._apply_grant, target_grant)
+        object_type, object_id = target_grant.this_type_and_key()
+        return partial(self._apply_grant, object_type, object_id, target_grant)
 
-    def _apply_grant(self, grant: Grant):
+    def _apply_grant(self, object_type: str, object_id: str, grant: Grant):
```
nfx, line 94: You can get object type and ID from grant itself, don't add extras arguments here

src/databricks/labs/ucx/workspace_access/base.py
```diff
@@ -32,3 +35,12 @@ def get_apply_task(self, item: Permissions, migration_state: MigrationState) ->
...
     def object_types(self) -> set[str]:
         """This method returns a set of strings, that represent object types that are applicable by this instance."""
+
+    @abstractmethod
+    def verify(
+        self,
+        object_type: str,
+        object_id: str,
+        acl: list[iam.AccessControlRequest | sql.AccessControl | iam.ComplexValue] | Grant | workspace.AclItem,
+    ) -> bool:
```
nfx, line 45:
```suggestion
    def verify(self, item: Permissions) -> bool:
```
You can't add union type with child-related classes here, because it leaks implementation details upwards and common sense doesn't allow us to merge this as it is now. It simply means no sense and breaks the dependency flow, increasing coupling.

Please rewrite the underlying implementations, the way both applier tasks and verify call a common private method. And implementation of verify() converts Permission into a call of that private method.

I also don't see any external calls of this verify() method - why do we need it, again?

src/databricks/labs/ucx/workspace_access/scim.py
```diff
@@ -94,22 +95,31 @@ def _crawler_task(group: iam.Group, property_name: str):
             raw=json.dumps([e.as_dict() for e in getattr(group, property_name)]),
         )
 
-    def _inflight_check(self, group_id: str, value: list[iam.ComplexValue], property_name: str):
+    def verify(
```
nfx, line 98: Okay, I see what you're mean by incompatibility. So, in this case: keep in_flight_check() methods mostly as is and call them from verify() method, which has to have a very narrow interface.

nfx (changes requested):
Please rewrite, this PR doesn't follow principles of object-oriented design.

src/databricks/labs/ucx/workspace_access/base.py
mwojtyczka: the external call could be used if we want to build cli command for verification

mwojtyczka:
Refactored the code. Need to add more tests for the verify method. Also plan to add CLI command for verification.

src/databricks/labs/ucx/cli.py
```diff
@@ -146,6 +147,26 @@ def repair_run(w: WorkspaceClient, step):
     installer.repair_run(step)
 
 
+@ucx.command
+def validate_groups_permissions(w: WorkspaceClient):
+    """Validate that all the crawled permissions are applied correctly to the destination groups"""
+    logger.info("Running validation of permissions applied to groups.")
+    installation_manager = InstallationManager(w)
+    installation = installation_manager.for_user(w.current_user.me())
```
nfx, line 155:
```suggestion
    installation = Installation.current(w, "ucx")
```

src/databricks/labs/ucx/workspace_access/manager.py
```diff
@@ -140,6 +140,46 @@ def apply_group_permissions(self, migration_state: MigrationState):
...
+            if object_type not in appliers:
+                msg = f"Could not find support for {object_type}. Please check the inventory table."
+                raise ValueError(msg)
+
+        for object_type, items_subset in supports_to_items.items():
+            relevant_support = appliers[object_type]
+            tasks_for_support: list[Callable[..., bool]] = []  # type: ignore[annotation-unchecked]
+            for item in items_subset:
+                if not item:
+                    continue
```
nfx, line 164: can you simplify lines 147-164 to be just a single loop? we don't need `supports_to_items` intermediary structure at all.

src/databricks/labs/ucx/workspace_access/manager.py
```diff
@@ -140,6 +140,46 @@ def apply_group_permissions(self, migration_state: MigrationState):
...
+                tasks_for_support.append(task)
+            if len(tasks_for_support) == 0:
+                continue
+            logger.info(f"Total tasks for {object_type}: {len(tasks_for_support)}")
+            verifier_tasks.extend(tasks_for_support)
+
+        logger.info(f"Starting to verify permissions applied to groups. Total tasks: {len(verifier_tasks)}")
+
+        _, errors = Threads.gather("verify group permissions", verifier_tasks)
```
nfx, line 176: can you do `Threads.strict`, so that an error is thrown?

nfx (commented):
interface looks good, no union type. main comments are about `verify_group_permissions` simplification

src/databricks/labs/ucx/cli.py
mwojtyczka: we use the same in all cli commands. Installation does not have "current" method

src/databricks/labs/ucx/workspace_access/manager.py
mwojtyczka: done

src/databricks/labs/ucx/workspace_access/manager.py
mwojtyczka: done

src/databricks/labs/ucx/cli.py
nfx: https://github.com/databrickslabs/ucx/pull/860 there's no more installation manager

mwojtyczka: got it but the other PR must be merged first

nfx: merged :)

tests/unit/workspace_access/test_generic.py
```diff
@@ -814,3 +814,116 @@ def perms(object_type: str, object_id: str):
             tasks.append(task)
     assert len(tasks) == 1
     assert tasks[0].object_type == "clusters"
+
+
+def test_verify_task_should_return_true_if_permissions_applied():
+    ws = MagicMock()
```
nfx, line 820: could you use `create_autospec(WorkspaceClient)` instead? it's "safer"

nfx (changes requested):
please rebase on main - code looks mostly good, thanks for finishing it.

how long does it take to run? if it's less than 10 minutes - doing it in CLI is fine, if it's more - then perhaps we should add a new workflow for it, no?...

tests/unit/workspace_access/test_generic.py
mwojtyczka: done

mwojtyczka:
The cli commands takes a bit more than 10 minutes in a pretty big workspace. But yes, maybe having a job would actually be better than cli command.

nfx:
@mwojtyczka 10 minutes is fine - but need a job if it's more than 15

nfx:
Technically, this could run as a task after migration in the same workflow

mwojtyczka:
> after mig

Actually we run this verification as part of each apply so theoretically we don't need a separate job or cli at all. 
If someone wants to run it would be more a double check. Being able to run this independently may have some use.

mwojtyczka (approved):

nfx (approved):

PR #868, opened 2024-02-01
[Draft] Migrate Azure Service Principals to UC Storage Credentials

### LINKED ISSUE #339: Migrate Azure Service Principals that access storage to UC Storage Credentials
We should automate ACLs based on Instance Profiles / Service Principals and other legacy security mechanisms.

See the data collected in:
- https://github.com/databrickslabs/ucx/pull/764

Expected flow:

1. check all service principals in `azure_storage_account_info.csv`
2. check all storage credentials
3. see which service principals have matching storage credentials
4. report what credentials are missing
5. prompt-confirm creation of storage credential from service principal
7. give user three options: terraform config, invoke AWS CLI, pick an existing role, create via API

Related issues:

qziyuan:
## Changes
Draft for #339 

### Linked issues

Resolves #339

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/migration/azure_sp.py
```diff
@@ -0,0 +1,162 @@
...
+    client_secret: str
+    # if this is a managed identity
+    if_mi: bool
+
+
+class AzureServicePrincipalMigration:
+
+    def __init__(self, ws: WorkspaceClient, csv: str, replace_with_ac: bool):
+        self._ws = ws
+        self._csv = csv if csv is not None else f"/Users/{ws.current_user.me().user_name}/.ucx/azure_storage_account_info.csv"
```
nfx, line 32: Inject a class from assessment that creates service principal mapping - that has to be the single place that knows how to work with that CSV file

src/databricks/labs/ucx/migration/azure_sp.py
```diff
@@ -0,0 +1,162 @@
...
+    def _load_sp_csv(self):
+        """
+        Load SP info from azure_storage_account_info.csv
+        :return:
+        """
+        #TODO: check the download status
+        csv_source = self._ws.workspace.download(self._csv, format=ExportFormat.AUTO)
+        csv_textio = io.TextIOWrapper(csv_source, encoding='utf-8')
+
+        csv_reader = csv.DictReader(csv_textio)
```
nfx, line 44: Use Installation.load(list[ClassName])

src/databricks/labs/ucx/migration/azure_sp.py
```diff
@@ -0,0 +1,162 @@
...
+    already_in_sc: bool
+    # SP's client_secret stored in Databricks secret
+    client_secret: str
+    # if this is a managed identity
+    if_mi: bool
+
+
+class AzureServicePrincipalMigration:
+
+    def __init__(self, ws: WorkspaceClient, csv: str, replace_with_ac: bool):
```
nfx, line 30: Also inject AzureServicePrincipalCrawler here - that's the place which knows about secret scopes where spn credentials are

src/databricks/labs/ucx/migration/azure_sp.py
```diff
@@ -0,0 +1,162 @@
...
+    principal: str
+    privilege: str
+    # if create access manager and managed identity for this SP
+    replace_with_ac: bool
+    # if a storage credential using this SP already exists
+    already_in_sc: bool
+    # SP's client_secret stored in Databricks secret
+    client_secret: str
+    # if this is a managed identity
+    if_mi: bool
```
nfx, line 25: Doesn't belong to this class

src/databricks/labs/ucx/migration/azure_sp.py
```diff
@@ -0,0 +1,162 @@
...
+        # check if the sp with given client_id is Service Principal or Managed Identity
+        # May use https://learn.microsoft.com/en-us/graph/api/directoryobject-getbyids?view=graph-rest-1.0&tabs=http
+        # Managed Identity should have "@odata.type": "#microsoft.graph.managedIdentity"
+        # https://learn.microsoft.com/en-us/graph/api/resources/managedidentity?view=graph-rest-beta (warning: beta api)
+        # update if_mi field of the sp in sp_list
+        # TODO: find out if azure databricks access connector with system assigned managed identity
+        # TODO: has role assignment on storage, what ID will be returned by providers/Microsoft.Authorization/roleAssignments
+        return
+
+    def _list_sc(self):
```
nfx, line 79: What is SC? Be more explicit with names

src/databricks/labs/ucx/migration/azure_sp.py
```diff
@@ -0,0 +1,162 @@
+import csv
+import io
```
nfx, line 2: Rename file to azure.py or azure_credentials.py

nfx (changes requested):

src/databricks/labs/ucx/migration/azure_sp.py
qziyuan: done

src/databricks/labs/ucx/migration/azure_sp.py
qziyuan: it's for storage credential. Just updated the function name.

src/databricks/labs/ucx/migration/azure_sp.py
qziyuan: done

src/databricks/labs/ucx/migration/azure_sp.py
qziyuan: Do you want it to be moved to `assessment/azure.py`? I think this dataclass will be only used in this migration code.

src/databricks/labs/ucx/migration/azure_sp.py
qziyuan: updated the code, please have a look

src/databricks/labs/ucx/migration/azure_sp.py
qziyuan: updated the code, please have a look

src/databricks/labs/ucx/assessment/azure.py
```diff
@@ -498,6 +503,8 @@ def _map_storage(self, storage: AzureResource) -> list[StoragePermissionMapping]
                     StoragePermissionMapping(
                         prefix=f"abfss://{container.container}@{container.storage_account}.dfs.core.windows.net/",
                         client_id=role_assignment.principal.client_id,
+                        principal_id=role_assignment.principal.object_id,
+                        principal_type=role_assignment.principal.principal_type,
```
qziyuan, line 507: @nfx Please review this change and see if it make sense. It's just a draft so I did not touch the tests yet. 

The idea is the `servicePrincipalType` in `directoryObjects` response would help to determine if it is a service principal ("Application") or a managed identity ("ManagedIdentity"), so my migration class can get this info from the `azure_storage_account_info.csv` and do not need to call the graph api to check again.

Another thing is I need the id/object_id/principal_id to call `directoryObjects` again to fetch the `alternativeNames` field for managed identity, it is the value I have to use to check if this managed identity is already used in UC storage credentials. (The extra `directoryObjects` call can be avoid if the `alternativeNames` is stored in `StoragePermissionMapping`, but I'm concerned it will make the `alternativeNames`  too complex.)

src/databricks/labs/ucx/assessment/azure.py
```diff
@@ -534,3 +541,7 @@ def _get_storage_accounts(self) -> list[str]:
                 if storage_acct not in storage_accounts:
                     storage_accounts.append(storage_acct)
         return storage_accounts
+
+    @property
+    def filename(self):
```
nfx, line 546: Why do we need to expose filename to public? Keep all read/write operations with this file to this class.

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,249 @@
...
+    # if a storage credential using this SP already exists
+    already_in_storage_credential: bool
+    # SP's client_secret stored in Databricks secret
+    client_secret: str
+    # if this is a managed identity
+    if_managed_identity: bool
+
+    @classmethod
+    def from_storage_permission_mapping(cls, storage_permission_mapping, **kwargs):
+        attrs = {field.name: getattr(storage_permission_mapping, field.name) for field in fields(storage_permission_mapping)}
```
nfx, line 32: Don't use reflection! Static code analysis would miss detection bugs in your code.

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,249 @@
...
+        attrs.update(kwargs)
+        return cls(**attrs)
+
+class AzureServicePrincipalMigration:
+
+    def __init__(self, installation: Installation, ws: WorkspaceClient, azure_resource_permissions: AzureResourcePermissions,
+                 azure_sp_crawler: AzureServicePrincipalCrawler, customized_csv: str, replace_with_access_connector: bool):
+        self._final_sp_list = None
```
nfx, line 40: Don't have shared mutable state. We had many bugs here because of it

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,249 @@
...
+        self._installation = installation
+        self._ws = ws
+        self._azure_resource_permissions = azure_resource_permissions
+        self._azure_sp_crawler = azure_sp_crawler
+
+        if customized_csv is not None:
+            self._csv = customized_csv
```
nfx, line 47: Work with files through classes that manage them

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,249 @@
...
+        # check AzureServicePrincipalInfo from AzureServicePrincipalCrawler, if AzureServicePrincipalInfo
+        # has secret_scope and secret_key not empty, fetch the client_secret and put it to the
+        # client_secret field
+
+        # fetch client_secrets of crawled service principal, if any
+        azure_sp_info_with_client_secret = {}
+        azure_sp_infos = self._azure_sp_crawler.snapshot()
+
+        for azure_sp_info in azure_sp_infos:
+            if azure_sp_info.secret_scope and azure_sp_info.secret_key:
```
nfx, line 149: Too much nesting, I refuse to read.

Rewrite "if not ...: continue"

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,249 @@
...
+                try:
+                    secret_response = self._ws.secrets.get_secret(azure_sp_info.secret_scope, azure_sp_info.secret_key)
```
nfx, line 151: Reading secrets is allowed only from DBR, just in case. Account for that.

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,249 @@
...
+
+                    # decode the binary string from GetSecretResponse to utf-8 string
+                    # TODO: handle different encoding if we have feedback from the customer
+                    binary_value_str = secret_response.value
```
nfx, line 155: See ws.dbutils.secrets.get() method

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,249 @@
...
+        return
+
+
+    def generate_migration_list(self):
+        """
+        Create the list of SP that need to be migrated, output an action plan as a csv file for users to confirm
+        :return:
+        """
+        # load sp list from azure_storage_account_info.csv
+        loaded_sp_list = self._load_sp_csv()
```
nfx, line 203: Are you saying you were storing a customer secret in a CSV file in plain text? Are you serious?

qziyuan: Of course not. client_secret will be took out when save the action plan for user to confirmation. The plan is to use a new dataclass that has less info, but more human readable than `ServicePrincipalMigrationInfo` for action plan.

src/databricks/labs/ucx/assessment/azure.py
```diff
@@ -498,6 +503,8 @@ def _map_storage(self, storage: AzureResource) -> list[StoragePermissionMapping]
                     StoragePermissionMapping(
                         prefix=f"abfss://{container.container}@{container.storage_account}.dfs.core.windows.net/",
                         client_id=role_assignment.principal.client_id,
+                        principal_id=role_assignment.principal.object_id,
```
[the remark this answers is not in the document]
qziyuan: The issue is the `directoryObjects` api does not return the resource id which is used in UC. See an example response below. The resource id is actually one of the items in `alternativeNames`. So I need object_id to fetch the `alternativeNames` and try match them with the resource id in UC.

```json
{
    "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#directoryObjects/$entity",
    "@odata.type": "#microsoft.graph.servicePrincipal",
    "@microsoft.graph.tips": "Use $select to choose only the properties your app needs, as this can lead to performance improvements. For example: GET directoryObjects('<guid>')?$select=deletedDateTime",
    "id": "e6c13c39-41ae-46d5-9361-cfa5066e29b9",
    "deletedDateTime": null,
    "accountEnabled": true,
    "alternativeNames": [
        "isExplicit=False",
        "/subscriptions/3f2e4d32-8e8d-46d6-82bc-5bb8d962328b/resourcegroups/pepsi_rg/providers/Microsoft.Databricks/accessConnectors/ag-conn-access-connector"
    ],
    "appDisplayName": null,
    "appDescription": null,
    "appId": "74d98221-94a4-45da-be65-ebe69250e340",
    "applicationTemplateId": null,
    "appOwnerOrganizationId": null,
    "appRoleAssignmentRequired": false,
    "createdDateTime": "2023-08-22T10:51:01Z",
    "description": null,
    "disabledByMicrosoftStatus": null,
    "displayName": "ag-conn-access-connector",
    "homepage": null,
    "loginUrl": null,
    "logoutUrl": null,
    "notes": null,
    "notificationEmailAddresses": [],
    "preferredSingleSignOnMode": null,
    "preferredTokenSigningKeyThumbprint": null,
    "replyUrls": [],
    "servicePrincipalNames": [
        "74d98221-94a4-45da-be65-ebe69250e340",
        "https://identity.azure.net/VjtXYu8/MPezn5Hmx4p3GTmQnxvf3cz2/3MH6meQUeA="
    ],
    "servicePrincipalType": "ManagedIdentity",
    "signInAudience": null,
    "tags": [],
    "tokenEncryptionKeyId": null,
    "info": null,
    "samlSingleSignOnSettings": null,
    "addIns": [],
    "appRoles": [],
    "keyCredentials": [
        {
            "customKeyIdentifier": "CDE40D4DF02E39BDC7DCF61B77D620D60ED16198",
            "displayName": "CN=74d98221-94a4-45da-be65-ebe69250e340",
            "endDateTime": "2024-04-14T09:28:00Z",
            "key": null,
            "keyId": "12a28c38-c149-4e2e-98f8-d3cdc3338223",
            "startDateTime": "2024-01-15T09:28:00Z",
            "type": "AsymmetricX509Cert",
            "usage": "Verify"
        },
        {
            "customKeyIdentifier": "92E575B7BB4C661E65DE483CC784869E754F37AE",
            "displayName": "CN=74d98221-94a4-45da-be65-ebe69250e340",
            "endDateTime": "2024-02-14T07:19:00Z",
            "key": null,
            "keyId": "73ecc81a-39a3-49b6-99e4-3b4df13cb6a0",
            "startDateTime": "2023-11-16T07:19:00Z",
            "type": "AsymmetricX509Cert",
            "usage": "Verify"
        },
        {
            "customKeyIdentifier": "2F7F3E9B72A950227B8F5619062D43ECC9EBDB0A",
            "displayName": "CN=74d98221-94a4-45da-be65-ebe69250e340",
            "endDateTime": "2023-11-20T10:46:00Z",
            "key": null,
            "keyId": "7500be61-6c47-499b-a714-7a9b83ade126",
            "startDateTime": "2023-08-22T10:46:00Z",
            "type": "AsymmetricX509Cert",
            "usage": "Verify"
        }
    ],
    "oauth2PermissionScopes": [],
    "passwordCredentials": [],
    "resourceSpecificApplicationPermissions": [],
    "verifiedPublisher": {
        "displayName": null,
        "verifiedPublisherId": null,
        "addedDateTime": null
    }
}
```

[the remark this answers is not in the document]
qziyuan: we can extract the resource id from the `alternativeNames` and put it in `role_assignment.principal`, so I don't need the object_id and don't need to call `directoryObjects` again in my class. The concern is if this makes `azure_storage_account_info.csv` complex for customer to read?

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: No this is using SecretsAPI not dbutils. The secret value will be returned no matter the runtime. It's not like dbutils which has DBR dbutils and remote dbutils.

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: I have mistake here. I misread the bytes value as binary value, I just need to base64 decode the value, will fix it. 

But the comments for encoding still applies, customer may provide the bytes value instead of string value when create the secrets, and the provided bytes value may came from a different encoding than utf-8 and cause trouble when decode them. For example a bytes value like `T2zhLCBNdW5kbyE=` will not be correctly decode by `ws.dbutils.secrets.get()` as [it assume the charset is 'utf-8'](https://src.dev.databricks.com/databricks/runtime@master/-/blob/sql/core/src/main/scala/com/databricks/service/DBUtils.scala?L270:7&popover=pinned):
```scala
  def get(scope: String, key: String): String = {
    new String(getBytes(scope, key), Charsets.UTF_8)
  }
```
But we don't need to address it now, as I see the possibility of hitting this issue is extremely low.

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: refactored the if logic.

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: updated the code to base64 decode the secret value

src/databricks/labs/ucx/assessment/azure.py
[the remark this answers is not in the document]
nfx: Don't leak internal identifiers!

nfx:
@qziyuan Create an issue specifically for support of managed identities and let's put it out of scope. This trivial feature is already taking a long time to land

qziyuan:
Please review the PR #874 for reduced scope. Close this PR, and will handle the logic for managed identity with new issue.

PR #874, opened 2024-02-03
Migrate Azure Service Principals with secrete stored in Databricks Secret to UC Storage Credentials

### LINKED ISSUE #339: Migrate Azure Service Principals that access storage to UC Storage Credentials

qziyuan:
## Changes

1. In the `secrets.py` file, a method called `_get_secret_if_exists` has been updated to handle cases where the secret value cannot be decoded to a utf-8 string. When a `UnicodeDecodeError` is raised, the method will log a warning message and return `None`.
2. In the `access.py` file, a new class called `StoragePermissionMapping` has been added. This class contains information about a storage permission mapping, including the client ID, principal, privilege, and directory ID.
3. In the `credentials.py` file, several new classes and functions have been added to handle the migration of Azure Service Principals to UC storage credentials. The `ServicePrincipalMigrationInfo` class is used to represent a service principal with its client secret, and the `StorageCredentialValidationResult` class is used to represent the result of validating a storage credential. The `StorageCredentialManager` class is used to manage storage credentials, including listing existing storage credentials, creating storage credentials with a client secret, and validating storage credentials. The `ServicePrincipalMigration` class is used to migrate Azure Service Principals to UC storage credentials.
4. In the `resources.py` file, a new `directory_id` field has been added to the `Principal` class. This field is used to store the directory ID of a principal. Additionally, the `_get_principal` method has been updated to set the `directory_id` field of a principal.
5. In the `cli.py` file, a new command called `migrate_azure_service_principals` has been added. This command is used to migrate Azure Service Principals to UC storage credentials.
6. In the `test_credentials.py` file, several new tests have been added to test the functionality of the `StorageCredentialManager` and `ServicePrincipalMigration` classes.
7. In the `test_resources.py` file, the `Principal` class has been updated to include the `directory_id` field, and the `role_assignments` methods have been updated to use the `Principal` class with the `directory_id` field.
8. In the `conftest.py` file, several new fixtures have been added to test the functionality of the `StorageCredentialManager` and `ServicePrincipalMigration` classes. These fixtures include `StaticServicePrincipalMigration`, `StaticStorageCredentialManager`, `StaticServicePrincipalCrawler`, and `StaticResourcePermissions`.

### Linked issues

Related to #339


src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,170 @@
...
+        azure_sp_infos = self._azure_sp_crawler.snapshot()
+
+        for azure_sp_info in azure_sp_infos:
+            if azure_sp_info.secret_scope is None:
+                continue
+            if azure_sp_info.secret_key is None:
+                continue
+
+            try:
+                secret_response = self._ws.secrets.get_secret(azure_sp_info.secret_scope, azure_sp_info.secret_key)
```
nfx, line 86: Are you 1000% sure this call won't fail outside DBR? Did you test it on labs-azure? **When I implemented dbutils in the sdk, it wasn't the case.**

Create a new task/workflow in runtime.py and call this method of migration from there (within a job).

Then inject WorkspaceInstallation into this class and ".run_workflow()" it from here.

------

@nkvuong this is the context for you why you need to fix the crawler.

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,170 @@
...
+            except Unauthenticated:
+                logger.info(f"User is unauthenticated to fetch secret value. Cannot fetch the service principal "
+                            f"client_secret for {azure_sp_info.application_id}. Will not reuse this client_secret")
+                continue
+            except PermissionDenied:
+                logger.info(f"User does not have permission to read secret value for {azure_sp_info.secret_scope}.{azure_sp_info.secret_key}. "
```
nfx, line 92:
```suggestion
                logger.error (f"User does not have permission to read secret value for {azure_sp_info.secret_scope}.{azure_sp_info.secret_key}. "
```
That's silly...

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,170 @@
...
+                            f"Cannot fetch the service principal client_secret for {azure_sp_info.application_id}. "
```
nfx, line 93: Also, why do we log this particular case? I think we should rethrow, as UCX is run as an workspace admin

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,170 @@
...
+        #
+        # The input StoragePermissionMapping may have managed identity mixed in, we will ignore them for now, as
+        # they won't have any client_secret, we will process managed identity in the future.
+
+        # fetch client_secrets of crawled service principal, if any
+        azure_sp_info_with_client_secret = {}
+        azure_sp_infos = self._azure_sp_crawler.snapshot()
+
+        for azure_sp_info in azure_sp_infos:
+            if azure_sp_info.secret_scope is None:
```
nfx, line 80: Invert the logic, this method is too long: given the spn info instance, retrieve a secret from it and return as string

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,170 @@
...
+                            f"Will not reuse this client_secret")
+
+        # update the list of ServicePrincipalMigrationInfo if client_secret is found
+        for sp in sp_list:
+            if sp.client_id in azure_sp_info_with_client_secret:
+                yield ServicePrincipalMigrationInfo.from_storage_permission_mapping(sp, azure_sp_info_with_client_secret[sp.client_id])
+
+
+    def _save_action_plan(self, sp_list_with_secret) -> str | None:
+        # save action plan to a file for customer to review.
```
nfx, line 124: We may need that file, but for the workflow to pickup.

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,170 @@
...
+        return self._installation.save(sp_list_wo_secret, filename=self._action_plan)
+
+
+    def generate_migration_list(self):
+        """
+        Create the list of SP that need to be migrated, output an action plan as a csv file for users to confirm
+        :return:
+        """
+        # load sp list from azure_storage_account_info.csv
+        loaded_sp_list = self._azure_resource_permissions.load_spn_permission()
```
nfx, line 139: What does "loaded" mean here? 😝

nfx (changes requested):

qziyuan:
Just pushed some new commits not related to above review comments. Will address above comments later.

src/databricks/labs/ucx/migration/azure_credentials.py
nkvuong: this is correct - You create secrets using the REST API or CLI, but you must use the [Secrets utility (dbutils.secrets)](https://docs.databricks.com/en/dev-tools/databricks-utils.html#dbutils-secrets) in a notebook or job to read a secret.

qziyuan: Yes, I have tested fetching secret from field-eng-east workspace in my local laptop with `az login`:
image

nfx: @qziyuan That workspace probably has a custom setting enabled. Try it on labs-azure workspace.

src/databricks/labs/ucx/cli.py
```diff
@@ -273,5 +274,40 @@ def save_aws_iam_profiles(w: WorkspaceClient, aws_profile: str | None = None):
...
+
+    The Azure Service Principals to location mapping are listed in /Users/{user_name}/.ucx/azure_storage_account_info.csv
+    which is generated by save_azure_storage_accounts command. Please review the file and delete the Service Principals
+    you do not want to be migrated.
+
+    The command will only migrate the Service Principals that have client secret stored in Databricks Secret.
+    """
+    logger.info("Running migrate_azure_service_principals command")
+    prompts = Prompts()
+    if not w.config.is_azure:
```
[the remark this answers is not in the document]
qziyuan: done. Will add unit test later

src/databricks/labs/ucx/cli.py
```diff
@@ -273,5 +274,40 @@ def save_aws_iam_profiles(w: WorkspaceClient, aws_profile: str | None = None):
...
+        return
+
+    csv_confirmed = prompts.confirm(f"Have you reviewed the azure_storage_account_info.csv "
+                                    f"and confirm listed service principals are allowed to be checked for migration?")
+    if csv_confirmed is not True:
+        return
+
+    service_principal_migration = AzureServicePrincipalMigration.for_cli(w)
+    action_plan_file = service_principal_migration.generate_migration_list()
+    logger.info("Azure Service Principals subject for migration are checked")
```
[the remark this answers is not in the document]
qziyuan: refactored the code and move them away from cli.

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: @nfx , just tried the labs-azure workspace, same behavior:
image

src/databricks/labs/ucx/cli.py
```diff
@@ -273,5 +274,40 @@ def save_aws_iam_profiles(w: WorkspaceClient, aws_profile: str | None = None):
...
+        logger.error("Workspace is not on azure, please run this command on azure databricks workspaces.")
+        return
+
+    csv_confirmed = prompts.confirm(f"Have you reviewed the azure_storage_account_info.csv "
+                                    f"and confirm listed service principals are allowed to be checked for migration?")
+    if csv_confirmed is not True:
+        return
+
+    service_principal_migration = AzureServicePrincipalMigration.for_cli(w)
+    action_plan_file = service_principal_migration.generate_migration_list()
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: so we may want the workflow to execute the migration, not just using cli? 
For now, I just removed the save action plan code, just print the info to console for user to confirm.

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: split it into two functions now

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,285 @@
...
+        for storage_credential in storage_credentials:
+            # only add service principal's application_id, ignore managed identity based storage_credential
+            if storage_credential.azure_service_principal:
+                storage_credential_app_ids.add(storage_credential.azure_service_principal.application_id)
+        logger.info(
+            f"Found {len(storage_credential_app_ids)} distinct service principals already used in UC storage credentials"
+        )
+        return storage_credential_app_ids
+
+    def _check_sp_in_storage_credentials(self, sp_list, sc_set) -> list[StoragePermissionMapping]:
```
FastLee, line 120: A little cumbersome. You can use list comprehension. Doesn't require a separate method.

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,285 @@
...
+        # if sp is already used, take it off from the sp_list
+        filtered_sp_list = []
+        for service_principal in sp_list:
+            if service_principal.client_id not in sc_set:
+                filtered_sp_list.append(service_principal)
+
+        return filtered_sp_list
+
+    def _read_databricks_secret(self, scope: str, key: str, application_id: str) -> str | None:
+        try:
```
FastLee, line 130: Why do we need to catch all these exceptions can we recover from any of these. We can just call the get_secret and let the user respond to the exception.

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,285 @@
...
+        read_only = False
+        if sp.privilege == Privilege.READ_FILES:
+            read_only = True
+        # create the storage credential
+        storage_credential = self._ws.storage_credentials.create(
+            name=name, azure_service_principal=azure_service_principal, comment=comment, read_only=read_only
+        )
+
+        validation_result = self._validate_storage_credential(storage_credential, sp.prefix)
+        yield validation_result
```
FastLee, line 247: The method is half concrete half generator, it doesn't make sense. Don't use yield

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def _validate_storage_credential(self, storage_credential, location) -> StorageCredentialValidationResult:
+        validation = self._ws.storage_credentials.validate(
+            storage_credential_name=storage_credential.name, url=location
+        )
+        return StorageCredentialValidationResult.from_storage_credential_validation(storage_credential, validation)
+
+    def execute_migration(self, prompts: Prompts):
+        if not self._ws.config.is_azure:
+            logger.error("Workspace is not on azure, please run this command on azure databricks workspaces.")
```
FastLee, line 257: Do all these checks in the "for_cli" method

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,285 @@
...
+            return
+
+        csv_confirmed = prompts.confirm(
```
FastLee, line 260: All the prompts should be part of the for_cli method.

tests/unit/migration/test_azure_credentials.py
```diff
@@ -0,0 +1,93 @@
...
+
+    sp_migration = AzureServicePrincipalMigration(MagicMock(), w, MagicMock(), MagicMock())
+
+    expected = {"b6420590-5e1c-4426-8950-a94cbe9b6115"}
+    sp_migration._list_storage_credentials()
+
+    assert expected == sp_migration._list_storage_credentials()
+
+
+def test_sp_in_storage_credentials():
```
FastLee, line 44: Consider using @pytest.mark.parametrize for multiple runs by the same test.

FastLee (changes requested):

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,285 @@
+import base64
```
nfx, line 1: Please move these files into databricks.labs.ucx.azure folder.

nfx (changes requested):

src/databricks/labs/ucx/migration/azure_credentials.py
```diff
@@ -0,0 +1,285 @@
...
+)
+from databricks.labs.ucx.config import WorkspaceConfig
+from databricks.labs.ucx.framework.crawlers import StatementExecutionBackend
+from databricks.labs.ucx.hive_metastore.locations import ExternalLocations
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ServicePrincipalMigrationInfo(StoragePermissionMapping):
```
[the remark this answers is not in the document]
qziyuan: replaced the dataclass with a namedtuple

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: fixed

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: Removed the catch for authentication and permission denied, as such exception should be rethrow and fail the command. For the rest InternalError and ResourceDoesNotExist we need to catch it and continue the command execution.

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: fixed

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: fixed

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: fixed

tests/unit/migration/test_azure_credentials.py
qziyuan: used @pytest.mark.parametrize in other tests. And this test is actually removed since `_check_sp_in_storage_credentials` is replaced by list comprehension

src/databricks/labs/ucx/migration/azure_credentials.py
qziyuan: done

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,284 @@
...
+                f"Will not reuse this client_secret"
+            )
+            return None
+        except InternalError:
+            logger.info(
+                f"InternalError while reading secret {scope}.{key}. "
+                f"Cannot fetch the service principal client_secret for {application_id}. "
+                f"Will not reuse this client_secret"
+            )
+            print(
```
nfx, line 141: remove print

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,284 @@
...
+                azure_sp_info_with_client_secret[azure_sp_info.application_id] = secret_value
+
+        # update the list of ServicePrincipalMigrationInfo if client_secret is found
+        sp_list_with_secret = []
+        for sp in sp_list:
+            if sp.client_id in azure_sp_info_with_client_secret:
+                sp_list_with_secret.append(
+                    ServicePrincipalMigrationInfo(
+                        service_principal=sp, client_secret=azure_sp_info_with_client_secret[sp.client_id]
+                    )
```
nfx, line 190:
```suggestion
                    ServicePrincipalMigrationInfo(sp, azure_sp_info_with_client_secret[sp.client_id])
```
don't abuse keyword arguments - they hurt readability

nfx (changes requested):

src/databricks/labs/ucx/assessment/azure.py
```diff
@@ -534,3 +542,12 @@ def _get_storage_accounts(self) -> list[str]:
...
         return storage_accounts
+
+    def load(self) -> list[StoragePermissionMapping]:
+        """
+        Load StoragePermissionMapping info from azure_storage_account_info.csv
+        :return:
+        """
+        storage_account_infos = self._installation.load(list[StoragePermissionMapping], filename=self._filename)
+
+        return storage_account_infos
```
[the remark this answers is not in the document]
qziyuan: removed.

src/databricks/labs/ucx/azure/azure_credentials.py
qziyuan: removed.

src/databricks/labs/ucx/azure/azure_credentials.py
qziyuan: fixed

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+        return cls(
+            name=storage_credential.name or "",
+            azure_service_principal=storage_credential.azure_service_principal or AzureServicePrincipal("", "", ""),
+            created_by=storage_credential.created_by or "",
+            read_only=storage_credential.read_only or False,
+            results=validation.results or [],
+        )
+
+
+class AzureServicePrincipalMigration:
```
nfx, line 57: Can you remove Azure prefix from classes defined in this file, as this is clear it's about Azure 😉

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+        # if we are doing integration test
+        if self._integration_test_flag:
+            for storage_credential in storage_credentials:
+                if not storage_credential.azure_service_principal:
+                    continue
+                if self._integration_test_flag == storage_credential.name:
+                    # return the storage credential created during integration test
+                    return {storage_credential.azure_service_principal.application_id}
+            # return no storage credential if there is none created during integration test
+            return {""}
```
nfx, line 114: Why not empty set?

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+            if sp.client_id in azure_sp_info_with_client_secret:
+                sp_list_with_secret.append(
+                    ServicePrincipalMigrationInfo(sp, azure_sp_info_with_client_secret[sp.client_id])
+                )
+        return sp_list_with_secret
+
+    def _print_action_plan(self, sp_list_with_secret: list[ServicePrincipalMigrationInfo]):
+        # print action plan to console for customer to review.
+        for sp in sp_list_with_secret:
+            print(
```
nfx, line 192: Use logger.info

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+        # update the list of ServicePrincipalMigrationInfo if client_secret is found
+        sp_list_with_secret = []
+        for sp in sp_list:
+            if sp.client_id in azure_sp_info_with_client_secret:
+                sp_list_with_secret.append(
+                    ServicePrincipalMigrationInfo(sp, azure_sp_info_with_client_secret[sp.client_id])
+                )
+        return sp_list_with_secret
+
+    def _print_action_plan(self, sp_list_with_secret: list[ServicePrincipalMigrationInfo]):
```
nfx, line 189: Do not print anything with secret!

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+        """
+        # load sp list from azure_storage_account_info.csv
+        sp_list = self._azure_resource_permissions.load()
+        # list existed storage credentials
+        sc_set = self._list_storage_credentials()
+        # check if the sp is already used in UC storage credential
+        filtered_sp_list = [sp for sp in sp_list if sp.client_id not in sc_set]
+        # fetch sp client_secret if any
+        sp_list_with_secret = self._fetch_client_secret(filtered_sp_list)
+        self._final_sp_list = sp_list_with_secret
```
nfx, line 211: Can you rather return this as a result instead of modifying a shared state?
```suggestion
        return sp_list_with_secret
```

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+            directory_id=sp_migration.service_principal.directory_id,
+            application_id=sp_migration.service_principal.client_id,
+            client_secret=sp_migration.client_secret,
+        )
+        comment = f"Created by UCX during migration to UC using Azure Service Principal: {sp_migration.service_principal.principal}"
+        read_only = False
+        if sp_migration.service_principal.privilege == Privilege.READ_FILES.value:
+            read_only = True
+        # create the storage credential
+        storage_credential = self._ws.storage_credentials.create(
```
nfx, line 228: Do you need to use named arguments everywhere?.

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+        return validation_result
+
+    def _validate_storage_credential(
+        self, storage_credential, location: str, read_only: bool
+    ) -> StorageCredentialValidationResult:
+        # storage_credential validation creates a temp UC external location, which cannot overlap with
+        # existing UC external locations. So add a sub folder to the validation location just in case
+        try:
+            validation = self._ws.storage_credentials.validate(
+                storage_credential_name=storage_credential.name, url=location, read_only=read_only
```
nfx, line 244: Can you use positional arguments instead of named?

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+                    is_dir=None,
+                    results=[
+                        ValidationResult(
+                            message="The validation is skipped because an existing external location overlaps with the location used for validation."
+                        )
+                    ],
+                ),
+            )
+
+    def execute_migration(self, prompts: Prompts):
```
nfx, line 263:
```suggestion
    def apply(self, prompts: Prompts):
```
Better name:) or run()

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+            "Above Azure Service Principals will be migrated to UC storage credentials, please review and confirm."
+        )
+        if plan_confirmed is not True:
+            return
+
+        execution_result = []
+        for sp in self._final_sp_list:
+            execution_result.append(self._create_storage_credential(sp))
+
+        results_file = self._installation.save(execution_result, filename=self._output_file)
```
nfx, line 277: Does this results file contain secrets?

tests/integration/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,142 @@
...
+from databricks.labs.ucx.assessment.azure import AzureServicePrincipalInfo
+from databricks.labs.ucx.assessment.crawlers import _SECRET_PATTERN
+from databricks.labs.ucx.azure.access import StoragePermissionMapping
+from databricks.labs.ucx.azure.azure_credentials import AzureServicePrincipalMigration
+
+
+@pytest.fixture
+def prepare_spn_migration_test(ws, debug_env, make_random):
+    def inner(read_only=False) -> dict:
+        spark_conf = ws.clusters.get(cluster_id=debug_env["TEST_LEGACY_SPN_CLUSTER_ID"]).spark_conf
```
nfx, line 17:
```suggestion
        spark_conf = ws.clusters.get(debug_env["TEST_LEGACY_SPN_CLUSTER_ID"]).spark_conf
```
Don't abuse named arguments!

tests/integration/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,142 @@
...
+
+        end_point = spark_conf.get("fs.azure.account.oauth2.client.endpoint")
+        directory_id = end_point.split("/")[3]
+
+        name = f"testinfra_storageaccess_{make_random(4).lower()}"
+
+        azure_resource_permissions = MagicMock()
+        azure_resource_permissions.load.return_value = [
+            StoragePermissionMapping(
+                prefix="abfss://things@labsazurethings.dfs.core.windows.net/avoid_ext_loc_overlap",
```
nfx, line 41: There should be an env var for this

tests/integration/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,142 @@
...
+                secret_key=secret_key,
+                tenant_id="test",
+                storage_account="test",
+            )
+        ]
+
+        installation = MagicMock()
+        installation.save.return_value = "azure_service_principal_migration_result.csv"
+
+        return {
```
nfx, line 63: Don't return it a bunch of variables as untyped dictionary. Refactor to return one or two variables. 

This is impossible to maintain

nfx (changes requested):
First pass

tests/integration/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,142 @@
...
+
+from databricks.labs.ucx.assessment.azure import AzureServicePrincipalInfo
+from databricks.labs.ucx.assessment.crawlers import _SECRET_PATTERN
+from databricks.labs.ucx.azure.access import StoragePermissionMapping
+from databricks.labs.ucx.azure.azure_credentials import AzureServicePrincipalMigration
+
+
+@pytest.fixture
+def prepare_spn_migration_test(ws, debug_env, make_random):
+    def inner(read_only=False) -> dict:
```
nfx, line 16: 1) don't return `dict` - it's broken
2) why do you need `inner`? do you cleanup anything you've created here?

rewrite.

tests/integration/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,142 @@
...
+            assert any(
+                (res.operation is not None) and ("WRITE" in res.operation.value) and ("PASS" in res.result.value)
+                for res in validation_result.results
+            )
+            assert any(
+                (res.operation is not None) and ("DELETE" in res.operation.value) and ("PASS" in res.result.value)
+                for res in validation_result.results
+            )
+    finally:
+        ws.storage_credentials.delete(name=variables["storage_credential_name"], force=True)
```
nfx, line 142: `name` is a positional argument! why do you use it as keyed?

https://databricks-sdk-py.readthedocs.io/en/latest/workspace/catalog/storage_credentials.html#databricks.sdk.service.catalog.StorageCredentialsAPI.create

tests/integration/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,142 @@
...
+        validation_result = spn_migration._installation.save.call_args.args[0][0]
+        if read_only:
+            # We only assert that write validation are not performed for read only storage credential here.
+            # In real life, the READ validation for read only storage credential may fail if there is no file,
+            # but that is fine, as the storage credential is created, and we just cannot validate it until it's really used.
+            assert not any(
+                (res.operation is not None) and ("WRITE" in res.operation.value) for res in validation_result.results
+            )
+        else:
+            assert any(
```
nfx, line 133: this is unreadable

tests/integration/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,142 @@
...
+
+        assert spn_migration._final_sp_list[0].service_principal.principal == variables["storage_credential_name"]
+        assert ws.storage_credentials.get(variables["storage_credential_name"]).read_only is read_only
+
+        validation_result = spn_migration._installation.save.call_args.args[0][0]
+        if read_only:
+            # We only assert that write validation are not performed for read only storage credential here.
+            # In real life, the READ validation for read only storage credential may fail if there is no file,
+            # but that is fine, as the storage credential is created, and we just cannot validate it until it's really used.
+            assert not any(
```
nfx, line 129: this is unreadable. rewrite with for-loops

tests/integration/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,142 @@
...
+        application_id=variables["application_id"],
+        client_secret=variables["client_secret"],
+        directory_id=variables["directory_id"],
+    )
+
+    # test that the spn migration will be skipped due to above storage credential is existed
+    spn_migration = execute_migration(variables, integration_test_flag=variables["storage_credential_name"])
+
+    # because storage_credential is existing, no spn should be migrated
+    assert not spn_migration._final_sp_list
```
nfx, line 111: using private methods is not allowed outside of classes that they're declared in. this is going to break.

tests/integration/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,142 @@
...
+            "azure_sp_crawler": azure_sp_crawler,
+            "installation": installation,
+        }
+
+    return inner
+
+
+@pytest.fixture
+def execute_migration(ws):
+    def inner(variables: dict, integration_test_flag: str) -> AzureServicePrincipalMigration:
```
nfx, line 78: `variables: dict` is an anti-pattern and it's broken.

tests/unit/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,428 @@
...
+    def download(path: str) -> io.BytesIO:
+        if path not in state:
+            raise NotFound(path)
+        return io.BytesIO(state[path].encode('utf-8'))
+
+    ws = create_autospec(WorkspaceClient)
+    ws.config.host = 'https://localhost'
+    ws.current_user.me().user_name = "foo"
+    ws.workspace.download = download
+    ws.statement_execution.execute_statement.return_value = sql.ExecuteStatementResponse(
```
nfx, line 97: use `MockBackend`

tests/unit/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,428 @@
...
+
+    def download(path: str) -> io.BytesIO:
+        if path not in state:
+            raise NotFound(path)
+        return io.BytesIO(state[path].encode('utf-8'))
+
+    ws = create_autospec(WorkspaceClient)
+    ws.config.host = 'https://localhost'
+    ws.current_user.me().user_name = "foo"
+    ws.workspace.download = download
```
nfx, line 96: use `MockInstallation`

tests/unit/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,428 @@
...
+            "prefix": "overlap_with_external_location",
+            "client_id": "app_secret4",
+            "principal": "principal_overlap",
+            "privilege": "WRITE_FILES",
+            "directory_id": "directory_id_2",
+        },
+    ]
+    csv_output = io.StringIO()
+    fieldnames = storage_permission_mappings[0].keys()
+    csv_writer = csv.DictWriter(csv_output, fieldnames=fieldnames, dialect="excel")
```
nfx, line 68: use `MockInstallation`

tests/unit/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,428 @@
...
+    sp_migration._list_storage_credentials()
+    assert expected == sp_migration._list_storage_credentials()
+
+    # test storage credential is not picked up
+    # if integration test does not create storage credential, we use dummy integration_test_flag to filter out other existing storage credentials
+    sp_migration = AzureServicePrincipalMigration(
+        MagicMock(), ws, MagicMock(), MagicMock(), integration_test_flag="other_spn"
+    )
+    sp_migration._list_storage_credentials()
+    assert {""} == sp_migration._list_storage_credentials()
```
nfx, line 181: it's not allowed to call private methods in unit tests. rewrite unit tests to call only public methods.

nfx (changes requested):
many security and maintainability issues

src/databricks/labs/ucx/azure/azure_credentials.py
qziyuan: @nfx Are you sure about this? `AzureServicePrincipal` returned by `WorkspaceClient.storage_credentials.create` suppose to have None `client_secret`. Or is there a config that controls this API behavior that I'm not aware of? I don't believe so. 
But in case in the future bugs introduced to storage_credentials API that may expose client_secret in the response, I can add a guard rail that explicitly set `client_secret` to None before save the migration result.

qziyuan: > why do you need to save a secret 
No we are not saving secret

> why do we need to save this file for installation?
We want recording what have been migrated successfully. Especially for those storage credentials that are created successfully, but failed the validation.

> why do we need a secret there, after we did migration?
We don't need and don't store secret after migration.

src/databricks/labs/ucx/azure/azure_credentials.py
qziyuan: @nfx this function does not print or log any secret.

tests/integration/azure/test_azure_credentials.py
qziyuan: No there is no env var for this, please add one, so this code as well as the [test_access.py](https://github.com/databrickslabs/ucx/blob/f951aac01cecbf70584fb71ef59f06cedd08a6f0/tests/integration/azure/test_access.py#L19C2-L33C92) can use it

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
+import base64
```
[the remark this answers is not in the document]
qziyuan: No there is no env var for this, please add one, so this code as well as the [test_access.py](https://github.com/databrickslabs/ucx/blob/f951aac01cecbf70584fb71ef59f06cedd08a6f0/tests/integration/azure/test_access.py#L19C2-L33C92) can use it

qziyuan:
Refactored the code and integration tests to improve the testability and readability. Will work on unit tests and other comments tomorrow.

src/databricks/labs/ucx/azure/azure_credentials.py
qziyuan: Not for this one. `_ws.storage_credentials.validate` requires keyword arguments.

tests/integration/azure/test_azure_credentials.py
qziyuan: fixed

tests/integration/azure/test_azure_credentials.py
qziyuan: fixed

src/databricks/labs/ucx/azure/azure_credentials.py
qziyuan: changed to run() :)

src/databricks/labs/ucx/azure/azure_credentials.py
qziyuan: fixed

src/databricks/labs/ucx/azure/azure_credentials.py
qziyuan: fixed

src/databricks/labs/ucx/azure/azure_credentials.py
qziyuan: fixed

src/databricks/labs/ucx/azure/azure_credentials.py
qziyuan: fixed

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+from databricks.labs.ucx.azure.resources import AzureResources
+from databricks.labs.ucx.config import WorkspaceConfig
+from databricks.labs.ucx.framework.crawlers import StatementExecutionBackend
+from databricks.labs.ucx.hive_metastore.locations import ExternalLocations
+
+logger = logging.getLogger(__name__)
+
+
+# A namedtuple to host service_principal and its client_secret info
+ServicePrincipalMigrationInfo = namedtuple("ServicePrincipalMigrationInfo", "service_principal client_secret")
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/access.py
```diff
@@ -91,3 +94,9 @@ def _get_storage_accounts(self) -> list[str]:
                 if storage_acct not in storage_accounts:
                     storage_accounts.append(storage_acct)
         return storage_accounts
+
+    def load(self) -> list[StoragePermissionMapping]:
+        """
+        Load StoragePermissionMapping info from azure_storage_account_info.csv
+        """
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+    read_only: bool
+    results: list[ValidationResult]
+
+    @classmethod
+    def from_storage_credential_validation(
+        cls, storage_credential: StorageCredentialInfo, validation: ValidateStorageCredentialResponse
+    ):
+        return cls(
+            name=storage_credential.name or "",
+            azure_service_principal=storage_credential.azure_service_principal or AzureServicePrincipal("", "", ""),
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+            created_by=storage_credential.created_by or "",
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/azure_credentials.py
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+
+class AzureServicePrincipalMigration:
+
+    def __init__(
+        self,
+        installation: Installation,
+        ws: WorkspaceClient,
+        azure_resource_permissions: AzureResourcePermissions,
+        azure_sp_crawler: AzureServicePrincipalCrawler,
+        integration_test_flag="",
```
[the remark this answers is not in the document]
qziyuan: the logic is removed with new StaticStorageCredentialManager

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+        if not ws.config.is_azure:
+            logger.error("Workspace is not on azure, please run this command on azure databricks workspaces.")
+            return None
+
+        csv_confirmed = prompts.confirm(
+            "Have you reviewed the azure_storage_account_info.csv "
+            "and confirm listed service principals are allowed to be checked for migration?"
+        )
+        if csv_confirmed is not True:
+            return None
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/azure_credentials.py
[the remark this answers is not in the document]
qziyuan: the logic is removed with new StaticStorageCredentialManager

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+        storage_credentials = self._ws.storage_credentials.list(max_results=0)
+
+        # if we are doing integration test
+        if self._integration_test_flag:
+            for storage_credential in storage_credentials:
+                if not storage_credential.azure_service_principal:
+                    continue
+                if self._integration_test_flag == storage_credential.name:
+                    # return the storage credential created during integration test
+                    return {storage_credential.azure_service_principal.application_id}
```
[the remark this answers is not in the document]
qziyuan: Please review the new StaticStorageCredentialManager

src/databricks/labs/ucx/azure/azure_credentials.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def _fetch_client_secret(self, sp_list: list[StoragePermissionMapping]) -> list[ServicePrincipalMigrationInfo]:
+        # check AzureServicePrincipalInfo from AzureServicePrincipalCrawler, if AzureServicePrincipalInfo
+        # has secret_scope and secret_key not empty, fetch the client_secret and put it to the client_secret field
+        #
+        # The input StoragePermissionMapping may have managed identity mixed in, we will ignore them for now, as
+        # they won't have any client_secret, we will process managed identity in the future.
+
+        # fetch client_secrets of crawled service principal, if any
+        azure_sp_info_with_client_secret: dict[str, str] = {}
```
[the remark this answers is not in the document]
qziyuan: fixed

tests/integration/azure/test_azure_credentials.py
```diff
@@ -0,0 +1,142 @@
...
+                tenant_id="test",
+                storage_account="test",
+            )
+        ]
+
+        installation = MagicMock()
+        installation.save.return_value = "azure_service_principal_migration_result.csv"
+
+        return {
+            "storage_credential_name": name,
```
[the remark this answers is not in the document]
qziyuan: refactored the integration test

tests/integration/azure/test_azure_credentials.py
qziyuan: 1. replaced with a dataclass. Also separated the logic of fetching the test info like application_id etc. and creating parameters for ServicePrincipalMigration object to run the migration test.
2. `inner` is easier for passing in parameters with different round of tests. Here we do not create anything in the lab-azure env.

tests/integration/azure/test_azure_credentials.py
qziyuan: fixed

tests/integration/azure/test_azure_credentials.py
qziyuan: refactored, please review

tests/integration/azure/test_azure_credentials.py
qziyuan: refactored, please review

tests/integration/azure/test_azure_credentials.py
qziyuan: fixed

tests/integration/azure/test_azure_credentials.py
qziyuan: fixed

qziyuan:
@nfx Please check this [comment](https://github.com/databrickslabs/ucx/pull/874#discussion_r1490205996), we need to sync on this. I don't see the code is exposing client_secret in logs or saved files. Storage credential create api already removes client_secret from response, and now I add an extra check to even eliminate any possibility.

tests/unit/azure/test_azure_credentials.py
qziyuan: done

tests/unit/azure/test_azure_credentials.py
qziyuan: fxied

tests/unit/azure/test_azure_credentials.py
qziyuan: fixed

nfx:
@qziyuan Create a dataclass without a secret field in it and save that

qziyuan:
@nfx I have refactored the `StorageCredentialValidationResult` dataclass so the client_secret field is removed entirely.
Also the `_print_action_plan` now take `StoragePermissionMapping` as input, so this function have zero visibility on client_secret.

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,303 @@
...
+        locations = ExternalLocations(ws, sql_backend, config.inventory_database)
+
+        resource_permissions = AzureResourcePermissions(installation, ws, azurerm, locations)
+        sp_crawler = AzureServicePrincipalCrawler(ws, sql_backend, config.inventory_database)
+
+        storage_credential_manager = StorageCredentialManager(ws)
+
+        return cls(installation, ws, resource_permissions, sp_crawler, storage_credential_manager)
+
+    def _read_databricks_secret(self, scope: str, key: str, application_id: str) -> str | None:
```
nkvuong, line 182: I've created a SecretMixin class in #950, so you reuse/update this method from there

tests/integration/azure/test_credentials.py
```diff
@@ -0,0 +1,143 @@
...
+    return inner
+
+
+def test_spn_migration_existed_storage_credential(extract_test_info, make_storage_credential_from_spn, run_migration):
+    # create a storage credential for this test
+    make_storage_credential_from_spn(
+        credential_name=extract_test_info.credential_name,
+        application_id=extract_test_info.application_id,
+        client_secret=extract_test_info.client_secret,
+        directory_id=extract_test_info.directory_id,
```
nkvuong, line 107: don't need keyword for positional arguments

tests/unit/azure/test_credentials.py
```diff
@@ -0,0 +1,359 @@
...
+    ]
+
+    ws.storage_credentials.create.side_effect = side_effect_create_storage_credential
+    ws.storage_credentials.validate.side_effect = side_effect_validate_storage_credential
+
+    return StorageCredentialManager(ws)
+
+
+def test_list_storage_credentials(credential_manager):
+    assert {"b6420590-5e1c-4426-8950-a94cbe9b6115", "app_secret2"} == credential_manager.list_storage_credentials()
```
nkvuong, line 112: flip this assert statement

nkvuong (commented):
a few minor comments

src/databricks/labs/ucx/azure/credentials.py
qziyuan: great, I applied SecretMixin. I also let `_get_secret_if_exists` to catch the UnicodeDecodeError just in case, please have a look.

tests/integration/azure/test_credentials.py
qziyuan: removed

tests/unit/azure/test_credentials.py
qziyuan: done

tests/integration/conftest.py
```diff
@@ -154,3 +167,50 @@ def load(self):
...
+        for storage_credential in storage_credentials:
+            if not storage_credential.azure_service_principal:
+                continue
+            if storage_credential.name in self._credential_names:
+                application_ids.add(storage_credential.azure_service_principal.application_id)
+
+        logger.info(
+            f"Found {len(application_ids)} distinct service principals already used in storage credentials during integration test"
+        )
+        return application_ids
```
nfx, line 198:
```suggestion
```
use `create_autospec(WorkspaceClient)` and `ws.storage_credentials.list.return_value = [...]` in unit tests. For integration tests, just add `include_names: list[str] | None` argument to `ServicePrincipalMigration.run()` method.

Having `StorageCredentialManager` makes code more complicated than it should

tests/integration/conftest.py
```diff
@@ -154,3 +167,50 @@ def load(self):
...
+        return self._spn_infos
+
+
+class StaticResourcePermissions(AzureResourcePermissions):
+    def __init__(self, permission_mappings: list[StoragePermissionMapping], *args):
+        super().__init__(*args)
+        self._permission_mappings = permission_mappings
+
+    def load(self) -> list[StoragePermissionMapping]:
+        return self._permission_mappings
```
nfx, line 216:
```suggestion
```
you don't need this, there's `MockInstallation`, which can load/save files - https://github.com/databrickslabs/blueprint?tab=readme-ov-file#unit-testing-installation-state

tests/unit/azure/test_credentials.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+@pytest.fixture
+def ws():
+    state = {"/Users/foo/.ucx/config.yml": yaml.dump({'version': 2, 'inventory_database': 'ucx'})}
+
+    def download(path: str) -> io.StringIO:
+        if path not in state:
+            raise NotFound(path)
+        return io.StringIO(state[path])
```
nfx:
```suggestion
```
use mock installation

tests/unit/azure/test_credentials.py
```diff
@@ -0,0 +1,337 @@
...
+
+    ws_mock = create_autospec(WorkspaceClient)
+    ws_mock.config.host = 'https://localhost'
+    ws_mock.current_user.me().user_name = "foo"
+    ws_mock.workspace.download = download
```
nfx, line 60:
```suggestion
```

tests/unit/azure/test_credentials.py
```diff
@@ -0,0 +1,337 @@
...
+        name=name, azure_service_principal=azure_service_principal, comment=comment, read_only=read_only
+    )
+
+
+def side_effect_validate_storage_credential(storage_credential_name, url, read_only):  # pylint: disable=unused-argument
+    if "overlap" in storage_credential_name:
+        raise InvalidParameterValue
+    if read_only:
+        response = {"isDir": True, "results": [{"message": "", "operation": "READ", "result": "PASS"}]}
+        return ValidateStorageCredentialResponse.from_dict(response)
```
nfx, line 75:
```suggestion
        return ValidateStorageCredentialResponse(is_dir=True, results=[...])
```
don't use dictionaries, where you can use typed dataclasses

tests/unit/azure/test_credentials.py
```diff
@@ -0,0 +1,337 @@
...
+    ws_mock.config.host = 'https://localhost'
+    ws_mock.current_user.me().user_name = "foo"
+    ws_mock.workspace.download = download
+    return ws_mock
+
+
+def side_effect_create_storage_credential(name, azure_service_principal, comment, read_only):
+    return StorageCredentialInfo(
+        name=name, azure_service_principal=azure_service_principal, comment=comment, read_only=read_only
+    )
```
nfx, line 67:
```suggestion
```

tests/unit/azure/test_credentials.py
```diff
@@ -0,0 +1,337 @@
...
+            azure_service_principal=AzureServicePrincipal(
+                "62e43d7d-df53-4c64-86ed-c2c1a3ac60c3",
+                "b6420590-5e1c-4426-8950-a94cbe9b6115",
+                "secret",
+            )
+        ),
+        StorageCredentialInfo(azure_service_principal=AzureServicePrincipal("directory_id_1", "app_secret2", "secret")),
+    ]
+
+    ws.storage_credentials.create.side_effect = side_effect_create_storage_credential
```
nfx, line 97:
```suggestion
    ws.storage_credentials.create.return_value = StorageCredentialInfo(
        name=name, azure_service_principal=azure_service_principal, comment=comment, read_only=read_only
    )
```
don't make code more complicated than it should be

tests/unit/azure/test_credentials.py
```diff
@@ -0,0 +1,337 @@
...
+    assert validation.read_only is True
+    assert validation.name == sp_2.permission_mapping.principal
+    for result in validation.results:
+        if result.operation.value == "READ":
+            assert result.result.value == "PASS"
+
+    # prefix used for validation overlaps with existing external location
+    validation = credential_manager.validate_storage_credential(sc_3, sp_3)
+    assert (
+        validation.results[0].message
```
nfx, line 211: definitely simplify this data structure, so that tests are easier to read

nfx (changes requested):

src/databricks/labs/ucx/assessment/secrets.py
```diff
@@ -22,7 +22,10 @@ def _get_secret_if_exists(self, secret_scope, secret_key) -> str | None:
             assert secret.value is not None
             return base64.b64decode(secret.value).decode("utf-8")
         except NotFound:
-            logger.warning(f'removed on the backend: {secret_scope}{secret_key}')
+            logger.warning(f'removed on the backend: {secret_scope}/{secret_key}')
+            return None
+        except UnicodeDecodeError:
+            logger.info(f"Secret {secret_scope}/{secret_key} has Base64 bytes that cannot be decoded to utf-8 string.")
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+
+        for storage_credential in storage_credentials:
+            # only add service principal's application_id, ignore managed identity based storage_credential
+            if storage_credential.azure_service_principal:
+                application_ids.add(storage_credential.azure_service_principal.application_id)
+
+        logger.info(f"Found {len(application_ids)} distinct service principals already used in UC storage credentials")
+        return application_ids
+
+    def create_storage_credential(self, spn: ServicePrincipalMigrationInfo) -> StorageCredentialInfo:
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+
+@dataclass
+class StorageCredentialValidationResult:
+    name: str | None = None
+    application_id: str | None = None
+    directory_id: str | None = None
+    created_by: str | None = None
+    read_only: bool | None = None
+    validated_on: str | None = None
+    results: list[ValidationResult] | None = None
```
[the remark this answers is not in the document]
qziyuan: refactored the code to parse the validation result and only capture failures if any.

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+class ServicePrincipalMigration(SecretsMixin):
+
+    def __init__(
+        self,
+        installation: Installation,
+        ws: WorkspaceClient,
+        resource_permissions: AzureResourcePermissions,
+        sp_crawler: AzureServicePrincipalCrawler,
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+        )
+        read_only = False
+        if spn.permission_mapping.privilege == Privilege.READ_FILES.value:
+            read_only = True
+        # create the storage credential
+        return self._ws.storage_credentials.create(
+            name, azure_service_principal=service_principal, comment=comment, read_only=read_only
+        )
+
+    def validate_storage_credential(
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+        for spn in sp_list_with_secret:
+            storage_credential = self._storage_credential_manager.create_storage_credential(spn)
+            execution_result.append(
+                self._storage_credential_manager.validate_storage_credential(storage_credential, spn)
+            )
+
+        if execution_result:
+            results_file = self.save(execution_result)
+            logger.info("Completed migration from Azure Service Principal migrated to UC Storage credentials")
+            print(
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+        comment = (
+            f"Created by UCX during migration to UC using Azure Service Principal: {spn.permission_mapping.principal}"
+        )
+        read_only = False
+        if spn.permission_mapping.privilege == Privilege.READ_FILES.value:
+            read_only = True
+        # create the storage credential
+        return self._ws.storage_credentials.create(
+            name, azure_service_principal=service_principal, comment=comment, read_only=read_only
+        )
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+            application_id,
+            directory_id,
+            storage_credential.created_by,
+            storage_credential.read_only,
+            prefix,
+            validation.results,
+        )
+
+
+class StorageCredentialManager:
```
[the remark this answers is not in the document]
qziyuan: Yes, the main reason is for StaticCredentialManager for now. But I'm also thinking when we work on the managed identity based storage credential, there will be more logic and it's better to have a dedicated class for it.

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+        return self._ws.storage_credentials.create(
+            name, azure_service_principal=service_principal, comment=comment, read_only=read_only
+        )
+
+    def validate_storage_credential(
+        self, storage_credential: StorageCredentialInfo, spn: ServicePrincipalMigrationInfo
+    ) -> StorageCredentialValidationResult:
+        read_only = False
+        if spn.permission_mapping.privilege == Privilege.READ_FILES.value:
+            read_only = True
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+                storage_credential_name=storage_credential.name,
+                url=spn.permission_mapping.prefix,
+                read_only=read_only,
+            )
+            return StorageCredentialValidationResult.from_validation(
+                storage_credential, validation, spn.permission_mapping.prefix
+            )
+        except InvalidParameterValue:
+            logger.warning(
+                "There is an existing external location overlaps with the prefix that is mapped to the service principal and used for validating the migrated storage credential. Skip the validation"
```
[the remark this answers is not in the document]
qziyuan: fixed

tests/unit/azure/test_credentials.py
qziyuan: refactored

tests/unit/azure/test_credentials.py
qziyuan: fixed

tests/integration/azure/test_credentials.py
```diff
@@ -0,0 +1,140 @@
...
+
+        storage_credential = ws.storage_credentials.get(extract_test_info.credential_name)
+    finally:
+        ws.storage_credentials.delete(extract_test_info.credential_name, force=True)
+
+    assert storage_credential is not None
+    assert storage_credential.read_only is read_only
+
+    # assert the storage credential validation results
+    for res in migration_results[0].results:
```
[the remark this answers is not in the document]
qziyuan: refactored

tests/integration/azure/test_credentials.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # assert no spn migrated since migration_result will be empty
+    assert not migration_result
+
+
+@pytest.mark.parametrize("read_only", [False, True])
+def test_spn_migration(ws, extract_test_info, run_migration, read_only):
+    try:
+        migration_results = run_migration(extract_test_info, {"lets_migrate_the_spn"}, read_only)
+
```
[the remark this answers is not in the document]
qziyuan: fixed

tests/integration/azure/test_credentials.py
```diff
@@ -0,0 +1,140 @@
...
+    def inner(
+        test_info: MigrationTestInfo, credentials: set[str], read_only=False
+    ) -> list[StorageCredentialValidationResult]:
+        installation = Installation(ws, 'ucx')
+        azurerm = AzureResources(ws)
+        locations = ExternalLocations(ws, sql_backend, "dont_need_a_schema")
+
+        permission_mappings = [
+            StoragePermissionMapping(
+                "abfss://things@labsazurethings.dfs.core.windows.net/avoid_ext_loc_overlap",
```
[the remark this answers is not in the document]
qziyuan: yes, we need this "abfss://things@labsazurethings.dfs.core.windows.net/" in env

tests/integration/azure/test_credentials.py
```diff
@@ -0,0 +1,140 @@
...
+
+    return MigrationTestInfo(credential_name, application_id, directory_id, secret_scope, secret_key, client_secret)
+
+
+@pytest.fixture
+def run_migration(ws, sql_backend):
+    def inner(
+        test_info: MigrationTestInfo, credentials: set[str], read_only=False
+    ) -> list[StorageCredentialValidationResult]:
+        installation = Installation(ws, 'ucx')
```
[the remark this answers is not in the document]
qziyuan: fixed

tests/integration/azure/test_credentials.py
```diff
@@ -0,0 +1,140 @@
...
+    secret_key: str
+    client_secret: str
+
+
+@pytest.fixture
+def extract_test_info(ws, debug_env, make_random):
+    random = make_random(6).lower()
+    credential_name = f"testinfra_storageaccess_{random}"
+
+    spark_conf = ws.clusters.get(debug_env["TEST_LEGACY_SPN_CLUSTER_ID"]).spark_conf
```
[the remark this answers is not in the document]
qziyuan: fixed

tests/integration/azure/test_credentials.py
```diff
@@ -0,0 +1,140 @@
...
+    credential_name: str
+    application_id: str
+    directory_id: str
+    secret_scope: str
+    secret_key: str
+    client_secret: str
+
+
+@pytest.fixture
+def extract_test_info(ws, debug_env, make_random):
```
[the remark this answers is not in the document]
qziyuan: fixed

nfx (changes requested):
Few refactorings remain before we can merge this

src/databricks/labs/ucx/cli.py
```diff
@@ -282,5 +283,23 @@ def _aws_principal_prefix_access(w: WorkspaceClient, aws_profile: str):
...
 
 
+@ucx.command
+def migrate_azure_service_principals(w: WorkspaceClient):
+    """Migrate Azure Service Principals, which have Storage Blob Data Contributor,
+    Storage Blob Data Reader, Storage Blob Data Owner roles on ADLS Gen2 locations that are being used in
+    Databricks, to UC storage credentials.
+
+    The Azure Service Principals to location mapping are listed in /Users/{user_name}/.ucx/azure_storage_account_info.csv
+    which is generated by save_azure_storage_accounts command. Please review the file and delete the Service Principals
```
nkvuong, line 293:
```suggestion
    which is generated by principal_prefix_access command. Please review the file and delete the Service Principals
```

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,269 @@
...
+
+        execution_result = []
+        for spn in sp_list_with_secret:
+            storage_credential = self._storage_credential_manager.create_with_client_secret(spn)
+            execution_result.append(self._storage_credential_manager.validate(storage_credential, spn))
+
+        if execution_result:
+            results_file = self.save(execution_result)
+            logger.info(
+                f"Completed migration from Azure Service Principal migrated to UC Storage credentials"
```
nkvuong, line 264:
```suggestion
                f"Completed migration from Azure Service Principal to UC Storage credentials"
```

tests/unit/azure/test_credentials.py
qziyuan: fixed

tests/unit/azure/test_credentials.py
qziyuan: fixed

tests/integration/conftest.py
qziyuan: fixed

tests/integration/conftest.py
qziyuan: fixed

tests/integration/conftest.py
```diff
@@ -154,3 +167,50 @@ def load(self):
 
     def save(self, tables: TablesCrawler, workspace_info: WorkspaceInfo) -> str:
         raise RuntimeWarning("not available")
+
+
+class StaticServicePrincipalMigration(ServicePrincipalMigration):
+    def save(self, migration_results: list[StorageCredentialValidationResult]) -> str:
+        return "azure_service_principal_migration_result.csv"
```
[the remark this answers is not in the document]
qziyuan: fixed

tests/unit/azure/test_credentials.py
```diff
@@ -0,0 +1,385 @@
...
+    assert sp_1.permission_mapping.principal == storage_credential.name
+    assert storage_credential.read_only is False
+
+    storage_credential = credential_manager.create_with_client_secret(sp_2)
+    assert sp_2.permission_mapping.principal == storage_credential.name
+    assert storage_credential.read_only is True
+
+
+def test_validate_storage_credentials(credential_manager):
+    service_principal = MagicMock()
```
nfx, line 188: Why do you need to mock if we have a dataclass?

src/databricks/labs/ucx/cli.py
```diff
@@ -282,5 +283,23 @@ def _aws_principal_prefix_access(w: WorkspaceClient, aws_profile: str):
     logger.info(f"UC roles and bucket info saved {uc_role_path}")
 
 
+@ucx.command
+def migrate_azure_service_principals(w: WorkspaceClient):
```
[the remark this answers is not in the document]
qziyuan: fixed

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+                continue
+            if not sp_info.secret_key:
+                continue
+
+            secret_value = self._get_secret_if_exists(sp_info.secret_scope, sp_info.secret_key)
+
+            if secret_value:
+                sp_info_with_client_secret[sp_info.application_id] = secret_value
+            else:
+                logger.info(
```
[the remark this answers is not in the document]
qziyuan: fixed

tests/unit/azure/test_credentials.py
```diff
@@ -0,0 +1,385 @@
...
+
+    # validate normal storage credential
+    validation = credential_manager.validate(storage_credential, service_principal)
+    assert validation.read_only is False
+    assert validation.name == storage_credential.name
+    assert not validation.failures
+
+
+def test_validate_read_only_storage_credentials(credential_manager):
+    service_principal = MagicMock()
```
nfx, line 209: Same

tests/unit/azure/test_credentials.py
```diff
@@ -0,0 +1,385 @@
...
+
+    # validate read-only storage credential
+    validation = credential_manager.validate(storage_credential, service_principal)
+    assert validation.read_only is True
+    assert validation.name == storage_credential.name
+    assert not validation.failures
+
+
+def test_validate_storage_credentials_overlap_location(credential_manager):
+    service_principal = MagicMock()
```
nfx, line 230: Same

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+# A dataclass to host service_principal info and its client_secret info
+@dataclass
+class ServicePrincipalMigrationInfo:
+    permission_mapping: StoragePermissionMapping
+    client_secret: str
+
+
+@dataclass
+class StorageCredentialValidationResult:
+    name: str | None = None
```
nfx, line 38:
```suggestion
    name: str
```
Cannot be none

nfx (changes requested):
Found a bug. Last one

tests/unit/azure/test_credentials.py
qziyuan: I use mock just to save lines of code populating those dataclass fields that are not actually used. In this case, only the privilege field is needed.

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -0,0 +1,276 @@
...
+    application_id: str | None = None
```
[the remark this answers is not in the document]
qziyuan: right, let me fix this

tests/unit/azure/test_credentials.py
nfx: It's incorrect. You mock behavior, not data. Dataclasses are data. Mocking them is a bug, because these tests won't catch a bug anymore when someone else changes your code

nfx (approved):
Lgtm

PR #942, opened 2024-02-13
Azure service principal crawler bug fix & refactoring

### LINKED ISSUE #896: [BUG]: SPN crawling logic does not work correctly for cluster

### Current Behavior

`test_spn_crawler_with_available_secrets` detects secret id on an existing legacy spn cluster, but could not detect the spn secret stored in secret scope or the storage account

### Expected Behavior

Both secret scope & storage account should be populated

### Steps To Reproduce

_No response_

### Cloud

Azure

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

nkvuong:
## Changes
- Fix tenant_id detection logic In the case where spn endpoint configuration is also inside a secret scope. 
- Refactoring spn crawler logic to make it more readable

### Linked issues
Closes #896

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/assessment/azure.py
```diff
@@ -23,161 +23,113 @@
...
-                    {
-                        "application_id": spn_application_id,
-                        "secret_scope": secret_scope,
-                        "secret_key": secret_key,
-                        "tenant_id": tenant_id,
-                        "storage_account": storage_account,
-                    }
-                )
-        return spn_list
+        return list(generate_service_principals(deduped_service_principals))
```
nfx, line 64: if you make a set of `AzureServicePrincipalInfo` and return those directly instead of working with dictionaries - you don't need a dedupe block - dataclasses already have `__hash__`.

src/databricks/labs/ucx/assessment/azure.py
```diff
@@ -23,161 +23,113 @@
...
+    secret_key: str | None
     # fs.azure.account.oauth2.client.endpoint: "https://login.microsoftonline.com/${local.tenant_id}/oauth2/token"
-    tenant_id: str
+    tenant_id: str | None
     # Azure Storage account to which the SP has been given access
-    storage_account: str
+    storage_account: str | None
+
+
+def generate_service_principals(service_principals: list[dict]):
```
nfx, line 37: we don't need this top-level function ;)

src/databricks/labs/ucx/assessment/azure.py
```diff
@@ -23,161 +23,113 @@
 @dataclass
 class AzureServicePrincipalInfo:
     # fs.azure.account.oauth2.client.id
-    application_id: str
+    application_id: str | None
```
nfx, line 26: isn't it broken if all fields are none? we need at least `application_id`

nfx (changes requested):

src/databricks/labs/ucx/assessment/azure.py
nkvuong: also cleaned up some nonsensical tests

src/databricks/labs/ucx/assessment/azure.py
```diff
@@ -188,45 +137,77 @@ def _safe_get_cluster_policy(self, policy_id: str) -> Policy | None:
...
+                matching_tenant_keys = [key for key in config.keys() if re.search(tenant_key, key)]
+                if len(matching_tenant_keys) == 0 or matching_tenant_keys[0] is None:
+                    tenant_id = ""
+                else:
+                    client_endpoint_list = self._get_key_from_config(matching_tenant_keys[0], config)[0].split("/")
+                    if len(client_endpoint_list) == CLIENT_ENDPOINT_LENGTH:
+                        tenant_id = client_endpoint_list[3]
+
+                # add output to spn list
+                spn_list.append(
```
nfx, line 173: i meant creating the `AzureServicePrincipalInfo` from here directly, without using intermediate dictionary structure. can you change this place and the other couple of places where we construct this dict?

src/databricks/labs/ucx/assessment/azure.py
```diff
@@ -188,45 +137,77 @@ def _safe_get_cluster_policy(self, policy_id: str) -> Policy | None:
...
+
+    def _get_secret_if_exists(self, secret_scope, secret_key) -> str | None:
+        """Get the secret value given a secret scope & secret key. Log a warning if secret does not exist"""
+        try:
+            # Return the decoded secret value in string format
+            secret = self._ws.secrets.get_secret(secret_scope, secret_key)
+            assert secret.value is not None
+            return base64.b64decode(secret.value).decode("utf-8")
+        except NotFound:
+            logger.warning(f'removed on the backend: {secret_scope}"/"{secret_key}')
```
nfx, line 212:
```suggestion
            logger.warning(f'removed on the backend: {secret_scope}/{secret_key}')
```

nfx (changes requested):

nfx (approved):
Lgtm

Issue #941, opened 2024-02-13
[BUG]: Groups in config file not considered for group migration job


jincejames:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When you add the groups to be migrated under 'include_group_names' attribute in the config file after the assessment job has been run, and run the group migration job is executed after this change, all the groups that were detected during the assessment job are renamed/migrated instead of only the groups mentioned under 'include_group_names'. 

### Expected Behavior

The group migration workflow should be migrating only the groups mentioned in the config file and not all of the groups detected during the assessment phase.

### Steps To Reproduce

1. Run the assessment job without 'include_group_names' attribute added in the config file.
2. After the assessment job completes, modify the config file to a subset of groups to be migrated using the 'include_group_names' attribute.
3. Run the group migration workflow.
4. Check the groups migrated/renamed -> You will see that all the groups that were detected as part of assessment job are renamed instead of the ones mentioned in the config file.

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

nfx:
Thank you for the feature request! Currently, the team operates in a limited capacity, carefully prioritizing, and we cannot provide a timeline to implement this feature. Please make a Pull Request if you'd like to see this feature sooner, and we'll guide you through the journey.

We recommend running with all groups by default.

PR #943, opened 2024-02-13
Fixing groups in config file not considered for group migration job

william-conti:
## Changes
If the customer runs the assessment, then changes the configuration file to add a new group, UCX won't be able to reflect the changes.
This change fixes the above scenario and aims towards an incremental group migration.

### Linked issues
Closes https://github.com/databrickslabs/ucx/issues/941

### Tests

- [X] manually tested
- [X] added unit tests


src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -374,7 +374,21 @@ def reflect_account_groups_on_workspace(self):
             raise ManyError(errors)
 
     def get_migration_state(self) -> MigrationState:
-        return MigrationState(self.snapshot())
+        state = MigrationState(self.snapshot())
```
mwojtyczka, line 377: I am not sure this is the best way to solve the issue. It would be better to filter when we fetch the groups (`_fetcher` method). We are already filtering the groups when we crawl. We just need to add simple filter when we fetch as well. This would be one-liner fix.

mwojtyczka (changes requested):
I am not sure this is the best way to solve the issue. It would be better to filter when we fetch the groups (_fetcher method). We are already filtering the groups when we crawl. We just need to add simple filter when we fetch as well. This would be one-liner fix.

william-conti:
@mwojtyczka I would like to identify groups specified in the conf that aren't in the groups table. Today if you specify random groups in the conf it is completely silented.

mwojtyczka:
> @mwojtyczka I would like to identify groups specified in the conf that aren't in the groups table. Today if you specify random groups in the conf it is completely silented.

Got it. It's another use case then.

william-conti:
@mwojtyczka After discussion with Jince, it's better to move the logic to the fetcher() because of what you mention and also for keeping the consistency of the logic. Let's keep things as clean as possible

nfx (approved):

mwojtyczka (approved):

Issue #944, opened 2024-02-14
[BUG]: During installation, ucx does not open config file in the browser even when prompted.


nkvuong:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

During installation, there is a prompt - Open config file in the browser and continue installing? (default: no)
Even when selected yes, no window is open

### Expected Behavior

A window should open

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

PR #945, opened 2024-02-14
Fixing installer not opening config file during the installation

william-conti:
## Changes
`self._installation.save(config)` returns a local path from the workspace.
Changed the installer to build an URL that can be resolved by webbrowser.open()

### Linked issues
fixes https://github.com/databrickslabs/ucx/issues/944

### Tests

- [X ] manually tested
- [X ] added unit tests

### DIFF
--- a/src/databricks/labs/ucx/install.py
+++ b/src/databricks/labs/ucx/install.py
@@ -265,8 +265,9 @@ class WorkspaceInstaller:
             spark_conf=spark_conf_dict,
             policy_id=policy_id,
         )
-        ws_file_url = self._installation.save(config)
-        if self._prompts.confirm("Open config file in the browser and continue installing?"):
+        ws_file = self._installation.save(config)
+        ws_file_url = self._installation.workspace_link(ws_file.replace(self._installation.install_folder(), ""))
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/install.py
```diff
@@ -265,8 +265,9 @@ def warehouse_type(_):
             spark_conf=spark_conf_dict,
             policy_id=policy_id,
         )
-        ws_file_url = self._installation.save(config)
-        if self._prompts.confirm("Open config file in the browser and continue installing?"):
+        ws_file = self._installation.save(config)
+        ws_file_url = self._installation.workspace_link(ws_file.replace(self._installation.install_folder(), ""))
```
nfx, line 269: why do you do that? why not `self._installation.workspace_link('config.yml')` ?

nfx (changes requested):
this looks like a change to https://github.com/databrickslabs/blueprint to me, not to ucx

src/databricks/labs/ucx/install.py
william-conti: Yeah that was irrelevant, although I prefer to use config.__file__ instead in case we change the conf name for whatever reason

PR #964, opened 2024-02-21
Reuse existing UCX Cluster policy if present

HariGS-DB:
## Changes
This change fixes two bugs related to cluster policy setup during ucx installation
 - If a user is rerunning installation (due to some issue previously but some steps completed like policy creation), if a UCX cluster policy is found, it reuses that instead of creating one
 - If the user is upgrading UCX where the initial installation steps are skipped, but the policy ID is not found in the config.yaml (due to manually deleting or upgrading from an older version), then raise an InvalidParameterValue with a custom msg saying policy id not found and request the user to uninstall and reinstall ucx completely.

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [X] manually tested
- [X] added unit tests
- [X] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/install.py
+++ b/src/databricks/labs/ucx/install.py
@@ -243,12 +243,21 @@ class WorkspaceInstaller:
                 cluster_policy = json.loads(self._prompts.choice_from_dict("Choose a cluster policy", cluster_policies))
                 instance_profile, spark_conf_dict = self._get_ext_hms_conf_from_policy(cluster_policy)
 
-        logger.info("Creating UCX cluster policy.")
-        policy_id = self._ws.cluster_policies.create(
-            name=f"Unity Catalog Migration ({inventory_database})",
-            definition=self._cluster_policy_definition(conf=spark_conf_dict, instance_profile=instance_profile),
-            description="Custom cluster policy for Unity Catalog Migration (UCX)",
-        ).policy_id
+        policy_name = f"Unity Catalog Migration ({inventory_database}) ({self._ws.current_user.me().user_name})"
+        policies = self._ws.cluster_policies.list()
+        policy_id = None
+        for policy in policies:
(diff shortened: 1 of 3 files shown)


src/databricks/labs/ucx/install.py
```diff
@@ -243,12 +243,21 @@ def warehouse_type(_):
                 cluster_policy = json.loads(self._prompts.choice_from_dict("Choose a cluster policy", cluster_policies))
                 instance_profile, spark_conf_dict = self._get_ext_hms_conf_from_policy(cluster_policy)
 
-        logger.info("Creating UCX cluster policy.")
-        policy_id = self._ws.cluster_policies.create(
-            name=f"Unity Catalog Migration ({inventory_database})",
-            definition=self._cluster_policy_definition(conf=spark_conf_dict, instance_profile=instance_profile),
-            description="Custom cluster policy for Unity Catalog Migration (UCX)",
-        ).policy_id
+        policy_name = f"Unity Catalog Migration ({inventory_database}) ({self._ws.current_user.me().user_name})"
```
nfx, line 246: Refactor this chunk to a separate method and cover with tests

src/databricks/labs/ucx/install.py
```diff
@@ -243,12 +243,21 @@ def warehouse_type(_):
...
+        policies = self._ws.cluster_policies.list()
+        policy_id = None
+        for policy in policies:
+            if policy.name == policy_name:
+                policy_id = policy.policy_id
+                logger.info(f"Cluster policy {policy_name} already present, reusing the same.")
+                break
```
nfx, line 253: Definitely should have been a method

nfx (changes requested):

nfx (approved):
lgtm

PR #992, opened 2024-02-29
Create UC external locations in Azure based on migrated storage credentials

### LINKED ISSUE #100: Create EXTERNAL LOCATIONs to map to External Tables (Azure)
Related issues:
- https://github.com/databrickslabs/ucx/issues/339
- https://github.com/databrickslabs/ucx/issues/338

Jobs to be done:

- use `ExternalLocations` from https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/hive_metastore/locations.py and `tables.TablesCrawler` to verify if we have matching external locations configured already - use `ws.external_locations.list()` API for this.
- validate the logic of `ExternalLocations` for correctness and corner-cases
- create `databricks labs ucx validate-external-locations` command to display a number of tables per external location and any HMS table locations that don't belong to an existing external location
- print Terraform resource definitions to have those external locations setup

qziyuan:
## Changes

Create UC external locations for missing locations reported by `ExternalLocations` in `hive_metastore/locations.py`.

### Linked issues

Partially resolve #100 

### Functionality 

- [x] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/azure/locations.py
+++ b/src/databricks/labs/ucx/azure/locations.py
@@ -167,7 +167,7 @@ class ExternalLocationsMigration:
             )
             for loc_url in leftover_loc:
                 logger.info(f"Not created external location: {loc_url}")
-        else:
-            logger.info("All UC external location are created.")
+            return leftover_loc
 
+        logger.info("All UC external location are created.")
         return leftover_loc
--- a/src/databricks/labs/ucx/azure/locations.py
+++ b/src/databricks/labs/ucx/azure/locations.py
@@ -3,6 +3,7 @@ from urllib.parse import urlparse
 
 from databricks.labs.blueprint.installation import Installation
 from databricks.sdk import WorkspaceClient
+from databricks.sdk.errors.platform import PermissionDenied
 
 from databricks.labs.ucx.azure.access import AzureResourcePermissions
 from databricks.labs.ucx.azure.resources import AzureResources
@@ -115,13 +116,28 @@ class ExternalLocationsMigration:
             return location_url
         # if no matched write privilege credential, try to create read-only external location
         if container_url in prefix_mapping_read:
-            self._ws.external_locations.create(
-                location_name,
-                location_url,
-                prefix_mapping_read[container_url],
-                comment="Created by UCX",
-                read_only=True,
-            )
+            try:
+                self._ws.external_locations.create(
+                    location_name,
+                    location_url,
+                    prefix_mapping_read[container_url],
+                    comment="Created by UCX",
+                    read_only=True,
+                )
+            except PermissionDenied as denied:
+                if "No file available under the location to read" in str(denied):
+                    # Empty location will cause failed READ permission check with read-only credential
+                    # Skip skip_validation in this case
+                    self._ws.external_locations.create(
+                        location_name,
+                        location_url,
+                        prefix_mapping_read[container_url],
+                        comment="Created by UCX",
+                        read_only=True,
+                        skip_validation=True,
+                    )
+                    return location_url
+                raise denied
             return location_url
         # if no credential found
         return None
--- a/src/databricks/labs/ucx/azure/locations.py
+++ b/src/databricks/labs/ucx/azure/locations.py
@@ -46,7 +46,7 @@ class ExternalLocationsMigration:
         app_id_mapping_read = {}
         all_credentials = self._ws.storage_credentials.list(max_results=0)
         for credential in all_credentials:
-            # cannot have none credential name, it required for external location
+            # cannot have none credential name, it's required for external location
             if not credential.name:
                 continue
             # if service principal based credential, use service principal's application_id directly
@@ -92,11 +92,7 @@ class ExternalLocationsMigration:
         # generate the UC external location name
         before_at, sep, after_at = location_url.partition('@')
         container_name = before_at.removeprefix("abfss://")
-        res_name = (
-            after_at.replace(".dfs.core.windows.net", "")
-            .rstrip("/")
-            .replace("/", "_")
-        )
+        res_name = after_at.replace(".dfs.core.windows.net", "").rstrip("/").replace("/", "_")
         return f"{container_name}_{res_name}"
 
     def _create_external_location(
--- a/src/databricks/labs/ucx/azure/locations.py
+++ b/src/databricks/labs/ucx/azure/locations.py
@@ -90,7 +90,7 @@ class ExternalLocationsMigration:
 
     def _create_location_name(self, location_url: str) -> str:
         # generate the UC external location name
-        before_at, sep, after_at = location_url.partition('@')
+        before_at, _, after_at = location_url.partition('@')
         container_name = before_at.removeprefix("abfss://")
         res_name = after_at.replace(".dfs.core.windows.net", "").rstrip("/").replace("/", "_")
         return f"{container_name}_{res_name}"
--- a/src/databricks/labs/ucx/azure/locations.py
+++ b/src/databricks/labs/ucx/azure/locations.py
@@ -1,0 +1,151 @@
+import logging
+from urllib.parse import urlparse
+from databricks.sdk import WorkspaceClient
+
+from databricks.labs.blueprint.installation import Installation
+from databricks.labs.ucx.azure.access import AzureResourcePermissions
+from databricks.labs.ucx.azure.resources import AzureResources
+from databricks.labs.ucx.config import WorkspaceConfig
+from databricks.labs.ucx.framework.crawlers import StatementExecutionBackend
+from databricks.labs.ucx.hive_metastore import ExternalLocations
+
+logger = logging.getLogger(__name__)
+
+
+class ExternalLocationsMigration:
+    def __init__(
+            self,
+            ws: WorkspaceClient,
+            hms_locations: ExternalLocations,
+            resource_permissions: AzureResourcePermissions,
+            azurerm: AzureResources
+    ):
+        self._ws = ws
+        self._hms_locations = hms_locations
+        self._resource_permissions = resource_permissions
+        self._azurerm = azurerm
+
+    @classmethod
+    def for_cli(cls, ws: WorkspaceClient, installation: Installation):
+        config = installation.load(WorkspaceConfig)
+        sql_backend = StatementExecutionBackend(ws, config.warehouse_id)
+        locations = ExternalLocations(ws, sql_backend, config.inventory_database)
+        azurerm = AzureResources(ws)
+        resource_permissions = AzureResourcePermissions(installation, ws, azurerm, locations)
+
+        return cls(installation, ws, locations, resource_permissions, azurerm)
+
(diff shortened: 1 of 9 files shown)


nfx (changes requested):
logic seems okay at a glance, cover it with tests - it's hard to review the code without them:
image

src/databricks/labs/ucx/azure/locations.py
```diff
@@ -0,0 +1,170 @@
...
+                continue
+            # if service principal based credential, use service principal's application_id directly
+            if credential.azure_service_principal:
+                if not credential.read_only:
+                    app_id_mapping_write[credential.azure_service_principal.application_id] = credential.name
+                    continue
+                if credential.read_only:
+                    app_id_mapping_read[credential.azure_service_principal.application_id] = credential.name
+            # if managed identity based credential, fetch the application_id of the managed identity
+            if credential.azure_managed_identity:
```
nfx, line 60: Nit: This code can be de-nested with if-not-continue

src/databricks/labs/ucx/azure/locations.py
```diff
@@ -0,0 +1,170 @@
...
+            if permission_mapping.client_id in app_id_mapping_write:
+                prefix_mapping_write[permission_mapping.prefix] = app_id_mapping_write[permission_mapping.client_id]
+                continue
+            if permission_mapping.client_id in app_id_mapping_read:
+                prefix_mapping_read[permission_mapping.prefix] = app_id_mapping_read[permission_mapping.client_id]
+        return prefix_mapping_write, prefix_mapping_read
+
+    def _create_location_name(self, location_url: str) -> str:
+        # generate the UC external location name
+        before_at, _, after_at = location_url.partition('@')
```
nfx, line 93: Oh, nice - didn't know about this method in python

src/databricks/labs/ucx/azure/locations.py
```diff
@@ -0,0 +1,170 @@
...
+        # get managed identity/service principal's application id to storage credential name mapping
+        # for all non read-only and read-only credentials
+        app_id_mapping_write, app_id_mapping_read = self._app_id_credential_name_mapping()
+
+        # use the application id to storage credential name mapping to create prefix to storage credential name mapping
+        prefix_mapping_write = {}
+        prefix_mapping_read = {}
+        for permission_mapping in self._resource_permissions.load():
+            if permission_mapping.client_id in app_id_mapping_write:
+                prefix_mapping_write[permission_mapping.prefix] = app_id_mapping_write[permission_mapping.client_id]
```
nfx, line 85: This data structure is incorrect, because if prefix has multiple readers or writers - the last one overrides the previous.

Correct data structure is dict[str,set[str]]

src/databricks/labs/ucx/azure/locations.py
```diff
@@ -0,0 +1,170 @@
...
+        # TODO: considering put this logic into the StorageCredentialManager
+        app_id_mapping_write = {}
+        app_id_mapping_read = {}
+        all_credentials = self._ws.storage_credentials.list(max_results=0)
+        for credential in all_credentials:
+            # cannot have none credential name, it's required for external location
+            if not credential.name:
+                continue
+            # if service principal based credential, use service principal's application_id directly
+            if credential.azure_service_principal:
```
nfx, line 53: Nit: here the code would be more readable if you extract this as a variable. Same for the next condition

src/databricks/labs/ucx/azure/locations.py
```diff
@@ -0,0 +1,170 @@
...
+                if not credential.read_only:
+                    app_id_mapping_write[credential.azure_service_principal.application_id] = credential.name
+                    continue
+                if credential.read_only:
```
nfx, line 57: This statement is redundant

src/databricks/labs/ucx/azure/locations.py
```diff
@@ -0,0 +1,170 @@
...
+        return f"{container_name}_{res_name}"
+
+    def _create_external_location(
+        self, location_url: str, prefix_mapping_write: dict[str, str], prefix_mapping_read: dict[str, str]
+    ) -> str | None:
+        location_name = self._create_location_name(location_url)
+
+        # get container url as the prefix
+        parsed_url = urlparse(location_url)
+        container_url = f"{parsed_url.scheme}://{parsed_url.netloc}/"
```
nfx, line 105: This is loosing data - what if the prefix contains a path?

src/databricks/labs/ucx/azure/locations.py
```diff
@@ -0,0 +1,170 @@
...
+                    )
+                    return location_url
+                raise denied
+            return location_url
+        # if no credential found
+        return None
+
+    def run(self):
+        # list missing external locations in UC
+        missing_locs = [loc.location for loc in self._hms_locations.match_table_external_locations()[1]]
```
nfx, line 143: What is [1] there?

src/databricks/labs/ucx/azure/locations.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # get prefix to storage credential name mapping
+        prefix_mapping_write, prefix_mapping_read = self._prefix_credential_name_mapping()
+
+        # if missing external location is in prefix to storage credential name mapping
+        # create a UC external location with mapped storage credential name
```
nfx, line 149: Missing handling for external locations that are already there - what if this command runs twice? What if locations are created by terraform or by hand? What if only part of them are created by hand?

src/databricks/labs/ucx/azure/locations.py
```diff
@@ -0,0 +1,170 @@
...
+                "External locations below are not created in UC. You may check following cases and rerun this command:"
+                "1. Please check the output of 'migrate_credentials' command for storage credentials migration failure."
+                "2. If you use service principal in extra_config when create dbfs mount or use service principal "
+                "in your code directly for storage access, UCX cannot automatically migrate them to storage credential."
+                "Please manually create those storage credentials first."
+            )
+            for loc_url in leftover_loc:
+                logger.info(f"Not created external location: {loc_url}")
+        else:
+            logger.info("All UC external location are created.")
```
nfx, line 168: Else is redundant

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -242,3 +242,36 @@ def _role_name(self, role_definition_id) -> str | None:
...
+            return None
+        if identity.get("type") == "UserAssigned":
+            if not user_assigned_identity_id:
+                return None
+            identities = identity.get("userAssignedIdentities")
+            if user_assigned_identity_id in identities:
+                return identities.get(user_assigned_identity_id).get("clientId")
+            # sometimes we see "resourceGroups" instead of "resourcegroups" in the response from the API
+            if user_assigned_identity_id.replace("resourcegroups", "resourceGroups") in identities:
+                return identities.get(user_assigned_identity_id.replace("resourcegroups", "resourceGroups")).get(
```
nfx, line 266: Nit: extract a var to make more readable

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -160,7 +160,7 @@ def _get_ext_location_definitions(self, missing_locations: list[ExternalLocation
             cnt += 1
         return tf_script
 
-    def _match_table_external_locations(self) -> tuple[list[list], list[ExternalLocation]]:
+    def match_table_external_locations(self) -> tuple[list[list], list[ExternalLocation]]:
```
nfx, line 163: Return type is incorrect. Add more unit tests for this method if you make it public

tests/integration/azure/test_locations.py
```diff
@@ -0,0 +1,131 @@
...
+        # If test failed with exception threw before external location is created,
+        # don't fail the test with external location cannot be deleted error,
+        # instead let the original exception be reported.
+        pass
+
+
+@pytest.mark.skip
+def test_run(caplog, ws, sql_backend, inventory_schema):
+    locations = [
+        ExternalLocation("abfss://uctest@ziyuanqintest.dfs.core.windows.net/one", 1),
```
nfx, line 25: Okay, as a follow-up after this PR is merged, make a PR to labs-infra which will setup these with Terraform. Or maybe within this PR. I'll do you a quick intro

nfx (changes requested):

src/databricks/labs/ucx/azure/locations.py
qziyuan: wWell this is intentional. We cannot have multiple UC external locations pointing to same prefix, so when multiple credentials have access to one same prefix, we can only pick one credential to create the external location. So the logic is first try pick one that has write access, if not pick one that has read access. So the data structure does not need to contain all readers and writers

src/databricks/labs/ucx/azure/locations.py
qziyuan: `access.py` crawls [container to principal mapping](https://github.com/databrickslabs/ucx/blob/e0da29aee2c10e91f01b1c315812fa0c3e3eadf9/src/databricks/labs/ucx/azure/access.py#L79), so for a given location we extract the container url and use it as the key to find the principal and hence the storage credential to be used.

tests/integration/azure/test_locations.py
qziyuan: The storage account, managed identities etc. I'm using here belong to the field-eng subscription and they are subject to be removed with `removeAfter` tag.
If we use Terraform to populate them in labs-infra, the problem is I don't have read permission for Azure resources in labs-infra, those azure graph and resource manager api call to get application_id will fail.

src/databricks/labs/ucx/azure/locations.py
qziyuan: The `match_table_external_locations()` will first call UC apt to list existed UC external locations. So the returned `missing_locs` will not include any created UC external locations no matter how many times this command runs.

src/databricks/labs/ucx/azure/resources.py
qziyuan: fixed.

src/databricks/labs/ucx/azure/locations.py
qziyuan: fixed

src/databricks/labs/ucx/azure/locations.py
qziyuan: fixed.

src/databricks/labs/ucx/azure/locations.py
qziyuan: fixed.

src/databricks/labs/ucx/azure/locations.py
qziyuan: fixed.

src/databricks/labs/ucx/azure/locations.py
qziyuan: refactored for better readability

nfx (approved):
Lgtm

Issue #928, opened 2024-02-08
[FEATURE]: Command to re-map cluster type


chase-edwards-db:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

UC migrations require that users move from non-UC to UC cluster types. This is done manually today, which is tedious and requires significant time investment. A migration tool for cluster types would accelerate this component of the migration process.

### Proposed Solution

This feature should:

- Account for clusters and policies
- Map the above to their UC counterparts

### Additional Context

This is a nuanced problem, due to the differences in capability in UC clusters. The tool may require guidance from a field engineering architect, but the tool could suggest changes at a minimum.

nfx:
@chase-edwards-db This can easily be implemented as CLI command. Get an RSA for a few days to do it and we'll guide how

chase-edwards-db:
Appreciate that I didn't get to this until later, saw it came in. Thanks UCX team!

PR #994, opened 2024-03-01
Added `databricks labs ucx cluster-remap` command to remap legacy cluster configurations to UC-compatible

prajin-29:
## Changes
Adding Cli command to Remap the cluster to UC `databricks labs ucx cluster-remap`

### Linked issues
#928 

Resolves #..

### Functionality 

- [x] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


nkvuong:
@prajin-29 need to understand how this command can be used - in its current state, users have to specify individual cluster_id to be remapped - this won't be of much use to our users

prajin-29:
> @prajin-29 need to understand how this command can be used - in its current state, users have to specify individual cluster_id to be remapped - this won't be of much use to our users

So in that case should the command iterates through all the clusters and convert that to UC enabled? Instead of specifying the cluster id.

src/databricks/labs/ucx/install.py
```diff
@@ -906,6 +906,21 @@ def _repair_workflow(self, workflow):
         run_id = latest_job_run.run_id
         return job_id, run_id
 
+    def map_cluster_to_uc(self, cluster_id):
```
nfx, line 909: it doesn't belong to `WorkspaceInstallation` class. for now, create the `workspace_access/clusters.py` and add this remapping logic there. we'll be expanding that class a lot

src/databricks/labs/ucx/install.py
```diff
@@ -906,6 +906,21 @@ def _repair_workflow(self, workflow):
...
+        try:
+            spark_version = self._ws.clusters.select_spark_version(latest=True)
+            security_modes = {"Single User": "SINGLE_USER", "Shared": "USER_ISOLATION"}
```
nfx, line 912:
```suggestion
            security_modes = {"Single User": compute.DataSecurityMode.SINGLE_USER, "Shared": compute.DataSecurityMode.USER_ISOLATION}
```

nfx (changes requested):

qziyuan:
> > @prajin-29 need to understand how this command can be used - in its current state, users have to specify individual cluster_id to be remapped - this won't be of much use to our users
> 
> So in that case should the command iterates through all the clusters and convert that to UC enabled? Instead of specifying the cluster id.

I'm wondering if it is a good choice to directly edit existing clusters, which may break codes and query running there. For example AT&T had a lot of jobs failed just because they switch on UC and we automatically change all new job clusters to be single user. We should be careful here. In reality, many customer create new UC clusters and gradually migrate their codes to these new clusters. 
I think this cli will be used in following scenarios:

1. The users has migrated and tested that their codes and queries, and confirm they want to convert all their clusters into UC. Then we do in place cluster edit in bulk.
2. The users just want a UC cluster that is a counterpart of their existing cluster, and use it to test their codes. Then we should create one or list of new UC clusters.

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,54 @@
...
+        self._prompts = prompts
+
+    @classmethod
+    def current(cls, ws: WorkspaceClient):
+        prompts = Prompts()
+        return ClusterAccess(prompts, ws)
+
+    def map_cluster_to_uc(self, cluster_id: str | None = None):
+        spark_version = self._ws.clusters.select_spark_version(latest=True)
+        security_modes = {"Single User": DataSecurityMode.SINGLE_USER, "Shared": DataSecurityMode.USER_ISOLATION}
```
nfx, line 23: use this for automated upgrade logic: https://docs.databricks.com/en/archive/compute/cluster-ui-preview.html#how-does-backward-compatability-work-with-these-changes

tests/integration/workspace_access/test_clusters.py
```diff
@@ -0,0 +1,32 @@
...
+from databricks.sdk.errors import NotFound
+from databricks.sdk.retries import retried
+
+from databricks.labs.ucx.workspace_access.clusters import ClusterAccess
+
+logger = logging.getLogger(__name__)
+
+
+@retried(on=[NotFound], timeout=timedelta(minutes=5))
+def test_prepare_environment(ws, make_cluster):
```
nfx, line 14: integration test is probably not required for this :)

nfx (changes requested):

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,54 @@
...
+        try:
+            if cluster_id is None:
+                msg = "Cluster Id is not Provided. Please provide the cluster id"
+                raise InvalidParameterValue(msg)
+            access_mode = self._prompts.choice_from_dict("Select cluster access mode", security_modes)
+            cluster_details = self._ws.clusters.get(cluster_id)
+            self._ws.clusters.edit(
```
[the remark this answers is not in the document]
prajin-29: Updated the Cli command Logic

tests/integration/workspace_access/test_clusters.py
prajin-29: Removed Integration test

src/databricks/labs/ucx/cli.py
```diff
@@ -504,5 +505,43 @@ def create_catalogs_schemas(w: WorkspaceClient, prompts: Prompts):
...
+    cluster_ids = [
+        cluster_files.path.split("/")[-1].split(".")[0]
+        for cluster_files in installation.files()
+        if cluster_files.path is not None and cluster_files.path.find("backup/clusters") > 0
+    ]
+    if not cluster_ids:
+        logger.info("There is no cluster files in the backup folder.Skipping the reverting process")
+        return
+    for cluster in cluster_ids:
+        print(cluster)
```
dmoore247, line 538: Prob should use logger.info instead of print

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,117 @@
...
+        for cluster in clusters:
+            if cluster.cluster_source is not None and cluster.cluster_source.name != "JOB":
+                cluster_list[cluster.cluster_name] = cluster.cluster_id
+        return cluster_list
+
+    def _get_cluster_id(self, cluster_id: str):
+        if cluster_id != "<ALL>":
+            return [x.strip() for x in cluster_id.split(",")]
+        cluster_list = []
+        clusters = self._ws.clusters.list()
```
dmoore247, line 30: Using the Request parameter to perform remote filtering of results, reducing network, speeding api call `w.clusters.list(compute.ListClustersRequest())`
Eliminating some logic below

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,117 @@
...
+        spark_version = self._ws.clusters.select_spark_version(latest=True)
+        for cluster in cluster_id_list:
+            try:
+                cluster_details = self._ws.clusters.get(cluster)
+                if cluster_details.data_security_mode is None:
+                    logger.info(f"Data security Mode is None. Skipping the remapping for the cluster: {cluster}")
+                    continue
+                access_mode = self._get_access_mode(cluster_details.data_security_mode.name)
+                self._installation.save(cluster_details, filename=f'backup/clusters/{cluster_details.cluster_id}.json')
+                self._ws.clusters.edit(
```
dmoore247, line 53: Is there a way to make this resilient to cluster API changes? The Cluster UI and options are constantly changing, how can this code just focus on the specifics of credentials, spark configs and data security mode and pass through all the other configuration stuff. Will this code break frequently? [I don't know]

src/databricks/labs/ucx/cli.py
```diff
@@ -504,5 +505,43 @@ def create_catalogs_schemas(w: WorkspaceClient, prompts: Prompts):
...
+    cluster_list = prompts.question(
+        "Please provide the cluster id's as comma separated value from the above list", default="<ALL>"
+    )
+    cluster_class = ClusterAccess(installation, w, prompts)
```
dmoore247, line 542: Technically you're working with an instance of the cluster access (object) and not a class.

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,117 @@
...
+
+class ClusterAccess:
+    def __init__(self, installation: Installation, ws: WorkspaceClient, prompts: Prompts):
+        self._ws = ws
+        self._prompts = prompts
+        self._installation = installation
+
+    def list_cluster(self):
+        cluster_list = {}
+        clusters = self._ws.clusters.list()
```
dmoore247, line 20: Use the Request class to filter the results on the remote side
`all = w.clusters.list(compute.ListClustersRequest())`
https://databricks-sdk-py.readthedocs.io/en/latest/workspace/compute/clusters.html

src/databricks/labs/ucx/workspace_access/clusters.py
dmoore247: #DRY - dont repeat yourself

Doesn't make sense to have two independent calls to the` _ws.clusters.list()` API, for one to 'project' IDs and the other to bring back the entire object.
One should call the other, and do list comprehension to return the list id properties.

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def _get_access_mode(self, access_mode: str):
+        if access_mode in {"LEGACY_SINGLE_USER", "SINGLE_USER"}:
+            return DataSecurityMode.SINGLE_USER
+        return DataSecurityMode.USER_ISOLATION
+
+    def map_cluster_to_uc(self, cluster_id: str):
+
+        cluster_id_list = self._get_cluster_id(cluster_id)
+        spark_version = self._ws.clusters.select_spark_version(latest=True)
```
dmoore247, line 44: Is this latest point release version or latest LTS version. We really want latest LTS version.

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,117 @@
...
+                    aws_attributes=cluster_details.aws_attributes,
+                    ssh_public_keys=cluster_details.ssh_public_keys,
+                    enable_elastic_disk=cluster_details.enable_elastic_disk,
+                    cluster_source=cluster_details.cluster_source,
+                    instance_pool_id=cluster_details.instance_pool_id,
+                    enable_local_disk_encryption=cluster_details.enable_local_disk_encryption,
+                    driver_instance_pool_id=cluster_details.driver_instance_pool_id,
+                )
+            except InvalidParameterValue as e:
+                logger.warning(f"skipping cluster remapping: {e}")
```
dmoore247, line 77: Please make sure the context is included with the error message. 
What cluster_id caused the exception. What parameter caused the exception?

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,117 @@
...
+        if cluster_ids != "<ALL>":
+            cluster_list = [x.strip() for x in cluster_ids.split(",")]
+        else:
+            cluster_list = total_cluster_ids
+        logger.info(f"Reverting the configurations for the cluster {cluster_list}")
+        for cluster in cluster_list:
+            try:
+                cluster_details = self._installation.load(ClusterDetails, filename=f"/backup/clusters/{cluster}.json")
+                if cluster_details.spark_version is None:
+                    raise InvalidParameterValue("cluster does not have spark version")
```
dmoore247, line 89: Include the filepath as context. 
This the intent to blow up the entire process, or to just note one might have failed. I hope the latter.

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,117 @@
...
+                if cluster_details.cluster_id is None:
+                    raise InvalidParameterValue("cluster Id is not present in the config file")
```
dmoore247, line 91: Include the filepath as context in the error.
This the intent to blow up the entire process, or to just note one might have failed. I hope the latter. Just use a logger.error.
Context is important in shortening the troubleshooting to resolution cycle.

tests/unit/workspace_access/test_clusters.py
```diff
@@ -0,0 +1,41 @@
+from unittest.mock import create_autospec
+
+from databricks.labs.blueprint.installation import Installation
+from databricks.labs.blueprint.tui import MockPrompts
+from databricks.sdk import WorkspaceClient
+from databricks.sdk.service.compute import ClusterDetails
+
+from databricks.labs.ucx.workspace_access.clusters import ClusterAccess
+
+
```
dmoore247, line 10: Please test the unhappy paths indicated by the `raised` exceptions in the code.

dmoore247 (changes requested):
1. Clarify the code
2. DRY
3. Log problems as ERROR and continue processing instead of sometimes blowing up the list processing.

src/databricks/labs/ucx/cli.py
prajin-29: We are not logging here. It's something we are displaying to the user the list of clusters from which they can provide the input.

src/databricks/labs/ucx/workspace_access/clusters.py
prajin-29: It will be the latest LTS version.

src/databricks/labs/ucx/workspace_access/clusters.py
nfx: @dmoore247 That class doesn't exist.

src/databricks/labs/ucx/workspace_access/clusters.py
nfx: Then you need to use "long_term_support" argument or something like that

src/databricks/labs/ucx/cli.py
```diff
@@ -504,5 +505,43 @@ def create_catalogs_schemas(w: WorkspaceClient, prompts: Prompts):
...
+@ucx.command
+def cluster_remap(w: WorkspaceClient, prompts: Prompts):
+    """Re-mapping the cluster to UC"""
+    logger.info("Remapping the Clusters to UC")
+    installation = Installation.current(w, 'ucx')
+    cluster = ClusterAccess(installation, w, prompts)
+    cluster_list = cluster.list_cluster()
+    print("Cluster Name\tCluster Id")
+    for name, cluster_id in cluster_list.items():
+        print(f"{name}\t{cluster_id}")
```
nfx, line 517: Can you provide a screenshot of this working as a table?

Otherwise refactor.

src/databricks/labs/ucx/cli.py
nfx: Use logger. We're not displaying tables here

nfx (changes requested):

src/databricks/labs/ucx/workspace_access/clusters.py
prajin-29: This is to note the cluster is failed to migrate.

src/databricks/labs/ucx/cli.py
prajin-29: image
I have adjusted the spacing.

tests/unit/workspace_access/test_clusters.py
prajin-29: Added test case for unhappy path also

src/databricks/labs/ucx/workspace_access/clusters.py
prajin-29: Included the cluster id so that we can track by cluster

src/databricks/labs/ucx/workspace_access/clusters.py
prajin-29: Included the cluster id so that we can track by cluster.

src/databricks/labs/ucx/cli.py
```diff
@@ -504,5 +505,43 @@ def create_catalogs_schemas(w: WorkspaceClient, prompts: Prompts):
...
 
 
+@ucx.command
+def cluster_remap(w: WorkspaceClient, prompts: Prompts):
+    """Re-mapping the cluster to UC"""
+    logger.info("Remapping the Clusters to UC")
+    installation = Installation.current(w, 'ucx')
+    cluster = ClusterAccess(installation, w, prompts)
+    cluster_list = cluster.list_cluster()
+    print(f"{'Cluster Name':<80}\t{'Cluster Id':<80}")
```
nfx, line 515: assuming that the average screen width is 80 characters, isn't it too long?...

src/databricks/labs/ucx/cli.py
```diff
@@ -504,5 +505,43 @@ def create_catalogs_schemas(w: WorkspaceClient, prompts: Prompts):
...
+    """Reverting Re-mapping of  clusters from UC"""
+    logger.info("Reverting the Remapping of the Clusters from UC")
+    installation = Installation.current(w, 'ucx')
+    cluster_ids = [
+        cluster_files.path.split("/")[-1].split(".")[0]
+        for cluster_files in installation.files()
+        if cluster_files.path is not None and cluster_files.path.find("backup/clusters") > 0
+    ]
+    if not cluster_ids:
+        logger.info("There is no cluster files in the backup folder.Skipping the reverting process")
```
nfx, line 535:
```suggestion
        logger.info("There is no cluster files in the backup folder. Skipping the reverting process")
```
have consistent typography

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,119 @@
...
+        self._ws = ws
+        self._prompts = prompts
+        self._installation = installation
+
+    def list_cluster(self):
+        cluster_list = {}
+        clusters = self._ws.clusters.list()
+        for cluster in clusters:
+            if cluster.cluster_source is not None and cluster.cluster_source.name != "JOB":
+                cluster_list[cluster.cluster_name] = cluster.cluster_id
```
nfx, line 23: you know that clusters don't have unique names? this means that you can't have them as dict keys

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,119 @@
...
+                cluster_list.append(cluster.cluster_id)
+        return cluster_list
+
+    def _get_access_mode(self, access_mode: str):
+        if access_mode in {"LEGACY_SINGLE_USER", "SINGLE_USER"}:
+            return DataSecurityMode.SINGLE_USER
+        return DataSecurityMode.USER_ISOLATION
+
+    def map_cluster_to_uc(self, cluster_id: str):
+
```
nfx, line 42:
```suggestion
```

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,119 @@
...
+    def revert_cluster_remap(self, cluster_ids: str, total_cluster_ids: list):
+        if cluster_ids != "<ALL>":
+            cluster_list = [x.strip() for x in cluster_ids.split(",")]
+        else:
+            cluster_list = total_cluster_ids
+        logger.info(f"Reverting the configurations for the cluster {cluster_list}")
+        for cluster in cluster_list:
+            try:
+                cluster_details = self._installation.load(ClusterDetails, filename=f"/backup/clusters/{cluster}.json")
+                if cluster_details.spark_version is None:
```
nfx, line 88: spark version may be absent if cluster is using a policy, so this line is not necessary

src/databricks/labs/ucx/workspace_access/clusters.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def revert_cluster_remap(self, cluster_ids: str, total_cluster_ids: list):
+        if cluster_ids != "<ALL>":
+            cluster_list = [x.strip() for x in cluster_ids.split(",")]
+        else:
+            cluster_list = total_cluster_ids
+        logger.info(f"Reverting the configurations for the cluster {cluster_list}")
+        for cluster in cluster_list:
+            try:
+                cluster_details = self._installation.load(ClusterDetails, filename=f"/backup/clusters/{cluster}.json")
```
nfx, line 87:
```suggestion
                cluster_details = self._installation.load(ClusterDetails, filename=f"backup/clusters/{cluster}.json")
```
remove first forward slash. it'll confuse the installation

nfx (changes requested):

src/databricks/labs/ucx/workspace_access/clusters.py
prajin-29: Now the API call is happening only once. filtering the cluster using the list comprehension

src/databricks/labs/ucx/cli.py
prajin-29: image
Reduced the width to 50. It looks fine.

src/databricks/labs/ucx/workspace_access/clusters.py
prajin-29: Removed the forward slash

src/databricks/labs/ucx/cli.py
prajin-29: Done

src/databricks/labs/ucx/workspace_access/clusters.py
prajin-29: done

src/databricks/labs/ucx/workspace_access/clusters.py
prajin-29: Changed the logic. Now hitting the API only once . Using the result to filter out the cluster based on selection

nfx (approved):
looking good and well-covered with tests - would be easy to enhance after. Make sure integration tests pass

PR #995, opened 2024-03-01
Added group members difference to the output of validate-groups-membership cli command

mwojtyczka:
The `validate-groups-membership` command has been updated to include a comparison of group memberships at both the account and workspace levels, displaying the difference in members between the two levels in a new column. This enhancement allows for a more detailed analysis of group memberships, with the added functionality implemented in the `validate_group_membership` function in the `groups.py` file located in the `databricks/labs/ucx/workspace_access` directory. A new output field, "group\_members\_difference," has been added to represent the difference in the number of members between a workspace group and an associated account group. The corresponding unit test file, "test\_groups.py," has been updated to include a new test case that verifies the calculation of the "group\_members\_difference" value. This change provides users with a more comprehensive view of their group memberships and allows them to easily identify any discrepancies between the account and workspace levels. The functionality of the other commands remains unchanged.

### DIFF
--- a/labs.yml
+++ b/labs.yml
@@ -109,10 +109,10 @@ commands:
         description: AWS Profile to use for authentication
 
   - name: validate-groups-membership
-    description: Validate the groups to see if the groups at account level and workspace level have different membership
+    description: Validate groups to check if the groups at account level and workspace level have different memberships
     table_template: |-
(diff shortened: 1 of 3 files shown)


nfx (approved):
lgtm

PR #1003, opened 2024-03-04
Reducing flakiness of create account groups

william-conti:
## Changes
test_create_account_level_groups generally fails during the integration tests because the pattern that the code follows is not safe when there are multiple creation and deletion of groups in parallel.

The proposed changes assumes that all the groups.get() and groups.create() calls are not safe, thus they are surrounded by a try/except

### DIFF
--- a/src/databricks/labs/ucx/account.py
+++ b/src/databricks/labs/ucx/account.py
@@ -5,7 +5,7 @@ import requests
 from databricks.labs.blueprint.installation import Installation
 from databricks.labs.blueprint.tui import Prompts
 from databricks.sdk import AccountClient, WorkspaceClient
-from databricks.sdk.errors import NotFound
+from databricks.sdk.errors import NotFound, ResourceConflict
 from databricks.sdk.service.iam import ComplexValue, Group, Patch, PatchOp, PatchSchema
 from databricks.sdk.service.provisioning import Workspace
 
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/account.py
```diff
@@ -183,13 +190,24 @@ def _get_account_groups(self) -> dict[str | None, list[ComplexValue] | None]:
...
+            full_account_group = self._safe_groups_get(self._ac, acc_grp_id.id)
+            if not full_account_group:
+                continue
+            logger.debug(f"Found account group {full_account_group.display_name}")
             acc_groups[full_account_group.display_name] = full_account_group.members
 
         logger.info(f"{len(acc_groups)} account groups found")
         return acc_groups
 
+    def _safe_groups_get(self, interface, group_id) -> Group | None:
```
nkvuong, line 202: technically the issue is that Identity API is eventually consistent, so we need to retry with exponential backoff, and the upper limit should be something reasonable like 2m

william-conti: This feature is not a heavy create/get workload where indeed it would make sense to retry in this case.

The problem is that the code is doing a list, then iterates over the result of that list to get the group details. But it can happen that the group has been deleted after it was captured by the list command, then the get would fail.

This scenario happens because in the case of our integration tests, we create/delete a huge amount of groups in parallel.

nfx (approved):
Lgtm

PR #1019, opened 2024-03-07
Increased test coverage for verification suite

aminmovahed-db:
## Changes
**Five unite tests have been added to test_verification.py:**
test_verify
test_verify_applied_scope_acls
test_verify_applied_permissions_fails
test_fail_verify_roles_and_entitlements_missing_second_object
test_verify_roles_and_entitlements

This increases the unit test coverage of verification.py from 57% to 95%:

Screenshot 2024-03-10 at 2 55 34 pm
Screenshot 2024-03-10 at 2 55 00 pm


tests/unit/workspace_access/test_verification.py
```diff
@@ -79,3 +79,114 @@ def test_fail_get_permissions_missing_object(caplog):
...
+                temporary_name="db-temp-test",
+                members=None,
+                entitlements=None,
+                external_id="eid1",
+                roles=None,
+            )
+        ]
+    )
+    verify_applied_scope_acls = mocker.patch(
+        "databricks.labs.ucx.workspace_access.verification.VerificationManager.verify_applied_scope_acls"
```
nfx, line 104: Don't monkey patch any "databricks.*" code. This test is not easy to change and would be missed by static code analysis, when we change internals of the class.

inject mocks of dependencies through constructor instead and verify that dependencies are called.

aminmovahed-db: I have removed all patches and applied MagicMock and dependency injection wherever possible.

nfx (approved):
Lgtm

PR #1033, opened 2024-03-08
Added task-level retries to improve stability of workflows, especially under test

nfx:
This PR adds `max_retries` on a task level for most of the things in the runtime.py. A run is considered to be unsuccessful if it completes with the FAILED result_state or INTERNAL_ERROR life_cycle_state.

See https://docs.databricks.com/api/workspace/jobs/create#tasks-max_retries


dmoore247 (commented):
Fun fact, two-thirds, 19 out of 28 @task  set max_retries = 1 and the other ~11 take the default.
The impact could be a failure in the lengthier sections such as crawl permissions or workspace objects could take 2x longer before finally failing. The benefit is that on the re-try, the task might succeed.

dmoore247 (approved):

PR #1050, opened 2024-03-13
Improved table migrations logic

nkvuong:
## Changes
- Refactor table migration unit tests to load table mappings from json instead of inline structs. This requires pulling `assessment/init.py` one level up in order to reuse fixtures loading logic
- Add `escape_sql_identifier` where missing
- This prepares for ACLs migration, which relies on similar logic

### Tests

- [x] manually tested


src/databricks/labs/ucx/hive_metastore/mapping.py
```diff
@@ -27,6 +28,18 @@ class Rule:
     src_table: str
     dst_table: str
 
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]):
```
nfx, line 32: Why do we need it?

src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -169,7 +170,7 @@ def _get_revert_count(self, schema: str | None = None, table: str | None = None)
         return migration_list
 
     def is_upgraded(self, schema: str, table: str) -> bool:
-        result = self._backend.fetch(f"SHOW TBLPROPERTIES `{schema}`.`{table}`")
+        result = self._backend.fetch(f"SHOW TBLPROPERTIES {escape_sql_identifier(schema+'.'+table)}")
```
nfx, line 173: You need to escape two identifiers

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -71,17 +85,17 @@ def kind(self) -> str:
         return "VIEW" if self.view_text is not None else "TABLE"
 
     def sql_alter_to(self, target_table_key):
-        return f"ALTER {self.kind} {self.key} SET TBLPROPERTIES ('upgraded_to' = '{target_table_key}');"
+        return f"ALTER {self.kind} {escape_sql_identifier(self.key)} SET TBLPROPERTIES ('upgraded_to' = '{target_table_key}');"
```
nfx, line 88: 'Cat'.'db'.'tbl' is different from 'cat.db.tbl'

We have to get a special key escape

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -260,7 +274,9 @@ def _describe(self, catalog: str, database: str, table: str) -> Table | None:
                 upgraded_to=self._parse_table_props(describe.get("Table Properties", "").lower()).get(
                     "upgraded_to", None
                 ),
-                storage_properties=self._parse_table_props(describe.get("Storage Properties", "").lower()),  # type: ignore[arg-type]
+                storage_properties=self._parse_table_props(
+                    describe.get("Storage Properties", "").lower()
```
nfx, line 278: The value can be None and this code will crash

tests/unit/__init__.py
```diff
@@ -13,3 +27,96 @@
...
+    fixture = _load_fixture(f"assessment/pipelines/{pipeline_id}.json")
+    return GetPipelineResponse.from_dict(fixture)
+
+
+def _secret_not_found(secret_scope, _):
+    msg = f"Secret Scope {secret_scope} does not exist!"
+    raise NotFound(msg)
+
+
+def workspace_client_mock(
```
nfx, line 93: we need to PR it to blueprint or pytester

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/table_migrate.py
nkvuong: `escape_sql_identifier` automatically handles identifiers with multiple parts

src/databricks/labs/ucx/hive_metastore/mapping.py
nkvuong: `table_mapping_mock` uses `cls.from_dict(_)` to load a dataclass from a json

src/databricks/labs/ucx/hive_metastore/tables.py
nkvuong: `self.key` is `catalog.database.name`, and `escape_sql_identifier` will turn escape the relevant part with backticks if needed - otherwise the `ALTER` statement will error out

our database is case insensitive, so we always lower case them

tests/unit/__init__.py
nkvuong: 👍  this looks to fit nicely into pytester - I'll raise a PR for it

src/databricks/labs/ucx/hive_metastore/tables.py
nkvuong: `get("Storage Properties", "")` would return empty string if the key does not exist, so we should be safe

src/databricks/labs/ucx/hive_metastore/mapping.py
```diff
@@ -46,12 +59,20 @@ def as_uc_table_key(self):
     def as_hms_table_key(self):
         return f"hive_metastore.{self.src_schema}.{self.src_table}"
 
+    def sql_alter_to_acl_upgraded(self) -> str:
```
FastLee, line 62: Move it to the dataclass tabletomigrate

nkvuong: took it out, will introduce in the PR for table acl migration

nfx (approved):
lgtm

Issue #907, opened 2024-02-06
[FEATURE]: Create UC Storage Credential, Schema, and Table Grants based on Azure SPN access


nkvuong:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Currently, SPNs are used to govern access to Lakehouse tables. We need to recreate these ACLs in UC.

If the cluster, cluster policy, sql warehouse, or instance pool has Azure Service Principal configuration in it, grant relevant permissions of the same cluster to the storage credentials

Related issues:
- https://github.com/databrickslabs/ucx/issues/887
- https://github.com/databrickslabs/ucx/issues/94
- https://github.com/databrickslabs/ucx/issues/340

### Proposed Solution

- select clusters, policies, sql warehouses and instance pools that triggered the Azure service principal failure
- Map users/groups to those objects they can access directly or through an existing cluster.
- Create table ACLS based on the SPN access.

Issue #887, opened 2024-02-05
[FEATURE]: Create UC Storage Credential, Schema, and Table Grants based on AWS IAM role access


FastLee:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Currently Instance Profiles and AWS IAM roles and policies are used to govern access to Lakehouse tables.
We need to recreate these ACLs in UC

Related issues:
- https://github.com/databrickslabs/ucx/issues/94
- https://github.com/databrickslabs/ucx/issues/340
- https://github.com/databrickslabs/ucx/issues/891
- https://github.com/databrickslabs/ucx/issues/862
- https://github.com/databrickslabs/ucx/issues/337

### Proposed Solution

- Map the Instance Profiles to the tables they can access
- Map users/groups to instance profile they can access directly or through an existing cluster.
- Create table ACLS baed on the role access.

PR #1054, opened 2024-03-14
Added Legacy Table ACL grants migration

### LINKED ISSUE #340: [FEATURE]: Create UC Schema and Table Grants based on Legacy Table ACLs
## Background
Customers who wish to upgrade to UC are left in a position where their legacy table ACLs are incompatible with Unity Catalog. Customers desire that these legacy ACLs be migrated to UC.

Upstream dependencies:
- https://github.com/databrickslabs/ucx/issues/670

## Privilege Model Differences
By default, Table ACLs are permissive by design, whereas UC is not permissive by design. 

Specifically, if a grant does not exist on a table in UC, you will not have access to it. If a grant does that matches your user, then you will. 

Table ACLs work opposite to this. By default, you have access to all objects unless a grant exists on that table that does not give you access.

Table ACLs also support DENY, which does not exist in UC.

## Privilege Model Map Hive To UC
|Hive Metastore Privilege|Intended Functional Action|Object types|UC Metastore Privilege|
|--|--|--|--|
|SELECT| gives read access to an object.|Catalog, schema, table, view|SELECT|
|CREATE| gives ability to create an object (for example, a table in a schema).|schema, table, view|CREATE|
|MODIFY| gives ability to add, delete, and modify data to or from an object.|schema, table, view|MODIFY|
|USAGE| does not give any abilities, but is an additional requirement to perform any action on a schema object|schema, table, view|USAGE|
|READ_METADATA| gives ability to view an object and its metadata.|schema, table, view|**BROWSE**|
|CREATE_NAMED_FUNCTION| gives ability to create a named UDF in an existing catalog or schema.|function|CREATE FUNCTION|
|MODIFY_CLASSPATH| gives ability to add files to the Spark class path.|-|does not translate|
|ALL PRIVILEGES| gives all privileges (is translated into all the above privileges).|schema, table|ALL PRIVILEGES|

**READ_METADATA** translates to the BROWSE privilege in UC, and can be granted on all objects - this function is in preview and should be enabled for any customer using this.
**CREATE_NAMED_FUNCTION** translates to CREATE FUNCTION in UC.

port the mapping to https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/hive_metastore/grants.py#L93-L111

# WIP

## Dealing with 2-level to 3-level namespace changes
If a user has USAGE access to a schema, then they also need access to the translated catalog in UC.
All USAGE on schemas should be translated to USAGE on the schema in UC, as well as USAGE on the containing catalog.

## Recommended Migration Approach
Grab all ACLs (GRANTs/DENY) on tables, views, and schemas.
Generate a distinct list of objects not in this list that are in HMS.

**For each object in HMS that does not have an ACL (GRANT and DENY) on it directly**
- If this object is contained within a schema in HMS that has a USAGE grant on it, and that USAGE grant is not applied to all users. **Translate this to mean that this table is hidden from view except for the groups or users in that ACL.**
 - - Provide a BROWSE ACL on this particular schema for any users and groups that have USAGE on this schema as they mean the same thing in this context
 - - Translate the grants otherwise accordingly to the map.
- If this object is contained within a schema that has a DENY on it **Only grant USAGE on the schema to members identified**
- If this object is contained within a schema that does not have a USAGE grant. **Translate this to mean that this table is accessible by all users**

**For each ACL on an object

nkvuong:
### Linked issues

Resolves #340
To be followed up with PRs for  #887 #907

### Functionality 

- [x] modified existing workflow: `migrate-tables`

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


nfx (approved):
lgtm

PR #1067, opened 2024-03-16
Added initial version of `databricks labs ucx migrate-local-code` command

nfx:
The `databricks labs ucx migrate-local-code` command has been added to facilitate migration of local code to a Databricks environment. This initial version of the command is highly experimental, with support for migrating Python and SQL files only. The `.gitignore` file has been updated to exclude output files and specific configuration files from being committed to the repository. This command aims to help users and administrators manage code migration and maintain consistency across workspaces, while also enhancing the compatibility of local code with the Unity Catalog, a part of Databricks' offerings for data and AI.


src/databricks/labs/ucx/code/queries.py
```diff
@@ -0,0 +1,59 @@
...
+                    severity=Severity.ERROR,
+                    tags=[DiagnosticTag.DEPRECATED],
+                )
+
+    @staticmethod
+    def _catalog(table):
+        catalog = table.catalog
+        if not catalog:
+            catalog = 'hive_metastore'
+        return catalog
```
ganeshgirase, line 50: We can make this method affirmative as,
```suggestion
        catalog = table.catalog
        if catalog:
            return catalog
        return "hive_metastore"
```

src/databricks/labs/ucx/code/queries.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def apply(self, query: str) -> str:
+        new_statements = []
+        for statement in sqlglot.parse(query):
+            if not statement:
+                continue
+            for old_table in statement.find_all(Table):
+                catalog = self._catalog(old_table)
+                if catalog != 'hive_metastore':
```
ganeshgirase, line 59: If we just looking after hive_metastore, then there is no need to keep private static method "_catalog", just a method for checking hive_metastore would be useful.

src/databricks/labs/ucx/code/queries.py
```diff
@@ -0,0 +1,59 @@
...
+        for statement in sqlglot.parse(query):
+            if not statement:
+                continue
+            for table in statement.find_all(Table):
+                catalog = self._catalog(table)
+                if catalog != 'hive_metastore':
+                    continue
+                if not self._index.is_upgraded(table.db, table.name):
+                    continue
+                dst = self._index.get(table.db, table.name)
```
ganeshgirase, line 34: is_upgraded and get method doing iteration two times, first to check whether it's upgraded and second time to get it.  Rather, it can be combined to just a single method by using get method only. If we receive other than None from get method, we proceed else skip loop.

ganeshgirase (commented):
Started with first review. I may be wrong from functional perspective, but just providing technical inputs.

- Ganesh Girase. (ganeshgirase@gmail.com)

PR #1068, opened 2024-03-18
Fix some integration test failures

qziyuan:
## Changes

1. When `test_migrate_external_table` are being executed by two different integration tests at the same time, the latter test will fail because the first test creates an UC external table occupying the table location, since a location can only be used by one UC table. This PR proposes a new test fixture that will create a random location for UC external table migration test, and delete it when test is done to avoid the location overlapping error.
2. Update `test_policy_crawler` to add a random suffix to the cluster policy name to avoid the test failure due to leftover/uncleaned cluster policy named `test_policy_check` and `test_policy_check2`. 

### Linked issues

Resolves #..

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/mixins/fixtures.py
+++ b/src/databricks/labs/ucx/mixins/fixtures.py
@@ -1161,3 +1161,15 @@ def make_feature_table(ws, make_random):
             logger.info(f"Can't remove feature table {e}")
 
     yield from factory("Feature table", create, remove)
+
+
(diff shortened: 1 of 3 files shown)


nfx (approved):
Lgtm

PR #1074, opened 2024-03-20
Addressed Integration Test Issue

FastLee:
In the latest update to the `ucx` framework, the `trigger` function in `tasks.py` has been modified to include a new argument, `install_folder`, in the `Installation` object. This object is now created within the `trigger` function and is passed to the `run_task` function. The `install_folder` is calculated by obtaining the parent directory of the `config_path` variable, converting it to a POSIX-style path, and removing the leading "/Workspace" prefix. This change ensures that the `run_task` function receives the correct installation folder for the `ucx` framework, enhancing the overall functionality and accuracy of the `ucx` framework.


nfx (approved):
lgtm

PR #1084, opened 2024-03-21
Skip installation if remote and local version is the same, provide prompt to override

### LINKED ISSUE #803: [FEATURE]: Allow only single UCX installation per workspace or dynamically determine it

### Problem statement

Multiple UCX installations per workspace cause plenty of problems
- two installations pointing to the same database will corrupt the state, potentially resulting in a data loss
- account admin won't easily be able to update workspace mappings everywhere required
- some commands, like `databricks labs ucx sync-workspace-info`, `databricks labs ucx sync-azure-service-principal-access`, and `databricks labs ucx sync-aws-role-access` are to be executed by IAM/account admins, that don't per-se require a full installation of UCX jobs and dashboards, as they need to push data to existing installation instead
- https://github.com/databrickslabs/ucx/issues/350
- https://github.com/databrickslabs/ucx/issues/1072

### Proposed Solution

- by default, install into `/.ucx` (workspace root), giving `prompt.confirm("do you want to install ucx just for a user?")` to install it in `/Users/$currentUser.me/.ucx` instead. default name of ucx database should use `WorkspaceInstaller._app` (username suffix) then
- when being installed on a user level, check other installations and verify that no conflicting databases exist
- introduce `InstallationManager.current()` to pick installation either from current user or workspace root

### Additional Context

_No response_

### LINKED ISSUE #1072: [FEATURE]: If UCX installed remotely on the workspace has different version than the one installed locally and we don't launch `databricks labs update ucx`, simply install UCX locally as the CLI

### Problem statement

At the moment, we install UCX both remotely (on a workspace) and on a workstation (in `~/.databricks/labs/ucx` folder on a dev machine). We don't always need to re-install the workspace installation if the version that is local equals to the version that is remote (`/Applications/ucx/version.json`)

### Proposed Solution

1. read the `/Applications/ucx/version.json` (remote config) using the `WheelsV2` class
2. compare with `ProductInfo.released_version()`
3. if `released_version` is the same as the remotely installed version, skip the installer
4. optionally, prompt.confirm user if they want to change the existing installation (default false)

### Additional Context

_No response_

pritishpai:
## Changes
Compares the remote version and local wheel version and prompts for re-install confirmation if they are the same. 

### Linked issues
Closes #1072
Closes #803

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [x] modified existing workflow: `...`
- [ ] added a new table

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/install.py
+++ b/src/databricks/labs/ucx/install.py
@@ -13,7 +13,12 @@ from databricks.labs.blueprint.installation import Installation, SerdeError
 from databricks.labs.blueprint.parallel import ManyError, Threads
 from databricks.labs.blueprint.tui import Prompts
 from databricks.labs.blueprint.upgrades import Upgrades
-from databricks.labs.blueprint.wheels import ProductInfo, WheelsV2, find_project_root
+from databricks.labs.blueprint.wheels import (
+    ProductInfo,
+    Version,
+    WheelsV2,
+    find_project_root,
+)
 from databricks.labs.lsql.backends import SqlBackend, StatementExecutionBackend
 from databricks.labs.lsql.deployment import SchemaDeployer
 from databricks.sdk import WorkspaceClient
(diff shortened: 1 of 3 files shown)


src/databricks/labs/ucx/install.py
```diff
@@ -144,6 +157,18 @@ def run(
...
+        try:
+            local_version = self._product_info.released_version()
+            remote_version = self._installation.load(Version).version
+            if extract_major_minor(remote_version) == extract_major_minor(local_version):
+                logger.info(f"UCX v{self._product_info.version()} is already installed on this workspace")
+                msg = "Do you want to update the existing installation?"
+                if not self._prompts.confirm(msg):
+                    raise RuntimeWarning("Remote and Local versions are same and no override is requested. Exiting...")
+        except NotFound as err:
+            logger.warning(f"Remote version not found: {err}")
```
dmoore247, line 170: Remote version of what? I'm going to see this in a sea of messages, please provide context.

dmoore247 (approved):
Just one comment.

src/databricks/labs/ucx/install.py
pritishpai: Remote version of UCX installation. Expanded on the warning and error messages.

nfx (approved):

PR #1095, opened 2024-03-24
Experimental support for scanning Delta Tables inside Mount Points

william-conti:
## Changes
This feature adds a new crawler that scans inside mount points for Delta Tables.

### Functionality 

Required
- [x] Add --include-mounts to crawl only for specific mounts.  

Nice to have
- [x] Add --include-parquet-files to include parquet files to the result of the script.  
- [x] Add --filter-paths to filter for specific paths in a mount point.  
- [ ] Add --max-depth to filter only at a specific amount of sub-folders.


src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +251,80 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+        all_mounts = self._mc.snapshot()
+        all_tables = []
+        for mount in all_mounts:
+            if self._include_mounts and mount.name not in self._include_mounts:
+                continue
+            table_paths = self._find_delta_log_folders(mount.name)
+            for path in table_paths:
+                table = Table(
+                    catalog="hive_metastore",
+                    database="",
```
nfx, line 295: Database can't be empty. Infer it from path and default to "default"

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +251,80 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+        entries: list[FileInfo] = self._dbutils.fs.ls(root_dir)
+        if len(entries) == 0:
+            return delta_log_folders
+        for entry in entries:
+            if self._is_irrelevant(entry.name):
+                continue
+
+            # Since isDir() is not present in SDK dbutils, we assume that a folder is anything
+            # that doesn't contains a "." in it's name
+            if "." not in entry.name and entry.name == "_delta_log":
```
nfx, line 317: Second check isn't enough?

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +251,80 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+                logger.debug(f"Found {entry.path}")
+                delta_log_folders.append(entry.path.replace("/_delta_log", ""))
+            else:
+                self._find_delta_log_folders(entry.path, delta_log_folders)
+
+        return delta_log_folders
+
+    def _is_irrelevant(self, entry: str) -> bool:
+        patterns = [r'_SUCCESS', r'_committed_', r'_started_', r'=', r'\.']
```
nfx, line 326: This can be a set[str] field

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +251,80 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+        for pattern in patterns:
+            if bool(re.search(pattern, entry)):
```
nfx, line 328:
```suggestion
            if pattern in entry:
```

src/databricks/labs/ucx/hive_metastore/locations.py
nfx: And success and = are relevant for parquet tables

nfx (changes requested):

src/databricks/labs/ucx/config.py
```diff
@@ -47,6 +47,10 @@ class WorkspaceConfig:  # pylint: disable=too-many-instance-attributes
     # Whether the assessment should capture a specific list of databases, if not specified, it will list all databases.
     include_databases: list[str] | None = None
 
+    # Whether the tables in mounts crawler should crawl a specific list of mounts.
+    # If not specified, it will list all moubts.
+    include_mounts: list[str] | None = None
```
nfx, line 52: By the way, alternatively, we can ask for it during the installer

src/databricks/labs/ucx/hive_metastore/locations.py
william-conti: I don't really want to do the mapping of paths -> databases/tables there. As this will be handled later in the process. 
What if instead just using the mount name as a database ?

william-conti: Actually, it would even better to default the database of this to 'mounts' or something that would be easily searcheable for downstream applications that must use this table.

nfx: and what if there are duplicate table names?...

tests/unit/hive_metastore/test_locations.py
```diff
@@ -226,3 +229,21 @@ def test_match_table_external_locations():
     assert len(matching_locations) == 1
     assert ExternalLocation("gcs://test_location2/a/b/", 1) in missing_locations
     assert ExternalLocation("abfss://cont1@storagetest1/a/", 2) in missing_locations
+
+
+def test_mount_listing_one_table():
+    client = create_autospec(WorkspaceClient)
+    client.dbutils.fs.ls.return_value = [
+        FileInfo("/mnt/lmao/_delta_log", "_delta_log", "", ''),
```
nfx, line 237: use different name in public code ;)

src/databricks/labs/ucx/hive_metastore/locations.py
william-conti: yeah that can totally happen ... the strategy of picking the parent folder as the table name is irrelevant then. 
Maybe let's simply use the table path in mount as the table name instead ? ...

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,107 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+        Returns the parent directory of a given path.
+        Example: _get_table_name('/mnt/test/123/dxs/aaa.parquet') -> dxs
+        """
+        return re.search("([^/]+)/?$", path, re.IGNORECASE).group(1)
+
+    def _path_is_delta(self, delta_log_folders, path):
+        return delta_log_folders.get(path, {}) and delta_log_folders.get(path, {}).is_delta
+
+    def _is_parquet(self, entry: str) -> bool:
+        parquet_patterns = {'.parquet'}
```
nfx, line 351:
```suggestion
        parquet_patterns = {'.parquet', '='}
```
this way you'll detect partitions OR non-partitioned tables.

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,107 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+    def _find_delta_log_folders(self, root_dir, delta_log_folders=None) -> dict:
+        if delta_log_folders is None:
+            delta_log_folders = {}
+
+        entries = self._dbutils.fs.ls(root_dir)
+        for entry in entries:
+            if self._is_irrelevant(entry.name):
+                continue
+
+            path = self._get_absolute_directory(entry.path)
```
nfx, line 319:
```suggestion
            path = os.path.dirname(entry.path)
```
standard lib has it all

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,107 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+
+            if self._is_parquet(entry.path):
+                if self._path_is_delta(delta_log_folders, path):
+                    continue
+                delta_log_folders[path] = TableInMount(is_delta=False, is_parquet=True, is_partitioned=False)
+            elif entry.name == "_delta_log":
```
nfx, line 325: check for delta before checking for parquet.

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,107 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+
+        entries = self._dbutils.fs.ls(root_dir)
+        for entry in entries:
+            if self._is_irrelevant(entry.name):
+                continue
+
+            path = self._get_absolute_directory(entry.path)
+
+            if self._is_parquet(entry.path):
+                if self._path_is_delta(delta_log_folders, path):
```
nfx, line 322: this is a bug - check for delta before you check for parquet.

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,107 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+        Example: _get_absolute_directory('/mnt/test/123/dxs/aaa.parquet') -> /mnt/test/123/dxs/
+        """
+        return re.search(r'^(.*/)[^/]+/?', path).group(1)
+
+    def _get_table_name(self, path):
+        """
+        Returns the parent directory of a given path.
+        Example: _get_table_name('/mnt/test/123/dxs/aaa.parquet') -> dxs
+        """
+        return re.search("([^/]+)/?$", path, re.IGNORECASE).group(1)
```
nfx, line 345:
```suggestion
        return os.path.basename(os.path.dirname(path))
```

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,107 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+            table_paths = self._find_delta_log_folders(mount.name)
+            for path, entry in table_paths.items():
+                table = Table(
+                    catalog="hive_metastore",
+                    database="tables_in_mounts",
+                    name=self._get_table_name(path),
+                    object_type="EXTERNAL",
+                    table_format="DELTA" if entry.is_delta else "PARQUET",
+                    location=path,
+                )
```
nfx, line 306:
```suggestion
                guess_table = os.path.basename(path)
                guess_database = os.path.dirname(path)
                table = Table(
                    catalog="hive_metastore",
                    database=guess_database,
                    name=guess_table,
                    object_type="EXTERNAL",
                    table_format="DELTA" if entry.is_delta else "PARQUET",
                    location=path,
                )
```

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,107 @@ def _try_fetch(self) -> Iterable[Mount]:
             f"SELECT * FROM {escape_sql_identifier(self._schema)}.{escape_sql_identifier(self._table)}"
         ):
             yield Mount(*row)
+
+
+@dataclass
+class TableInMount:
+    is_delta: bool
+    is_parquet: bool
```
nfx, line 258:
```suggestion
    format: str
```
because we'll have `JSON`, `CSV`, ...

nfx (commented):
few bugs

src/databricks/labs/ucx/hive_metastore/locations.py
william-conti: Partition support should also work for delta tables, not parquet files only. And I would like to avoid scanning inside partitions to reduce the listing calls

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,109 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+                    object_type="EXTERNAL",
+                    table_format=entry.format,
+                    location=path,
+                    storage_properties="PARTITIONED" if entry.is_partitioned else None,
+                )
+                all_tables.append(table)
+        return all_tables
+
+    def _find_delta_log_folders(self, root_dir, delta_log_folders=None) -> dict:
+        if delta_log_folders is None:
```
nfx, line 318: add more logs as well :)

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,109 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+    def _try_load(self) -> Iterable[Table]:
+        """Tries to load table information from the database or throws TABLE_OR_VIEW_NOT_FOUND error"""
+        for row in self._fetch(f"SELECT * FROM {escape_sql_identifier(self.full_name)}"):
+            yield Table(*row)
+
+    def _crawl(self):
+        all_mounts = self._mc.snapshot()
+        all_tables = []
+        for mount in all_mounts:
+            if self._include_mounts and mount.name not in self._include_mounts:
```
ganeshgirase, line 300: First check in "if" condition is redundant, as self._include_mounts is already defined and checking key in set always result to amortized complexity of O(1).  So, just checking key absence is more than enough. Also, if possible to check there isn 't duplicate values in snapshots, then we can modify to get set of mounts instead of list of mounts.

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,109 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+                    name=guess_table,
+                    object_type="EXTERNAL",
+                    table_format=entry.format,
+                    location=path,
+                    storage_properties="PARTITIONED" if entry.is_partitioned else None,
+                )
+                all_tables.append(table)
+        return all_tables
+
+    def _find_delta_log_folders(self, root_dir, delta_log_folders=None) -> dict:
```
ganeshgirase, line 317: Type hints requires. if delta_log_folders is dict, it can be initialized with default value as empty dict.

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,109 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+        if delta_log_folders is None:
+            delta_log_folders = {}
+
+        entries = self._dbutils.fs.ls(root_dir)
```
ganeshgirase, line 321: From below code, it would be wise to rename it as paths. and rename "path" variable from line 326 to directory_name to precise.

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,109 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+                self._find_delta_log_folders(entry.path, delta_log_folders)
+
+        return delta_log_folders
+
+    def _path_is_delta(self, delta_log_folders, path: str) -> bool:
+        return delta_log_folders.get(path) and delta_log_folders.get(path).format == "DELTA"
+
+    def _is_partitioned(self, entry: str) -> bool:
+        partitioned_pattern = {'='}
+        return any(pattern in entry for pattern in partitioned_pattern)
```
ganeshgirase, line 351: Isn't just below line would suffice?

```suggestion
    def _is_partitioned(self, entry: str) -> bool:
        return '=' in entry
```

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,109 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+                if delta_log_folders.get(path) and delta_log_folders.get(path).is_partitioned:
+                    delta_log_folders[path] = TableInMount(format="DELTA", is_partitioned=True)
+                else:
+                    delta_log_folders[path] = TableInMount(format="DELTA", is_partitioned=False)
+            elif self._is_partitioned(entry.path):
+                delta_log_folders[path] = TableInMount(format="PARQUET", is_partitioned=True)
+            elif self._is_parquet(entry.path):
+                delta_log_folders[path] = TableInMount(format="PARQUET", is_partitioned=False)
+            else:
+                self._find_delta_log_folders(entry.path, delta_log_folders)
```
ganeshgirase, line 342: This seems dangerous logic, as you're forcing to use heap structure in recursion, that can slow down the process, if you've huge list of paths. Line no 328 will also contribute more to this slowdown. Rather, use a logic of processing single path, process it till leaf path, make TableInMount object there and return from stack to accumulate.

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -248,3 +250,109 @@ def _try_fetch(self) -> Iterable[Mount]:
...
+
+        return delta_log_folders
+
+    def _path_is_delta(self, delta_log_folders, path: str) -> bool:
+        return delta_log_folders.get(path) and delta_log_folders.get(path).format == "DELTA"
+
+    def _is_partitioned(self, entry: str) -> bool:
```
ganeshgirase, line 349: In caller method, saying that "entry" is an object, but here entry is string. That's misleading.

tests/unit/hive_metastore/test_locations.py
```diff
@@ -226,3 +229,213 @@ def test_match_table_external_locations():
...
+            'hive_metastore.test.tables': [],
+            'test.mounts': MOUNT_STORAGE[("/mnt/test_mount", "")],
+        }
+    )
+    mounts = Mounts(backend, client, "test")
+    results = TablesInMounts(backend, client, "test", mounts).snapshot()
+    assert results == [
+        Table(
+            "hive_metastore",
+            "mounted_/mnt/test_mount",
```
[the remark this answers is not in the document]
william-conti: IMO the database should be the mount name, this way it will be easier to find out where this table comes from

src/databricks/labs/ucx/hive_metastore/locations.py
william-conti: self._include_mounts is None by default, hence  mount.name not in self._include_mounts can fail.

There cannot be duplicate mount values as Databricks workspaces cannot have mounts with the same name

src/databricks/labs/ucx/runtime.py
```diff
@@ -560,6 +561,17 @@ def validate_groups_permissions_experimental(
     return validate_groups_permissions(cfg, ws, sql_backend, _install)
 
 
+@task("migrate-tables-in-mounts")
+def scan_tables_in_mounts(cfg: WorkspaceConfig, ws: WorkspaceClient, sql_backend: SqlBackend, _install: Installation):
+    """This workflow scans for Delta tables inside all mount points captured during the assessment.
```
nfx, line 566: Mark it as experimental, so that users wouldn't randomly trigger it

tests/unit/hive_metastore/test_locations.py
```diff
@@ -226,3 +230,308 @@ def test_match_table_external_locations():
...
+        rows={
+            'hive_metastore.test.tables': [],
+            'test.mounts': MOUNT_STORAGE[("/mnt/test_mount", "adls://bucket/")],
+        }
+    )
+    mounts = Mounts(backend, client, "test")
+    results = TablesInMounts(backend, client, "test", mounts).snapshot()
+    assert results == [
+        Table("hive_metastore", "mounted_/mnt/test_mount", "table1", "EXTERNAL", "DELTA", "adls://bucket/table1"),
+        Table("hive_metastore", "mounted_/mnt/test_mount", "table2", "EXTERNAL", "PARQUET", "adls://bucket/table2"),
```
nfx, line 264:
```suggestion
        Table("hive_metastore", "mounted_test_mount", "table2", "EXTERNAL", "PARQUET", "adls://bucket/table2"),
```
No slashes in schema name!

nfx (changes requested):
Fix db name detection

nfx (approved):
okay

PR #1134, opened 2024-03-26
Add AVRO to supported format to be upgraded by SYNC

qziyuan:


nfx (approved):

qziyuan:
@FastLee I think our document for [SYNC](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-aux-sync.html#common-status-codes-returned-by-sync) is missing AVRO. I did a test to confirm AVRO is supported by SYNC and filed a jira to fix the document.
But just in case I need your second eye on this.

nkvuong:
@qziyuan the SYNC command supports any file format that is supported by UC, namely (Delta, Csv, Json, Xml, Avro, Parquet, ORC, Text). Avro has always been supported, while xml is only added in DBR 14.3+

nkvuong (commented):
there is a unit test checking for Avro that is now failing

qziyuan:
> @qziyuan the SYNC command supports any file format that is supported by UC, namely (Delta, Csv, Json, Xml, Avro, Parquet, ORC, Text). Avro has always been supported, while xml is only added in DBR 14.3+

@nkvuong great, thanks for the confirmation. I don't think we need to migrate XML with SYNC, I believe there is no native Hive serde for XML. So customer must use some 3rd party serde if they ever have HMS table in XML format. In this case it's better to CTAS to delta for the migration.

nkvuong (approved):
lgtm

PR #1135, opened 2024-03-27
Add ACL migration to `migrate-tables` workflow

nkvuong:
## Changes
- Add ACL migration strategy as a user configuration
- Pass this strategy to `migrate-tables` workflow

### Functionality 

- [x] modified existing workflow: `migrate-tables`

### Tests

- [x] manually tested

### DIFF
--- a/src/databricks/labs/ucx/config.py
+++ b/src/databricks/labs/ucx/config.py
@@ -46,7 +46,6 @@ class WorkspaceConfig:  # pylint: disable=too-many-instance-attributes
     # Flag to see if terraform has been used for deploying certain entities
     is_terraform_used: bool = False
 
-    # Flag to see if terraform has been used for deploying certain entities
     grant_migration_strategy: AclMigrationWhat | None = None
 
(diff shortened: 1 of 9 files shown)


src/databricks/labs/ucx/config.py
```diff
@@ -44,6 +46,8 @@ class WorkspaceConfig:  # pylint: disable=too-many-instance-attributes
     # Flag to see if terraform has been used for deploying certain entities
     is_terraform_used: bool = False
 
+    grant_migration_strategy: AclMigrationWhat | None = None
```
nfx, line 49: i don't want to add any unstable enums to WorkspaceConfig, because it'll escalate when we change the enum and break customer configuration loading from previous formats. 

please rewrite the code NOT to use enums in config? you assume that ACLs are `either ... or ...`, but they have to be `... and ...`

image

nfx (changes requested):

nfx (approved):
lgtm

PR #1123, opened 2024-03-26
Add assessment of interactive cluster usage compared to UC compute limitations

dmoore247:
## Changes
Add new 'interactive' assessment report that providing insights into commands that have run on interactive compute and findings of potential issues in converting the clusters to Shared Unity Catalog data security mode.
image
image
image

### Linked issues
Resolves #810 
Resolves #498 

### Functionality 

- [x] added relevant user documentation
- [ ] added new CLI command
- [ ] modified existing command: `databricks labs ucx ...`
- [ ] added a new workflow
- [ ] modified existing workflow: `...`
- [ ] added a new table
- [ ] modified existing table: `...`
- [x] added new dashboard

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [x] verified on staging environment (screenshot attached)

PR #1139, opened 2024-03-27
Deploy static views needed by #1123 interactive dashboard

dmoore247:
## Changes
Deploy views `misc_patterns_vw` and `code_patterns_vw` that should have been deployed with #1123 and fixes the `interactive` dashboard queries.

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [x] added a new table

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] 
image


pritishpai:
This one looks fairly straightforward. However, I see that the other view queries and the actual views in the inventory database do not have a suffix "_vw". Just for the sake of consistency, we could remove it (unless there was a reason to specifically add them to these 2 views).

pritishpai (commented):
LGTM! This one looks fairly straightforward. However, I see that the other view queries and the actual views in the inventory database do not have a suffix "_vw". Just for the sake of consistency, we could remove it (unless there was a reason to specifically add them to these 2 views).

pritishpai (approved):

src/databricks/labs/ucx/install.py
```diff
@@ -89,6 +89,8 @@ def deploy_schema(sql_backend: SqlBackend, inventory_schema: str):
     deployer.deploy_view("objects", "queries/views/objects.sql")
     deployer.deploy_view("grant_detail", "queries/views/grant_detail.sql")
     deployer.deploy_view("table_estimates", "queries/views/table_estimates.sql")
+    deployer.deploy_view("misc_patterns", "queries/views/misc_patterns.sql")
```
nfx, line 92: this cannot be done during the main installation phase, which has to run on workspaces without UC enabled. these views rely on `system` schema, that does not exist yet. add a separate temporary CLI command to deploy these views and document it.

nfx (changes requested):

src/databricks/labs/ucx/install.py
dmoore247: @nfx Please take a look at the code in these two views. 
They are static SELECT FROM VALUES views to encode easy to change patterns etc.
The `system.access.audit` table is evaluated at open time of the dashboard. The dashboard header advises users of the requirements. Future releases can add commands to ensure system tables are turned on etc.

nfx (approved):

Issue #1153, opened 2024-03-28
[BUG]: The views captured by the assessment have NULL in table_format column instead of UNKNOWN


william-conti:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

After running the assessment, the xxx column in ucx.tables is NULL, while it should actually be UNKNOWN, this causes downstream applications to fail because this column should not be NULL accodring to the Table object 

![Capture d’écran 2024-03-28 à 14 26 31](https://github.com/databrickslabs/ucx/assets/107417584/95f07198-d8c7-46af-a76b-3d995b7cc998)

### Expected Behavior

Column table_format should have UNKNOWN instead of NULL

### Steps To Reproduce

Run the assessment
SELECT * FROM ucx.tables

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

PR #1156, opened 2024-03-28
Fixed `NULL` values in `ucx.views.table_format` to have `UNKNOWN` value instead

william-conti:
## Changes
Added a fix for Views that are having NULL in table_format column instead of UNKNOWN 

### Linked issues
Fixes https://github.com/databrickslabs/ucx/issues/1153

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/tables.scala
+++ b/src/databricks/labs/ucx/hive_metastore/tables.scala
@@ -64,7 +64,7 @@ def metadataForAllTables(databases: Seq[String], queue: ConcurrentLinkedQueue[Ta
           }
           val isPartitioned = if (partitionColumnNames != null && !partitionColumnNames.isEmpty) true else false
 
-          Some(TableDetails("hive_metastore", databaseName, tableName, table.tableType.name, table.provider.orNull,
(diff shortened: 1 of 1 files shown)


src/databricks/labs/ucx/hive_metastore/tables.scala
```diff
@@ -64,7 +64,7 @@ def metadataForAllTables(databases: Seq[String], queue: ConcurrentLinkedQueue[Ta
           }
           val isPartitioned = if (partitionColumnNames != null && !partitionColumnNames.isEmpty) true else false
 
-          Some(TableDetails("hive_metastore", databaseName, tableName, table.tableType.name, table.provider.orNull,
+          Some(TableDetails("hive_metastore", databaseName, tableName, table.tableType.name, table.provider.getOrElse("UNKNOWN"),
```
nfx, line 67: Also make a corresponding change to TablesCrawler, which is used in integration tests

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/tables.scala
william-conti: Actually the logic implemented there comes from TablesCrawler. This part were missing

PR #1161, opened 2024-03-28
Use experimental Permissions Migration API also for Legacy Table ACLs

nfx:
This PR removes verification step from the experimental group migration job, as platform already replaces permissions from workspace group to account group and old validation method is not required anymore.

### DIFF
--- a/src/databricks/labs/ucx/workspace_access/groups.py
+++ b/src/databricks/labs/ucx/workspace_access/groups.py
@@ -86,34 +86,50 @@ class MigrationState:
     def __len__(self):
         return len(self._name_to_group)
 
-    def apply_group_permissions_experimental(self, ws: WorkspaceClient) -> bool:
+    def apply_to_renamed_groups(self, ws: WorkspaceClient) -> bool:
+        """(Production) Apply migration state to groups that have been renamed in the workspace."""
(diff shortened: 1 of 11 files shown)


src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -86,40 +86,50 @@ def is_in_scope(self, name: str) -> bool:
...
-    def apply_group_permissions_experimental(self, ws: WorkspaceClient) -> bool:
+    def apply_to_renamed_groups(self, ws: WorkspaceClient) -> bool:
+        """(Production) Apply migration state to groups that have been renamed in the workspace."""
+        return self._apply_to_groups(ws, renamed=True)
+
+    def apply_to_groups_with_different_names(self, ws: WorkspaceClient) -> bool:
+        """(Integration Testing) Apply to groups that have different names in the workspace and account."""
+        return self._apply_to_groups(ws, renamed=False)
+
+    def _apply_to_groups(self, ws: WorkspaceClient, *, renamed: bool = False) -> bool:
```
ganeshgirase, line 97: curios to know why we have "*" as second parameter here, maybe we can remove if it's not needed.

PR #1199, opened 2024-04-01
[WIP] - Revamp catalog schema file

ganeshgirase:


nfx:
@ganeshgirase what features does it add? why do we need it? It doesn't seem to improve readability.

ganeshgirase:
> @ganeshgirase what features does it add? why do we need it? It doesn't seem to improve readability.

@nfx It's not feature. It is rewrite of the code in catalog_schema.py. It will only improve performance and may be readability as it's not performing complex logic like the existing code. In current code, we are maintaining two duplicate copies of catalog and schemas. That logic is re-written and just used what's bare minimum needed.

nfx:
@ganeshgirase please don't change the places that you cannot fully integration test - closing this PR. Pick up the smallest one of https://github.com/databrickslabs/ucx/issues?q=is%3Aissue+is%3Aopen+label%3AAST and open PRs.

PR #1209, opened 2024-04-01
Use service factory to resolve object dependencies

nfx:
This PR reduces the code duplication for object construction

| decorator | autocompletion in PyCharm | singleton behavior | overrides easy | pylint/mypy compatible |
| --- | --- | --- | --- | --- |
| `@functools.cached_property` | yes | yes | yes | yes |
| `@property` | yes | no | no | yes |
| `@functools.cache` + `@property` | ? | yes | no | yes |
| `@singleton` | no | yes | yes | an extension may be necessary| 
| `@singleton` extending `@property` | _surprisingly,_ no | yes | yes | an extension may be necessary| 
| `@singleton` + `@property` | no | yes | ? | ? |
| `@property` + `@singleton` | no | yes | ? | ? |

Rationale:
- Factories are only to be used to create single-instance objects.
- preferably have them created only once, when necessary. 
- more Spring Boot flavors are possible, but they may prevent code navination
- `__get__` protocol allows for storage in a target instance property, making it possible to run unit tests in parallel


src/databricks/labs/ucx/factories.py
```diff
@@ -0,0 +1,431 @@
...
+    def __str__(self):
+        return f"@bean: {self._qualname}"
+
+
+class GlobalContext:
+    # TODO: make flags only available in CLI contexts
+    flags: dict[str, str]
+
+    def __init__(self):
+        self.registry = {}
```
jasongrout-db, line 81: Perhaps this should start with an underscore (`_registry`) to indicate it is a private-ish kind of thing?

jasongrout-db (commented):
When you looked at functools, did you see it has a [functools.cached_property decorator](https://docs.python.org/3/library/functools.html#functools.cached_property)?

This also reminds me of pydantic (or traitlets) default values, which are factory functions that are called to construct the default value of a property. I don't know if those default values are lazily called, though.

nfx:
@jasongrout-db Let's try that cached property tomorrow, but need to verify if it can be overwritten for testing easily

nfx:
@jasongrout-db and `@cached_property` it'll be! wouldn't know about it without your hint.

src/databricks/labs/ucx/factories.py
nfx: luckily, we don't need it with `@cached_property` anymore :)

src/databricks/labs/ucx/factories.py
```diff
@@ -0,0 +1,412 @@
...
+from databricks.labs.ucx.hive_metastore.udfs import UdfsCrawler
+from databricks.labs.ucx.installer.workflows import WorkflowsInstallation
+from databricks.labs.ucx.source_code.files import Files
+from databricks.labs.ucx.source_code.languages import Languages
+from databricks.labs.ucx.workspace_access.clusters import ClusterAccess
+from databricks.labs.ucx.workspace_access.groups import GroupManager
+from databricks.labs.ucx.workspace_access.manager import PermissionManager
+
+
+class GlobalContext:
```
nfx, line 53:
```suggestion
class GlobalContextFactory:
```

nfx:
tests are passing locally:
image

PR #1210, opened 2024-04-01
Support table migration to Unity Catalog in Python code

ericvergnaud:
## Changes
Enhance SparkSql linter/fixer to support migration of spark.sql function calls that receive a table name as parameter

### Linked issues
Resolves # 1082

### Functionality 

- [ ] added new CLI command
- [x] modified existing command: `databricks labs ucx migrate_local_code`
- [ ] added a new workflow
- [ ] added a new table

### Tests
- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests

missing integration tests

### DIFF
--- a/src/databricks/labs/ucx/source_code/pyspark.py
+++ b/src/databricks/labs/ucx/source_code/pyspark.py
@@ -59,7 +59,7 @@ class QueryMatcher(Matcher):
         else:
             yield Advisory(
                 code='table-migrate',
-                message="Table",
+                message=f"Can't migrate '{node}' because its table name argument is not a constant",
                 start_line=node.lineno,
                 start_col=node.col_offset,
                 end_line=node.end_lineno or 0,
--- a/src/databricks/labs/ucx/source_code/pyspark.py
+++ b/src/databricks/labs/ucx/source_code/pyspark.py
@@ -14,14 +14,13 @@ class Matcher(ABC):
     min_args: int
     max_args: int
     table_arg_index: int
+    table_arg_name: str | None = None
 
     def matches(self, node: ast.AST):
-        return (
-            isinstance(node, ast.Call)
-            and isinstance(node.func, ast.Attribute)
-            and self.min_args <= len(node.args) <= self.max_args
-            and isinstance(node.args[self.table_arg_index], ast.Constant)
-        )
+        if not (isinstance(node, ast.Call) and isinstance(node.func, ast.Attribute)):
+            return False
+        arg = self._get_table_arg(node)
+        return isinstance(arg, ast.Constant)
 
     @abstractmethod
     def lint(self, from_table: FromTable, index: MigrationIndex, node: ast.Call) -> Iterator[Advice]:
@@ -31,12 +30,19 @@ class Matcher(ABC):
     def apply(self, from_table: FromTable, index: MigrationIndex, node: ast.Call) -> None:
         raise NotImplementedError()
 
+    def _get_table_arg(self, node: ast.Call):
+        if len(node.args) > 0:
+            return node.args[self.table_arg_index] if self.min_args <= len(node.args) <= self.max_args else None
+        assert self.table_arg_name is not None
+        arg = next(kw for kw in node.keywords if kw.arg == self.table_arg_name)
+        return arg.value if arg is not None else None
+
 
 @dataclass
 class QueryMatcher(Matcher):
 
     def lint(self, from_table: FromTable, index: MigrationIndex, node: ast.Call) -> Iterator[Advice]:
-        table_arg = node.args[self.table_arg_index]
+        table_arg = self._get_table_arg(node)
         assert isinstance(table_arg, ast.Constant)
         for advice in from_table.lint(table_arg.value):
             yield advice.replace(
@@ -47,7 +53,7 @@ class QueryMatcher(Matcher):
             )
 
     def apply(self, from_table: FromTable, index: MigrationIndex, node: ast.Call) -> None:
-        table_arg = node.args[self.table_arg_index]
+        table_arg = self._get_table_arg(node)
         assert isinstance(table_arg, ast.Constant)
         new_query = from_table.apply(table_arg.value)
         table_arg.value = new_query
@@ -57,7 +63,7 @@ class QueryMatcher(Matcher):
 class TableNameMatcher(Matcher):
 
     def lint(self, from_table: FromTable, index: MigrationIndex, node: ast.Call) -> Iterator[Advice]:
-        table_arg = node.args[self.table_arg_index]
+        table_arg = self._get_table_arg(node)
         assert isinstance(table_arg, ast.Constant)
         dst = self._find_dest(index, table_arg.value)
         if dst is not None:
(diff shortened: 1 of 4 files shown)


src/databricks/labs/ucx/source_code/queries.py
```diff
@@ -11,6 +11,9 @@ class FromTable(Linter, Fixer):
     def __init__(self, index: MigrationIndex):
         self._index = index
 
+    def index(self):
```
nfx, line 14: let's keep the dependencies to be injectable via constructor only, except a couple of predefined places, otherwise it may create unnecessary coupling

src/databricks/labs/ucx/source_code/pyspark.py
```diff
@@ -1,10 +1,155 @@
...
+    @abstractmethod
+    def apply(self, from_table: FromTable, node: ast.Call) -> None:
+        raise NotImplementedError()
+
+
+@dataclass
+class QueryMatcher(Matcher):
+
+    def lint(self, from_table: FromTable, node: ast.Call) -> Iterator[Advice]:
+        table_arg = node.args[self.table_arg_index]
```
nfx, line 38: what if the argument is passed by name? do we want to defensively add an argument name to a matcher?

src/databricks/labs/ucx/source_code/pyspark.py
```diff
@@ -1,10 +1,155 @@
...
+    TableNameMatcher("createTable", 1, 1000, 0),
+    TableNameMatcher("createExternalTable", 1, 1000, 0),
+    TableNameMatcher("getTable", 1, 1, 0),
+    TableNameMatcher("isCached", 1, 1, 0),
+    TableNameMatcher("listColumns", 1, 2, 0),
+    TableNameMatcher("tableExists", 1, 2, 0),
+    TableNameMatcher("recoverPartitions", 1, 1, 0),
+    TableNameMatcher("refreshTable", 1, 1, 0),
+    TableNameMatcher("uncacheTable", 1, 1, 0),
+    # TODO listTables: raise warning ?
```
nfx, line 101: yep

src/databricks/labs/ucx/source_code/pyspark.py
```diff
@@ -16,44 +161,30 @@ def name(self) -> str:
     def lint(self, code: str) -> Iterable[Advice]:
         tree = ast.parse(code)
         for node in ast.walk(tree):
-            if not isinstance(node, ast.Call):
-                continue
-            if not isinstance(node.func, ast.Attribute):
+            matcher = SparkSql._find_matcher(node)
```
nfx, line 164:
```suggestion
            matcher = self._find_matcher(node)
```

src/databricks/labs/ucx/source_code/pyspark.py
```diff
@@ -1,10 +1,155 @@
...
+    def apply(self, from_table: FromTable, node: ast.Call) -> None:
+        raise NotImplementedError()
+
+
+@dataclass
+class QueryMatcher(Matcher):
+
+    def lint(self, from_table: FromTable, node: ast.Call) -> Iterator[Advice]:
+        table_arg = node.args[self.table_arg_index]
+        assert isinstance(table_arg, ast.Constant)
```
nfx, line 39: and if it is `ast.Name` or `ast.JoinedStr`? probably yielding a warning is better other than skipping it. We can make things more configurable as well.

src/databricks/labs/ucx/source_code/pyspark.py
```diff
@@ -1,10 +1,155 @@
...
+    def _find_dest(from_table: FromTable, value: str):
+        parts = value.split(".")
+        return None if len(parts) != 2 else from_table.index().get(parts[0], parts[1])
+
+
+# see https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.SparkSession.html
+SPARK_SESSION_MATCHERS = [QueryMatcher("sql", 1, 1000, 0), TableNameMatcher("table", 1, 1, 0)]
+
+# see https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.Catalog.html
+SPARK_CATALOG_MATCHERS = [
```
nfx, line 90: please use class fields instead of global variables and pass in `SparkSql` (or `Index`) as the first argument to matchers. This way we don't have to leak the `index` property (as added in `queries.py`) and have more homogenous common interfaces and freedom to inject more via constructor.

Existing example - a factory of PermissionsManager

tests/unit/source_code/test_pyspark.py
```diff
@@ -46,6 +46,75 @@ def test_spark_sql_match(migration_index):
     ] == list(sqf.lint(old_code))
 
 
+def test_spark_table_match(migration_index):
+    # import SPARK_MATCHERS from
+    for method_name in ["cacheTable", "createTable", "createExternalTable", "getTable", "isCached",
```
nfx, line 51: this usage hints at `@pytest.mark.parametrize`

nfx (changes requested):
Nice abstractions! can we use class fields instead of global variables? this would prevent a couple of crawling problems down the line

src/databricks/labs/ucx/source_code/queries.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/pyspark.py
ericvergnaud: you mean by keyword ? like spark.sql.table(tablename="old.thing") ? At this point I believe the `tablename` arg is always positional in the Spark API, so it wouldn't affect the usage of table_arg_index but I'll double check and add a test.

src/databricks/labs/ucx/source_code/pyspark.py
ericvergnaud: This is mimicked from the existing code ;-) but let's improve that indeed

src/databricks/labs/ucx/source_code/pyspark.py
ericvergnaud: done

tests/unit/source_code/test_pyspark.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/pyspark.py
nfx: tracking in https://github.com/databrickslabs/ucx/issues/1220 to possibly be addressed later. Junior data scientists may abuse kvargs for no good reason and we need to defend against that.

ericvergnaud: Done

src/databricks/labs/ucx/source_code/pyspark.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/pyspark.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/pyspark.py
```diff
@@ -1,13 +1,220 @@
...
+                yield advice.replace(
+                    start_line=node.lineno,
+                    start_col=node.col_offset,
+                    end_line=node.end_lineno,
+                    end_col=node.end_col_offset,
+                )
+        else:
+            yield Advisory(
+                code='table-migrate',
+                message="Table",
```
nfx, line 62:
```suggestion
                message=f"Can't migrate '{node}' because its table name argument is not a constant",
```

nfx (approved):

PR #1216, opened 2024-04-02
Remove prompts dependency from WorkflowsDeployment

qziyuan:
## Changes

WorkflowsDeployment handles workflow deployment only and should not be dependent on prompts. Remove prompts from the class and only inject the prompts when uploading wheels to non writeable dbfs and need to use override cluster.

### Linked issues

Resolves #..

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


nfx (approved):
Lgtm

PR #1225, opened 2024-04-02
Add table in mount migration

william-conti:
## Changes
Added a functionality to migrate tables in mounts to Unity Catalog.

### Functionality 

- [x] Creates a table in UC based on the table mapping csv file.  
- [x] Fixes include_paths_in_mount not being present in workflows.py
- [x] Adds a functionality to set a default ownership on each table created.

### Tests

- [X] manually tested
- [X] added unit tests
- [x] added integration tests


src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -367,3 +368,85 @@ def _try_fetch(self) -> Iterable[MigrationStatus]:
...
+        ws: WorkspaceClient,
+        backend: SqlBackend,
+        table_mapping: TableMapping,
+    ):
+        self._backend = backend
+        self._ws = ws
+        self._tm = table_mapping
+        self._tim = tables_in_mounts
+
+    def create_tables_in_uc(self):
```
nfx, line 386: How would this work with source code migration?

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/table_migrate.py
william-conti: It won't, and this PR has never been designed to.
Everything here has been made separatly because the TableMigrator class highly assumes that a table can be identified by a database.table, which is wrong in the case of a table in a mount point. A table in a mount point is identified by a path, it's used like this in customer codebases (same as external tables tho), and is mapped from path to catalog.database.table.

Because of this we either have to highly tweak the TableMigrator to include the special case of tables in a mount point, or treat this special case separatly.

Additionally, if tables in mounts needs to be compatible with the code migration,  we'll have to revamp the TableInMountCrawler a bit

src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -367,3 +368,85 @@ def _try_fetch(self) -> Iterable[MigrationStatus]:
     def _iter_schemas(self):
         for catalog in self._ws.catalogs.list():
             yield from self._ws.schemas.list(catalog_name=catalog.name)
+
+
+class TablesInMountsMigrator:
```
nfx, line 373:
```suggestion
class TablesInMountsMigrator(TabledMigrate):
```

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -305,7 +307,14 @@ def _append_records(self, items: Sequence[Table]):
     def _try_load(self) -> Iterable[Table]:
         """Tries to load table information from the database or throws TABLE_OR_VIEW_NOT_FOUND error"""
         for row in self._fetch(
-            f"SELECT * FROM {escape_sql_identifier(self.full_name)} WHERE NOT STARTSWITH(database, 'mounted_')"
+            f"SELECT * FROM {escape_sql_identifier(self.full_name)} WHERE NOT STARTSWITH(database, '{self.TABLE_IN_MOUNT_DB}')"
```
qziyuan, line 310: Could you elaborate why `_try_load` select tables that are not in `mounted_*` databases?

src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -367,3 +368,85 @@ def _try_fetch(self) -> Iterable[MigrationStatus]:
...
+    def create_tables_in_uc(self):
+        create_table_tasks = []
+        mapping_keys = {mapping.src_table: mapping for mapping in self._tm.load()}
+        for table in self._tim.fetch_all():
+            if self._tim.TABLE_IN_MOUNT_DB not in table.database:
+                logger.info(f"Path {table.location} hasn't been identified as a table in a mount, ignoring")
+                continue
+            if not table.location:
+                continue
+            if table.location not in mapping_keys:
```
qziyuan, line 395: Isn't the `table.location` the full path of the location, while the keys in `mapping_keys` are the source table name which is the `os.path.basename(path)` in this case? Should the code extract the basename from the `table.location` here and then check if it's in `mapping_keys`? Or do I misunderstand it?

src/databricks/labs/ucx/hive_metastore/table_migrate.py
qziyuan, line 386: What if there is already a HMS table on top on the same location crawled by `TablesInMounts`? Is it possible that the `tables` table in inventory database has two items that pointing to a same location? For example something like:
1. hive_metastore, schema, hms_table, EXTERNAL, s3://bucket/table_delta
2. hive_metastore, mounted_bucket, table_delta, EXTERNAL, s3://bucket/table_delta

qziyuan (changes requested):

src/databricks/labs/ucx/hive_metastore/locations.py
william-conti: Because this job aims to be incremental. Each time we run this job, we are fetching everything except all tables in mounts from previous run

src/databricks/labs/ucx/hive_metastore/table_migrate.py
william-conti: the keys in mapping_keys are the source table name, but it's not `os.path.basename(path)`. It's actually the full table path.

Basically we're checking here if the table path exists in the mapping file. Because the mapping file maps a table path to a catalog.database.table

src/databricks/labs/ucx/hive_metastore/table_migrate.py
qziyuan: In this case, the execution of `sql_migrate_table_in_mount` will fail. You should capture the exception and skip the migration if the failure is table already exists or location overlap.

qziyuan (changes requested):

src/databricks/labs/ucx/hive_metastore/table_migrate.py
qziyuan: In this case, the execution of `sql_migrate_table_in_mount` will fail. We should capture the failure and skip the migration if the failure is due to table exists or location overlap.

src/databricks/labs/ucx/hive_metastore/locations.py
qziyuan: I see, your override of `_append_records` is overwriting the `tables` inventory table. Please make sure to document to let user know this workflow will overwrite inventory table, so don't run it and the assessment workflow at the same time.

src/databricks/labs/ucx/hive_metastore/table_migrate.py
qziyuan: But the `os.path.basename(path)` is put as the table name [here](https://github.com/databrickslabs/ucx/blob/362c5475f2b8d6f38175787d61b48d14409dbe86/src/databricks/labs/ucx/hive_metastore/locations.py#L331C1-L335C38)

src/databricks/labs/ucx/hive_metastore/table_migrate.py
qziyuan: @william-conti Maybe we can filter out the table in mounts by skipping the `loaded_records` that have location already presented in `cached_results` in the `_snapshot` function, to solve this conflict issue.

qziyuan (changes requested):

qziyuan (changes requested):
Please add documents for the new workflow in the `Table Migration Workflow` section if README.md

src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -367,3 +368,85 @@ def _try_fetch(self) -> Iterable[MigrationStatus]:
...
+    ):
+        self._backend = backend
+        self._ws = ws
+        self._tm = table_mapping
+        self._tim = tables_in_mounts
+
+    def create_tables_in_uc(self):
+        create_table_tasks = []
+        mapping_keys = {mapping.src_table: mapping for mapping in self._tm.load()}
+        for table in self._tim.fetch_all():
```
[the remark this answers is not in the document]
william-conti: Yes, the _migrate_table_in_mount sets upgraded_from table property, this way get_seen_tables will work as intended

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -302,6 +304,12 @@ def _snapshot(self, fetcher: ResultFn, loader: ResultFn) -> list[Result]:
         self._append_records(loaded_records)
         return loaded_records
 
+    def _init_seen_tables(self, loaded_records: Iterable[Table]):
+        for rec in loaded_records:
+            if not rec.location:
+                continue
+            self._seen_tables[rec.location] = rec.key
```
nfx, line 311:
```suggestion
    def _init_seen_tables(self, loaded_records: Iterable[Table]) -> dict[str, str]:
        seen = {}
        for rec in loaded_records:
            if not rec.location:
                continue
            seen[rec.location] = rec.key
        return seen
```

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -94,6 +98,12 @@ def sql_alter_from(self, target_table_key, ws_id):
             f" , '{self.UPGRADED_FROM_WS_PARAM}' = '{ws_id}');"
         )
 
+    def sql_table_in_mount_alter_from(self, target_table_key):
```
nfx, line 101: add a type hint for an argument of the public method. it's not clear what the key is

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -156,6 +172,9 @@ def sql_migrate_dbfs(self, target_table_key):
     def sql_migrate_view(self, target_table_key):
         return f"CREATE VIEW IF NOT EXISTS {escape_sql_identifier(target_table_key)} AS {self.view_text};"
 
+    def sql_migrate_table_in_mount(self, target_table_key, schema):
```
nfx, line 175: add type hints

src/databricks/labs/ucx/hive_metastore/workflows.py
```diff
@@ -26,14 +38,9 @@ def migrate_dbfs_root_delta_tables(self, ctx: RuntimeContext):
...
-        super().__init__('migrate-tables-in-mounts-experimental')
-
-    @job_task
-    def scan_tables_in_mounts_experimental(self, ctx: RuntimeContext):
-        """[EXPERIMENTAL] This workflow scans for Delta tables inside all mount points
-        captured during the assessment. It will store the results under the `tables` table
-        located under the assessment."""
-        ctx.tables_in_mounts.snapshot()
+    @job_task(job_cluster="table_migration", depends_on=[ScanTablesInMounts.scan_tables_in_mounts_experimental])
+    def migrate_tables_in_mounts_experimental(self, ctx: RuntimeContext):
```
nfx, line 42: can we have it in a separate experimental workflow? if it's in the `migrate-tables` workflow, then we have no guarantee that `ScanTablesInMounts.scan_tables_in_mounts_experimental` was executed before

tests/integration/conftest.py
```diff
@@ -271,6 +271,20 @@ def with_table_mapping_rule(
...
+            self._tables.append(
+                TableInfo(
+                    catalog_name=table.catalog,
+                    schema_name=table.database,
+                    name=table.name,
+                    storage_location=table.location,
+                    table_type=TableType(table.object_type),
+                    data_source_format=DataSourceFormat(table.table_format),
+                )
+            )
```
nfx, line 286:
```suggestion
```
use `runtime_ctx.make_table`, don't break the class that used by literally every test.

tests/integration/hive_metastore/test_migrate.py
```diff
@@ -492,3 +498,64 @@ def test_migrate_managed_tables_with_principal_acl_azure(
...
+    make_schema,
+    make_table,
+    env_or_skip,
+    make_random,
+    runtime_ctx,
+):
+    if not ws.config.is_azure:
+        pytest.skip("temporary: only works in azure test env")
+    tbl_path = make_random(4).lower()
+    src_external_table = make_table(
```
nfx, line 517:
```suggestion
    src_external_table = runtime_ctx.make_table(
```

src/databricks/labs/ucx/hive_metastore/grants.py
```diff
@@ -415,13 +415,18 @@ def __init__(
         tables_crawler: TablesCrawler,
         mounts_crawler: Mounts,
         cluster_locations: dict[str, dict],
+        default_table_owner: str | None = None,
```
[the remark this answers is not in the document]
william-conti: Yep, but then we'll have to pylint ignore:too-many-arguments there then

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -293,6 +294,7 @@ def _snapshot(self, fetcher: ResultFn, loader: ResultFn) -> list[Result]:
         cached_results = []
         try:
             cached_results = list(fetcher())
+            self._init_seen_tables(cached_results)
```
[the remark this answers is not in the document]
william-conti: This is because the crawler can find tables that are already registered in the HMS of the current workspace, which will result in duplicates.
To avoid running into conflicts, it's better to filter tables in mounts that were found during the assessment.

src/databricks/labs/ucx/hive_metastore/workflows.py
```diff
@@ -27,13 +27,25 @@ def migrate_dbfs_root_delta_tables(self, ctx: RuntimeContext):
...
     def __init__(self):
-        super().__init__('migrate-tables-in-mounts-experimental')
+        super().__init__('scan-tables-in-mounts-experimental')
 
     @job_task
     def scan_tables_in_mounts_experimental(self, ctx: RuntimeContext):
         """[EXPERIMENTAL] This workflow scans for Delta tables inside all mount points
         captured during the assessment. It will store the results under the `tables` table
         located under the assessment."""
         ctx.tables_in_mounts.snapshot()
```
qziyuan, line 39: Please integration test TablesInMounts as well.

src/databricks/labs/ucx/hive_metastore/mapping.py
```diff
@@ -191,20 +191,41 @@ def _get_table_in_scope_task(self, table_to_migrate: TableToMigrate) -> TableToM
...
-                return True
-            upgraded_from = table_info.properties.get("upgraded_from")
-            if upgraded_from and upgraded_from != src_table.key:
-                raise ResourceConflict(
-                    f"Expected to be migrated from {src_table.key}, but got {upgraded_from}. "
-                    "You can skip this error using the CLI command: "
-                    "databricks labs ucx skip "
-                    f"--schema {src_table.database} --table {src_table.name}"
-                )
+        table_info = self._try_get_table_in_uc(target_key)
```
qziyuan, line 203: Here we just check if the target table is in UC or not. It's possible an UC table with different name took this location already and fail the table creation later. But there is no easy way to check if a location is already occupied in UC, so better to try catch the the table creation DDL execution.

qziyuan (changes requested):
More comments

README.md
```diff
@@ -413,6 +413,7 @@ See more details in [Table migration commands](#table-migration-commands)
 ### Table Migration Workflow Tasks
 - `migrate_dbfs_root_delta_tables` - Migrate delta tables from the DBFS root using deep clone, along with legacy table ACL migrated if any.
 - `migrate_external_tables_sync` - Migrate external tables using [`SYNC`](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-aux-sync.html) command, along with legacy table ACL migrated if any.
+- `migrate_tables_in_mounts_experimental` - An experimental task that migrates tables in mount points using a `CREATE TABLE` command, optinally sets a default tables owner if provided in `default_table_owner` conf parameter. You must run the `scan_tables_in_mounts_experimental` workflow before running this task, otherwise it will do nothing.
```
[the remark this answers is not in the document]
william-conti: After testing it somehow needs to have azure_storage_account_info.csv to work ...

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -371,7 +402,8 @@ def _find_delta_log_folders(self, root_dir: str, delta_log_folders=None) -> dict
     def _assess_path(
         self, file_info: FileInfo, delta_log_folders: dict[str, Table], root_path: str
     ) -> TableInMount | None:
-        if file_info.name == "_delta_log/":
+        # Depends of execution runtime, with SDK, dbutils.fs.list returns _delta_log, a cluster will return _delta_log/
+        if file_info.name in {"_delta_log/", "_delta_log"}:
```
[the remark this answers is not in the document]
william-conti: Could be, but I'll like to remain those out of scope for the moment and handle those in a separate PR

README.md
```diff
@@ -427,6 +428,20 @@ See more details in [Table migration commands](#table-migration-commands)
 - Consider create an instance pool, and set the instance pool id when being asked during the UCX installation. This instance pool will be put into the cluster policy used by all UCX workflows job clusters.
 - You may also manually edit the job cluster configration per job or per task after the workflows are deployed.
 
+### [EXPERIMENTAL] Scan tables in mounts Workflow
+- This experimental workflow attemps to find all Tables inside mount points that are present on your workspace.
+- If you do not run this workflow, then `migrate_tables_in_mounts_experimental` won't do anything.
+- It writes all results to `hive_metastore.<inventory_database>.tables`, you can query those tables found by filtering on database values that starts with `mounted_`
```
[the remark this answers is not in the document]
william-conti: It doesn't overwrite the entire table completely, only the tables found by TablesInMounts are overwritten, the other tables found in the assessment remains untouched

src/databricks/labs/ucx/hive_metastore/mapping.py
```diff
@@ -191,20 +191,41 @@ def _get_table_in_scope_task(self, table_to_migrate: TableToMigrate) -> TableToM
...
+        if not table_info:
+            return False
+        # Corner case for tables in mounts where the table exists in UC, but the location is not the same
+        # from the one provided in the mapping
+        if src_table.is_table_in_mount and table_info.storage_location != src_table.location:
+            raise ResourceConflict(
```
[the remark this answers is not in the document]
william-conti: Failing is better, because the user needs to do something for tables that has been recreated on a different storage location

src/databricks/labs/ucx/hive_metastore/mapping.py
william-conti: Indeed, this currently happens at my customer, the good thing is that the code throws a PERMISSION_DENIED error with the table name in UC that cause the issue, so it can be easily handled.

william-conti: Additionally we're having a situation where tables Delta tables are corrupted, and thus we aren't able to create tables there

README.md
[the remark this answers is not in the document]
qziyuan: You are doing `self._backend.save_table(self.full_name, items, Table, mode="overwrite")`. So there is a chance that this will happen:

1. `hive_metastore.<inventory_database>.tables` is empty at beginning.
2. `assessment` is running, it loads nothing from `hive_metastore.<inventory_database>.tables` and hence start crawl tables from scratch.
3. `scan_tables_in_mounts_experimental` is running, it loads nothing from `hive_metastore.<inventory_database>.tables` and hence the `loaded_records` will be all crawled tables in mount only.
4. `assessment` finished first, and append the crawled tables into `hive_metastore.<inventory_database>.tables`.
5. `scan_tables_in_mounts_experimental` finished later, and overwrite the `hive_metastore.<inventory_database>.tables` with crawled tables in mount only, and because of this the data in step 4 is lost.

src/databricks/labs/ucx/hive_metastore/locations.py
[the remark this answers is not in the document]
qziyuan: But the issue for this PR seems to be: if it sees `file_info.name` contains `=` it returns `TableInMount(format="PARQUET", is_partitioned=True)`. What would happen if the files inside this `file_info.name` partition folder are actually CSV and JSON?

src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -207,6 +215,20 @@ def _migrate_dbfs_root_table(self, src_table: Table, rule: Rule, grants: list[Gr
         self._backend.execute(src_table.sql_alter_from(rule.as_uc_table_key, self._ws.get_workspace_id()))
         return self._migrate_acl(src_table, rule, grants)
 
+    def _migrate_table_in_mount(self, src_table: Table, rule: Rule, grants: list[Grant] | None = None):
+        target_table_key = rule.as_uc_table_key
+        fields = []
+        for key, value, _ in self._backend.fetch(f"DESCRIBE TABLE delta.`{src_table.location}`;"):
```
qziyuan, line 221: The `src_table` could also be parquet, json, or csv right? If this PR migrate delta in mounts only, then the 
```python
if self.is_table_in_mount:
    return What.TABLE_IN_MOUNT
```
should be corrected.

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -329,18 +337,41 @@ def _crawl(self):
...
                 guess_table = os.path.basename(path)
+                table_location = self._get_table_location(mount, path)
+                if table_location in self._seen_tables:
+                    logger.info(
+                        f"Path {table_location} is identified as a table in mount, but is present in current workspace as a registered table {self._seen_tables[table_location]}"
+                    )
+                    continue
+                if path in self._seen_tables:
+                    logger.info(
+                        f"Path {table_location} is identified as a table in mount, but is present in current workspace as a registered table {self._seen_tables[path]}"
```
qziyuan, line 348: `table_location` should be `path`

qziyuan (changes requested):

src/databricks/labs/ucx/hive_metastore/table_migrate.py
william-conti: True, I'll edit this to add `self.is_delta` as a condition

src/databricks/labs/ucx/hive_metastore/locations.py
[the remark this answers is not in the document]
william-conti: Correct, in this case it's a bug and the logic has to be pretty much rewritten to handle those small cases.

Currently the logic handles partitioned Delta and Parquet, and by default partitioned CSVs and JSONs would be identified as partitioned.

src/databricks/labs/ucx/hive_metastore/workflows.py
william-conti: It's actually very hard to integration test TablesInMounts because `dbutils` acts differently when used in the SDK vs when used in a Databricks cluster.
For example, calling dbutils.fs.ls:
- returns dbfs:/mnt/xxx on SDK
- returns dbfs:/mnt/xxx/ on a cluster

This code is made to handle the behaviour of a Spark cluster as much as possible.

Additionally, some utility methods are missing in SDK, like isDir()

@nfx FYI

src/databricks/labs/ucx/hive_metastore/locations.py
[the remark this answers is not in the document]
william-conti: https://github.com/databrickslabs/ucx/issues/1389

nfx (approved):
lgtm

PR #1279, opened 2024-04-05
Build notebook dependency graph for `%run` cells

### LINKED ISSUE #1201: [FEATURE]: Detect notebook include graph by analysing `%run` magic cells

### Problem statement

From https://docs.databricks.com/en/notebooks/notebook-workflows.html:

The %run command allows you to include another notebook within a notebook. You can use %run to modularize your code, for example by putting supporting functions in a separate notebook. You can also use it to concatenate notebooks that implement the steps in an analysis. When you use %run, the called notebook is immediately executed and the functions and variables defined in it become available in the calling notebook.

image

### Proposed Solution

Detect `%run` magic cells and treat them as a dependency

### Additional Context

_No response_

ericvergnaud:
## Changes
 - create a Notebook class that parses source code and splits it into cells
 - create a NotebookDependencyGraph class and related utils
 - discover dependencies in `%run` cells

### Linked issues
Resolves #1201 

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests
- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -0,0 +1,331 @@
+from __future__ import annotations  # for type hints
```
nfx, line 1: do we really need it?..

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -0,0 +1,331 @@
...
+        try:
+            ast = parse_python(self._original_code)
+            return ast is not None
+        except SyntaxError:
+            return False
+
+    def build_dependency_graph(self, parent: NotebookDependencyGraph):
+        pass  # TODO
+        # see https://github.com/databrickslabs/ucx/issues/1200
+        # see https://github.com/databrickslabs/ucx/issues/1202
```
nfx, line 57:
```suggestion
        # TODO https://github.com/databrickslabs/ucx/issues/1200
        # TODO https://github.com/databrickslabs/ucx/issues/1202
        pass
```

small nit - i'm planning to create issues from TODO automatically and let's have them follow the same'ish convention

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -0,0 +1,331 @@
...
+            if start >= 0:
+                path = line[start + len(command) :].strip()
+                parent.register_dependency(path.strip('"'))
+                return
+        raise ValueError("Missing notebook path in %run command")
+
+
+class CellLanguage(Enum):
+    # long magic_names must come first to avoid shorter ones being matched
+    PYTHON = Language.PYTHON, 'python', '#', PythonCell
```
nfx, line 139: oh, we can do that?..

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -0,0 +1,331 @@
...
+                line = line[len(magic_language_prefix) :]
+                return CellLanguage.of_magic_name(line.strip())
+            return None
+        return None
+
+    def new_cell(self, source: str) -> Cell:
+        return self._new_cell(source)
+
+
+def extract_cells(source: str, default_language: CellLanguage) -> list[Cell] | None:
```
nfx, line 193: shouldn't it be a method on `CellLanguage`?

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -0,0 +1,331 @@
...
+    cell_lines: list[str] = []
+    separator = f"{default_language.comment_prefix}{CELL_SEPARATOR}"
+    for i in range(1, len(lines)):
+        line = lines[i].strip()
+        if line.startswith(separator):
+            cell = make_cell(cell_lines)
+            cells.append(cell)
+            cell_lines = []
+        else:
+            cell_lines.append(lines[i])
```
nfx, line 222:
```suggestion
            continue
        cell_lines.append(lines[i])
```

tests/unit/source_code/test_files.py
```diff
@@ -68,7 +68,7 @@ def test_files_walks_directory():
     languages.linter(Language.PYTHON).lint.return_value = [Mock(code='some-code')]
     languages.fixer.return_value = None
     files = Files(languages)
-    path = Path(__file__).parent
+    path = Path(Path(__file__).parent.parent, "aws/")
```
ericvergnaud, line 71: seems unrelated but fails with this PR because source_code folder now contains non-python files

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -0,0 +1,331 @@
...
+        return self._new_cell(source)
+
+
+def extract_cells(source: str, default_language: CellLanguage) -> list[Cell] | None:
+    lines = source.split('\n')
+    header = f"{default_language.comment_prefix}{NOTEBOOK_HEADER}"
+    if not lines[0].startswith(header):
+        raise ValueError("Not a Databricks notebook source!")
+
+    def make_cell(lines_: list[str]):
```
nfx, line 199: probably better as a private method on CellLanguage

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -0,0 +1,331 @@
...
+
+    @classmethod
+    def of_language(cls, language: Language) -> CellLanguage:
+        return next((cl for cl in CellLanguage if cl.language == language))
+
+    @classmethod
+    def of_magic_name(cls, magic_name: str) -> CellLanguage | None:
+        return next((cl for cl in CellLanguage if magic_name.startswith(cl.magic_name)), None)
+
+    def read_cell_language(self, lines: list[str]) -> CellLanguage | None:
```
nfx, line 173:
```suggestion
    def of_lines(self, lines: list[str]) -> CellLanguage | None:
```

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -0,0 +1,331 @@
...
+        self._path = path
+        self._parent = parent
+        self._locator = locator
+        self._dependencies: dict[str, NotebookDependencyGraph] = {}
+
+    @property
+    def path(self):
+        return self._path
+
+    def register_dependency(self, path: str) -> NotebookDependencyGraph:
```
nfx, line 242:
```suggestion
    def register(self, path: str) -> NotebookDependencyGraph:
```
`_dependency` is a bit redundant

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -0,0 +1,331 @@
...
+            self._dependencies[path] = child_graph
+            return child_graph
+        # nay, create the child graph and populate it
+        child_graph = NotebookDependencyGraph(path, self, self._locator)
+        self._dependencies[path] = child_graph
+        notebook = self._locator(path)
+        notebook.build_dependency_graph(child_graph)
+        return child_graph
+
+    def locate_dependency(self, path: str) -> NotebookDependencyGraph | None:
```
nfx, line 255:
```suggestion
    def locate(self, path: str) -> NotebookDependencyGraph | None:
```

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -0,0 +1,331 @@
...
+        else:
+            cell_lines.append(lines[i])
+    if len(cell_lines) > 0:
+        cell = make_cell(cell_lines)
+        cells.append(cell)
+
+    return cells
+
+
+class NotebookDependencyGraph:
```
nfx, line 230:
```suggestion
class DependencyGraph:
```
might be less redundant this way - we're in `notebooks.py` anyway

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -0,0 +1,331 @@
...
+MAGIC_PREFIX = ' MAGIC'
+LANGUAGE_PREFIX = ' %'
+
+
+class Cell(ABC):
+
+    def __init__(self, source: str):
+        self._original_code = source
+
+    @property
```
nfx, line 23:
```suggestion
```
code migration should be an explicit action, properties are implicit actions and preferably immutable

tests/unit/source_code/notebooks/01_HL7Streaming.scala
```diff
@@ -0,0 +1,221 @@
+// Databricks notebook source
```
nfx, line 1: let's be consistent and add `.sample` suffix also to non-python notebooks. 

you can also add `.txt` :)

tests/unit/source_code/test_files.py
nfx: doesn't seem to be relevant.

tests/unit/source_code/test_notebook.py
```diff
@@ -0,0 +1,297 @@
...
+import pytest
+from databricks.sdk.service.workspace import Language
+
+from databricks.labs.ucx.source_code.notebook import Notebook, NotebookDependencyGraph
+from tests.unit import _load_sources
+
+PYTHON_NOTEBOOK_SAMPLE = (
+    "00_var_context.py.sample",
+    Language.PYTHON,
+    ['md', 'md', 'md', 'python', 'python', 'python', 'md', 'python', 'md'],
```
nfx, line 12: can you remove the irrelevant test cells?

tests/unit/source_code/test_notebook.py
```diff
@@ -0,0 +1,297 @@
...
+    ],
+)
+R_NOTEBOOK_SAMPLE = (
+    "3_SparkR_Fine Grained Demand Forecasting.r",
+    Language.R,
+    [
+        'md',
+        'md',
+        'md',
+        'r',
```
nfx, line 85: looks too long and noisy as for the test sample

nfx (changes requested):

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: you mean class enums ?
yes we can ! 
(I might win the next US elections with this catch phrase)

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: I don't think so, it doesn't change from 1 language to another, except for the comment prefix. CellLanguage is really about those variations. But it wouldn't be bad.

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: that's really a preference, correct ? Isn't it easier to reason about if/elif/else ?

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: that would return the migrated_code once it's generated. I'm not planning to implement its generation here. This property is needed for checking consolidation of the notebook from small bits.

tests/unit/source_code/test_notebook.py
ericvergnaud: mmm... it was actually super useful to use real notebooks to discover edge cases, and testing against an untouched sample creates a lot of confidence

tests/unit/source_code/test_notebook.py
ericvergnaud: maybe we can move these to json files ?

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: yes for forward type hints

tests/unit/source_code/test_files.py
ericvergnaud: see my comment, this PR breaks the test

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: I also don't like https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html

tests/unit/source_code/notebooks/01_HL7Streaming.scala
ericvergnaud: done

tests/unit/source_code/test_notebook.py
ericvergnaud: it'd be much less intrusive if `make fmt` didn't force 1 item per line. Is there a way to configure this ?

tests/unit/source_code/test_notebook.py
ericvergnaud: I'm not alone, see https://github.com/psf/black/issues/1281
Fixed by disabling `fmt` for these lines

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: Second thoughts, I think you're right. I changed it.

tests/unit/source_code/test_notebook.py
ericvergnaud: Fixed by disabling fmt for these lines

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: without it, you have to use `'MyType'` instead of `MyType`, and `'MyType' or None` instead of `MyType | None`

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: not sure about this one, the `of_xxx` methods are static and selecting an Enum value from their property, this instance method has a very different contract

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: Yeah I was hesitant about this when writing it... done

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: true if you're looking at it from the DependencyGraph class, not so much if you're looking at it from a caller point of view:  
`parent.register_dependency(path.strip('"'))`
is much better imho than
`parent.register(path.strip('"'))`

src/databricks/labs/ucx/source_code/notebook.py
ericvergnaud: same here

src/databricks/labs/ucx/source_code/notebook.py
nfx: didn't realise we can just pass a tuple in there

src/databricks/labs/ucx/source_code/notebook.py
nfx: this method is still static, no?

nfx (approved):
let's merge this

PR #1285, opened 2024-04-05
Add support for migrating Table ACL for interactive cluster in AWS using Instance Profile

### LINKED ISSUE #887: [FEATURE]: Create UC Storage Credential, Schema, and Table Grants based on AWS IAM role access

HariGS-DB:
## Changes
This PR contains the following changes for migrating Table ACL for interactive cluster used through Instance Profile in AWS scenario.

Identifies all the interactive clusters
identifies all the instance profiles used by each interactive cluster
identify the list of principals who have access to the cluster (user, group, spn)
get the prefix these instance role arn have access and the type of permission from the aws_instance_profile_info csv
get the list of external locations matching the list of prefix
get the list of tables having the matching external location prefix from the table mapping CSV
If a instance profile arn has read access on a container then give:
select on the table and view
If a instance profile arn has write access on a container then give:
all privileges on the table and view

### Linked issues
Note: The permission for schema, catalog and external location should be done as separate PR as part of the create_catalogs_schemas and migrate_locations
Resolves https://github.com/databrickslabs/ucx/issues/887

Resolves #887 

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [X] manually tested
- [X] added unit tests
- [X] added integration tests


nfx:
the code would need to be adjusted to https://github.com/databrickslabs/ucx/pull/1209

tests/integration/hive_metastore/test_migrate.py
```diff
@@ -460,6 +460,7 @@ def prepared_principal_acl(runtime_ctx, env_or_skip, make_dbfs_data_copy, make_c
     ]
     runtime_ctx.with_table_mapping_rules(rules)
     runtime_ctx.with_dummy_azure_resource_permission()
+    runtime_ctx.with_dummy_aws_resource_permission()
```
nfx, line 463:
```suggestion
```
Test has to pass without these dummy locations, also remove these for azure. It looks like a substantial bug in the logic if we require this in every test.

tests/integration/test_installation.py
```diff
@@ -690,6 +690,10 @@ def test_table_migration_job(
         ],
         filename='azure_storage_account_info.csv',
     )
+    installation.save(
+        [AWSRoleAction('arn:aws:iam::12345:instance-profile/role1', 's3', 'WRITE_FILES', 's3://storage5/*')],
```
nfx, line 694: Why do we need these for azure tests to pass? This is a bug

src/databricks/labs/ucx/contexts/application.py
```diff
@@ -102,6 +103,10 @@ def is_azure(self) -> bool:
     def is_aws(self) -> bool:
         return self.connect_config.is_aws
 
+    @cached_property
+    def is_gcp(self) -> bool:
+        return self.connect_config.is_gcp
```
nfx, line 108:
```suggestion
        return not self.is_aws and not self.is_azure
```

src/databricks/labs/ucx/contexts/application.py
```diff
@@ -236,11 +241,23 @@ def azure_acl(self):
...
+            self.sql_backend,
+            self.installation,
+        )
+
     @cached_property
     def principal_acl(self):
-        if not self.is_azure:
-            raise NotImplementedError("Azure only for now")
-        eligible = self.azure_acl.get_eligible_locations_principals()
+        eligible = {}
```
nfx, line 254: Introduce a method to increase readability

nfx (changes requested):
Looks like there are couple of hidden bugs in the logic

tests/integration/test_installation.py
HariGS-DB: we dont need this, i just tried to reuse the same fixture, but will spilt this logic separately into the azure and was test case

tests/integration/test_installation.py
```diff
@@ -676,9 +672,37 @@ def test_table_migration_job(
...
+                    'dummy_application_id',
+                    'principal_1',
+                    'WRITE_FILES',
+                    'Application',
+                    'directory_id_ss1',
+                )
+            ],
+            filename='azure_storage_account_info.csv',
+        )
+    if ws.config.is_aws:
```
nkvuong, line 692: aws shouldn't use `StoragePermissionMapping`

tests/integration/test_installation.py
```diff
@@ -767,16 +778,11 @@ def test_table_migration_job_cluster_override(  # pylint: disable=too-many-local
...
 ):
-    # TODO - DBFS API does not support instance profile based mounts
-    if not ws.config.is_azure:
-        pytest.skip("Temporarily does not work on AWS")
     # create external and managed tables to be migrated
     schema = make_schema(catalog_name="hive_metastore", name=f"migrate_{make_random(5).lower()}")
     src_managed_table = make_table(schema_name=schema.name)
     existing_mounted_location = f'dbfs:/mnt/{env_or_skip("TEST_MOUNT_NAME")}/a/b/c'
-    new_mounted_location = f'dbfs:/mnt/{env_or_skip("TEST_MOUNT_NAME")}/a/b/{make_random(4)}'
-    make_dbfs_data_copy(src_path=existing_mounted_location, dst_path=new_mounted_location)
```
nkvuong, line 778: this will be fixed in #1403

HariGS-DB: Cool, but even the current approach is not wrong. Is there a risk doing this way?

HariGS-DB:
Screenshot of the failed test run manually
image

Issue #331, opened 2023-09-29
Migrate views


nfx:
1. Make a "best effort" attempt to upgrade view
1. Create a view in the new location
1. Upgrade table reference to the new tables (based on the `upgraded_to` table property)
1. Handle nested views
1. Handle or highlight other cases (functions/storage references/ETC)
1. Create an exception list with views failures
1. Topologically sort all views, as there might be cross-dependencies

Upstream dependencies:
- https://github.com/databrickslabs/ucx/issues/670

Related issues:
- https://github.com/databrickslabs/ucx/issues/315

zpappa:
Add in ordering for nested views

Issue #1215, opened 2024-04-01
[FEATURE]: Detect and process `USE <database>` statements


nfx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

USE statement is used to set the current database. After the current database is set, the unqualified database artifacts such as tables, functions and views that are referenced by SQLs are resolved from the current database. The default database name is ‘default’.

```
USE database_name
```

```
spark.sql("use something")
spark.table("anything").show()
```

the second statement would show records from `something.anything`, not `default.anything`.

See https://spark.apache.org/docs/3.0.0-preview/sql-ref-syntax-qry-select-usedb.html

The following test case has to produce three different table advice:

```
# Databricks notebook source
display(spark.table('people')) # we are looking at default.people table

# COMMAND ----------

# MAGIC %sql USE something

# COMMAND ----------

display(spark.table('persons')) # we are looking at something.persons table

# COMMAND ----------

spark.sql('USE whatever')

# COMMAND ----------

display(spark.table('kittens')) # we are looking at whatever.kittens table
```

### Proposed Solution

have linters/fixers pickup `USE database` statements if they occur in the code to resolve single-table-name lookups

nfx:
@jimidle

jimidle:
Currently the tests etc supply the schema and if they do not, it is defaulting to None. Should I make it default to 'default'? It seems so.

PR #1288, opened 2024-04-05
Added support for one level notation for query transformer.

FastLee:
Related to #331 
 The query transformer now supports one-level notation for schema and table name, with a new `use_schema` parameter in the `apply` method. A new test case, `test_fully_migrated_queries_match_no_db`, demonstrates this feature. The `FromTable` class is used to apply the transformation, with the `use_schema` parameter specifying the schema. This allows for more flexibility in applying transformations to queries with multiple tables and different schemas.

### DIFF
--- a/src/databricks/labs/ucx/source_code/queries.py
+++ b/src/databricks/labs/ucx/source_code/queries.py
@@ -41,7 +41,7 @@ class FromTable(Linter, Fixer):
             return table.catalog
         return 'hive_metastore'
 
-    def apply(self, code: str) -> str:
+    def apply(self, code: str, *, use_schema: str = "") -> str:
         new_statements = []
         for statement in sqlglot.parse(code):
             if not statement:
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/source_code/queries.py
```diff
@@ -41,7 +41,7 @@ def _catalog(table):
             return table.catalog
         return 'hive_metastore'
 
-    def apply(self, code: str) -> str:
+    def apply(self, code: str, *, use_schema: str = "") -> str:
```
nfx, line 44: You are not allowed to change the Fixer interface.

Why do you need a parameter here? Please explain. See also:

- https://github.com/databrickslabs/ucx/issues/1215

nfx (changes requested):

ganeshgirase (commented):
method lint and apply have almost around 8 - 10 lines common which is to get statement , validate statement, tables and their catalogs. This can be generalised into common code and we can achieve lazy evaluation here.

nfx:
as discussed, this is not the right way to implement it

Issue #1202, opened 2024-04-01
[FEATURE]: Detect notebook include graph by analysing workspace file imports and `sys.path` manipulation


nfx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

With Databricks Runtime 11.2 and above, you can create and manage source code files in the Databricks workspace, and then import these files into your notebooks as needed. 

Related info:
- https://docs.databricks.com/en/files/workspace-modules.html
- https://docs.databricks.com/en/notebooks/share-code.html

The following list orders precedence from highest to lowest. In this list, a lower number means higher precedence.

1. Libraries in the current working directory (Git folders only).
3. Libraries in the Git folder root directory (Git folders only).
5. Notebook-scoped libraries (%pip install in notebooks).
7. Cluster libraries (using the UI, CLI, or API).
9. Libraries included in Databricks Runtime.
11. Libraries installed with init scripts might resolve before or after built-in libraries, depending on how they are installed. Databricks does not recommend installing libraries with init scripts.
13. Libraries in the current working directory (not in Git folders).
15. Workspace files appended to the sys.path.

<img width="855" alt="image" src="https://github.com/databrickslabs/ucx/assets/259697/5c862db4-1c85-4aec-a1f7-85a9874514d2">

### Proposed Solution

Detect any import statements and check for `sys.path` manipulations to treat imported workspace files as a dependency

ericvergnaud:
@nfx the 'Python library precedence' link above returns an empty file (maybe it's only when called from outside DB account)

ericvergnaud:
Splitting into more granular issues:
#1342

ericvergnaud:
@nfx from its title, this ticket only applies to workspaces. Should we change the title or create a ticket for local file system ?

PR #1344, opened 2024-04-10
Refactor dependency graph

### LINKED ISSUE #1342: [FEATURE]: Support not only Notebooks in dependency graph

### Problem statement

Notebooks can run files that are not notebooks, than may require migration.
The dependency graph should reflect that.

### Proposed Solution

Have the dependency graph work with any object_type, not just notebooks

### Additional Context

Sub-ticker from #1202

ericvergnaud:
## Changes
Refactor `DependencyGraph`
Introduce 'DependencyLoader' and utility classes

### Linked issues
#1202 

Resolves #1342 

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests
- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests


ericvergnaud:
performance_test is throwing a cyclical-dependency error, I haven't yet found a way to get rid of it, any help appreciated...

src/databricks/labs/ucx/source_code/dependencies.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class Dependency:
+
+    @staticmethod
+    def from_object_info(object_info: ObjectInfo):
+        assert object_info.path is not None
+        return Dependency(object_info.object_type, object_info.path)
+
+    def __init__(self, object_type: ObjectType | None, path: str):
```
nfx, line 17:
```suggestion
    def __init__(self, object_type: ObjectType, path: str):
```
let's simplify the handling and ignore objects with no types

src/databricks/labs/ucx/source_code/dependencies.py
```diff
@@ -0,0 +1,164 @@
...
+    @abc.abstractmethod
+    def object_type(self) -> ObjectType:
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def build_dependency_graph(self, graph) -> None:
+        raise NotImplementedError()
+
+
+class DependencyLoader:
```
nfx, line 48: dependency loader has to be an interface. There has to be `LocalDependencyLoader` and `WorkspaceDependencyLoader`

nfx (changes requested):
dependency loader has to be an interface, the rest is lgtm

src/databricks/labs/ucx/source_code/dependencies.py
ericvergnaud: well the intent is to manage both workspace and local dependencies from here...

Issue #1282, opened 2024-04-05
[FEATURE]: Cli command to get errors and warnings from the latest workflow run


JCZuurmond:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

An extension to #1148, a cli command to get errors and warnings from the latest workflow run.

### Proposed Solution

The cli command fetches the records from the `logs` table and writes them to stdout so that they become visible to the cli user.

PR #1349, opened 2024-04-10
Forward remote logs on `run_workflow` and removed `destroy-schema` workflow in favour of `databricks labs uninstall ucx`

### LINKED ISSUE #1186: run_workflow(…) should forward remote logs through local logging interface after job finishes executing

### Problem statement

Integration tests and CLI commands, while running remote workflows, don’t show any logs 🪵 

this is especially frustrating for troubleshooting the integration test failures with traces onto github comments (and issues).

### Proposed Solution

- pytest doesn’t capture logs if they are just sys.stderr.write 
- Remote log retrieval and parsing is tracked in https://github.com/databrickslabs/ucx/issues/1148

### Additional Context

_No response_

nfx:
Fix #1186

### DIFF
--- a/src/databricks/labs/ucx/installer/logs.py
+++ b/src/databricks/labs/ucx/installer/logs.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from typing import TextIO
 
 from databricks.labs.lsql.backends import SqlBackend
+from databricks.sdk.errors import InternalError
 
 logger = logging.getLogger(__name__)
 
@@ -150,7 +151,7 @@ class TaskRunWarningRecorder:
         """Parse and store the logs for all tasks.
 
         Raises:
-            RuntimeError: When error or above log records are found, after storing the logs.
+            InternalError: When error or above log records are found, after storing the logs.
         """
         log_files = self._log_path.glob("*.log*")
 
@@ -170,10 +171,15 @@ class TaskRunWarningRecorder:
             mode="append",
         )
 
-        error_log_records = [
-            log_record for log_record in log_records if logging.getLevelName(log_record.level) >= logging.ERROR
-        ]
-        if len(error_log_records) > 0:
-            error_log_messages = "\n".join([str(log_record) for log_record in error_log_records])
-            raise RuntimeError(f"Found error log records:\n{error_log_messages}")
+        error_messages = []
+        for log_record in log_records:
+            if logging.getLevelName(log_record.level) < logging.ERROR:
+                continue
+            message = log_record.message
+            # Ignore the prompt to troubleshoot, which is the error
+            if 'databricks workspace export /' in message:
+                continue
+            error_messages.append(message)
+        if len(error_messages) > 0:
+            raise InternalError("\n".join(error_messages))
         return log_records
--- a/src/databricks/labs/ucx/installer/logs.py
+++ b/src/databricks/labs/ucx/installer/logs.py
@@ -9,6 +9,8 @@ from typing import TextIO
 from databricks.labs.lsql.backends import SqlBackend
 from databricks.sdk.errors import InternalError
 
+from databricks.labs.ucx.framework.tasks import TaskLogger
+
 logger = logging.getLogger(__name__)
 
 
@@ -90,6 +92,7 @@ class TaskRunWarningRecorder:
         job_run_id: int,
         sql_backend: SqlBackend,
         schema: str,
+        attempt: int = 0,
     ):
         """
         Initializes a LogProcessor instance.
@@ -111,7 +114,7 @@ class TaskRunWarningRecorder:
         self._sql_backend = sql_backend
         self._schema = schema
 
-        self._log_path = Path(install_dir) / "logs" / self._workflow / f"run-{self._job_run_id}"
+        self._log_path = TaskLogger.log_path(Path(str(install_dir)), workflow, job_run_id, attempt)
 
     @property
     def full_name(self) -> str:
--- a/src/databricks/labs/ucx/installer/workflows.py
+++ b/src/databricks/labs/ucx/installer/workflows.py
@@ -41,6 +41,7 @@ from databricks.sdk.errors import (
 from databricks.sdk.retries import retried
 from databricks.sdk.service import compute, jobs
 from databricks.sdk.service.jobs import RunLifeCycleState, RunResultState
+from databricks.sdk.service.workspace import ObjectType
 
 import databricks
 from databricks.labs.ucx.config import WorkspaceConfig
@@ -54,6 +55,7 @@ logger = logging.getLogger(__name__)
 EXTRA_TASK_PARAMS = {
     "job_id": "{{job_id}}",
     "run_id": "{{run_id}}",
+    "attempt": "{{job.repair_count}}",
     "parent_run_id": "{{parent_run_id}}",
 }
 DEBUG_NOTEBOOK = """# Databricks notebook source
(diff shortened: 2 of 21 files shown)


tests/unit/assessment/test_workflows.py
```diff
@@ -82,5 +82,5 @@ def test_incompatible_submit_runs(run_workflow):
 
 
 def test_runtime_destroy_schema(run_workflow):
```
JCZuurmond, line 84: The test name should be updated too

src/databricks/labs/ucx/runtime.py
```diff
@@ -71,24 +65,23 @@ def trigger(self, *argv):
         print(f"UCX v{__version__}")
         workflow = self._workflows[workflow_name]
         if task_name == "parse_logs":
-            if workflow != "099-destroy-schema":  # Cannot store logs after schema is destroyed
```
JCZuurmond, line 74: Good that this one is removed 👍

tests/unit/assessment/test_workflows.py
nfx, line 84:
```suggestion
def test_failing_task_raises_value_error(run_workflow):
```

src/databricks/labs/ucx/installer/logs.py
```diff
@@ -170,10 +171,15 @@ def snapshot(self) -> list[LogRecord]:
...
-        if len(error_log_records) > 0:
-            error_log_messages = "\n".join([str(log_record) for log_record in error_log_records])
-            raise RuntimeError(f"Found error log records:\n{error_log_messages}")
+        error_messages = []
+        for log_record in log_records:
+            if logging.getLevelName(log_record.level) < logging.ERROR:
+                continue
+            message = log_record.message
+            # Ignore the prompt to troubleshoot, which is the error
+            if 'databricks workspace export /' in message:
```
JCZuurmond, line 180: Where is this line coming from?

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -182,6 +189,26 @@ def validate_step(self, step: str) -> bool:
                 return run_new_state is not None and run_new_state.result_state == RunResultState.SUCCESS
         return False
 
+    def _relay_logs(self, workflow, run_id):
+        for log in self._fetch_logs(workflow, run_id):
+            task_logger = logging.getLogger(log.component)
+            # we're relaying almost raw log events, so we need to call the logger directly
```
JCZuurmond, line 195: Could you elaborate this? Why not use `task_logger.log`?

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -182,6 +189,26 @@ def validate_step(self, step: str) -> bool:
...
+            raw_log_event = task_logger._log  # pylint: disable=protected-access
+            raw_log_event(logging.getLevelName(log.level), log.message, ())
+
+    def _fetch_logs(self, workflow: str, run_id: str) -> Iterator[PartialLogRecord]:
```
JCZuurmond, line 199: If we want to reuse this logic in the cli command it should be moved out of this class, see #1282

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -182,6 +189,26 @@ def validate_step(self, step: str) -> bool:
...
+        for object_info in self._ws.workspace.list(log_path):
+            if not object_info.path:
+                continue
+            if '.log' not in object_info.path:
+                continue
+            task_name = os.path.basename(object_info.path).split('.')[0]
+            with self._ws.workspace.download(object_info.path) as raw_file:
+                text_io = StringIO(raw_file.read().decode())
+            for record in parse_logs(text_io):
+                yield replace(record, component=f'{record.component}:{task_name}')
```
JCZuurmond, line 210: Why postfix the component with a `:{task_name}`?

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -364,6 +397,9 @@ def create_jobs(self, prompts):
     def _config_file(self):
         return f"{self._installation.install_folder()}/config.yml"
 
+    def _is_testing(self):
+        return self._product_info.product_name() != "ucx"
```
JCZuurmond, line 401: Does mypy complain here? In the test the `product_name` look to be `None`

JCZuurmond (commented):
Generally LGTM, have some questions 👇

src/databricks/labs/ucx/installer/logs.py
nfx: From task runner, prompting to fetch logs directly

src/databricks/labs/ucx/installer/workflows.py
nfx: It doesn't have ability to set level

src/databricks/labs/ucx/installer/workflows.py
nfx: Can be still here, see next PR 😉

src/databricks/labs/ucx/installer/workflows.py
nfx: Because assessment has 20 tasks and prefix is swallowed

src/databricks/labs/ucx/installer/workflows.py
nkvuong: this may not work since blueprint infer the product_name based on the version folder if `product_name` is `None`
```
        if self._product_name:
            return self._product_name
        version_file_folder = self._version_file.parent
        return version_file_folder.name.replace("_", "-")
```

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -403,9 +436,21 @@ def _create_readme(self) -> str:
         readme_url = self._installation.workspace_link('README')
         return readme_url
 
+    def _create_dashboard_links(self, step_name, dashboard_link):
+        dashboards_per_step = [d for d in self._install_state.dashboards.keys() if d.startswith(step_name)]
+        for dash in dashboards_per_step:
+            if len(dashboard_link) == 0:
+                dashboard_link += "Go to the one of the following dashboards after running the job:\n"
```
ganeshgirase, line 443: Does this check needs before for loop?

src/databricks/labs/ucx/installer/workflows.py
JCZuurmond: it does, right? https://docs.python.org/3/library/logging.html#logging.Logger.log

src/databricks/labs/ucx/installer/workflows.py
JCZuurmond: If I understand it correctly, the goal is show the task name again when logging the log records in this module. It would be more clear to me, when we remove the replace here and initiate the logger as `logging.getLogger(f"{log.component":{log.task_name}")`

src/databricks/labs/ucx/installer/workflows.py
nfx: oh, okay

src/databricks/labs/ucx/installer/workflows.py
ganeshgirase: @nfx Isn't this valid comment?

src/databricks/labs/ucx/installer/workflows.py
nfx: https://github.com/databrickslabs/ucx/issues/1359

src/databricks/labs/ucx/installer/workflows.py
ganeshgirase: @nfx Maybe I didn't explained it correctly. this if statement check will happened for every dash, but it get's only initialized just once. Which means say there are 100 dash in dashboards_per_step, this check will be 100 times. If we can move this check before for loop like
```suggestion
            if dashboards_per_step and len(dashboard_link) == 0:
                dashboard_link += "Go to the one of the following dashboards after running the job:\n"
```
this will be checked just once. Hope this is clear.

PR #1375, opened 2024-04-11
Enhanced migrate views task to support views created with explicit column list

FastLee:
Added a lookup based on "SHOW CREATE TABLE" to retrieve the list of columns from the create script.

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/table_migrate.py
+++ b/src/databricks/labs/ucx/hive_metastore/table_migrate.py
@@ -8,7 +8,6 @@ from databricks.labs.blueprint.parallel import Threads
 from databricks.labs.lsql.backends import SqlBackend
 from databricks.sdk import WorkspaceClient
 
-from databricks.labs.ucx.framework.utils import escape_sql_identifier
 from databricks.labs.ucx.hive_metastore import TablesCrawler
 from databricks.labs.ucx.hive_metastore.grants import Grant, GrantsCrawler, PrincipalACL
 from databricks.labs.ucx.hive_metastore.mapping import (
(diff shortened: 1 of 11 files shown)


src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -202,7 +203,14 @@ def _get_view_update_sql(self, src_table: Table, rule: Rule) -> str:
         new_view_text = ViewToMigrate.get_view_updated_text(
             src_table.view_text, self._migration_status_refresher.index(), src_table.database
         )
-        return f"CREATE VIEW IF NOT EXISTS {escape_sql_identifier(rule.as_uc_table_key)} AS {new_view_text};"
+        # Getting columns definition from the show create table output
+        create_statement = self._backend.fetch(f"SHOW CREATE TABLE {src_table.safe_sql_key}")
+        create_str = next(iter(create_statement))["createtab_stmt"]
```
nfx, line 208: Can't we feed it to get_view_updated_text to get rid of that potentially buggy "get view columns?

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/table_migrate.py
FastLee: I looked into it.
It is a 50/50 split.
It has the tblproperties that we will have to strip out. And we will have to translate the name of the view (not a big challenge).

nfx: Shouldn't we keep the properties while migrating? What properties are there?

Changing view name is just a swap of identifier, which is easy

FastLee: We shouldn't.
Most properties are irrelevant and auto generated.

PR #1376, opened 2024-04-12
Create Databricks access connector

### LINKED ISSUE #888: [FEATURE]: Create access connector with managed identity

### Problem statement

 - Create access connector with managed identity, one for each storage account
 - ~assign appropriate permission to the identified storage accounts that are to be migrated.~ -> see #1383
 - ~replace existing storage credential with equivalent access connector-managed identity~ -> see #1384

Related issues:
- https://github.com/databrickslabs/ucx/issues/875
- https://github.com/databrickslabs/ucx/issues/339

### Proposed Solution

to be defined

### Additional Context

_No response_

JCZuurmond:
## Changes
- Create access connector-managed identity

### Linked issues

Resolves #888

### Functionality 

- [ ] added new CLI command
- [x] modified existing command: `databricks labs ucx ...`
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests


tests/integration/azure/conftest.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+from databricks.labs.ucx.azure.resources import AzureAPIClient, AzureResources
+
+
+@pytest.fixture
+def azure_resources(ws):
```
nfx, line 7: Remove. We don't have permissions to run integration tests that modify azure resources on github actions

tests/integration/azure/test_resources.py
```diff
@@ -0,0 +1,83 @@
+import pytest
+from unittest.mock import create_autospec
```
nfx, line 2: Integration tests have to use real objects, not the ones created by magicmock. This is a unit test

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -94,6 +96,45 @@ class AzureRoleAssignment:
...
+    type: str
+    location: str
+    # TODO: Add identity with reference to dataclass
+    # identity {"principalId": ..., "tenantId": ..., "type": ...}
+    # The raw API call contains the following fields as well:
+    # tags
+    # properties
+    # systemData
+
+    _pattern_id = re.compile(
```
nfx, line 111: Use AzureResource, it already has the parsing for arm ids

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -140,16 +181,64 @@ def post(self, path: str, body: dict | None = None):
...
+        return access_connectors
+
+    def create(self, access_connector: AccessConnector):
+        """Create access connector.
+
+        Docs:
+            https://learn.microsoft.com/en-us/rest/api/databricks/access-connectors/create-or-update?view=rest-databricks-2023-05-01&tabs=HTTP
+        """
+        # TODO: Manual test failed, fix!
+        body = {
```
nfx, line 228: When you create, use the tags from the resource group you're creating this in

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -140,16 +181,64 @@ def post(self, path: str, body: dict | None = None):
...
+        for access_connector_raw in response["value"]:
+            access_connector = AccessConnector(
+                id=access_connector_raw["id"],
+                type=access_connector_raw["type"],
+                location=access_connector_raw["location"],
+            )
+            access_connectors.append(access_connector)
+        return access_connectors
+
+    def create(self, access_connector: AccessConnector):
```
nfx, line 221:
```suggestion
    def create(self, subscription, resource_group, name, location, managed_identity, properties, tags):
```
Don't pass in single POJO, pass in arguments

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -140,16 +181,64 @@ def post(self, path: str, body: dict | None = None):
...
         query: dict[str, str] = {}
+        if api_version is not None:
+            query = {"api-version": api_version}
         return self.api_client.do("DELETE", path, query, headers)
 
     def token(self):
         return self._token_source.token()
 
 
+class AccessConnectorClient:
```
nfx, line 196: For simplicity, shouldn't we just add all these methods to AzureResources? That's where we have all other azure resource management. We're not building an SDK library for Azure, we're using the apis we need

nfx (changes requested):
You have to create a user managed identity first

nfx:
Perhaps the user assigned identity is not necessary - https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/databricks_access_connector

src/databricks/labs/ucx/azure/resources.py
JCZuurmond: I added a tags field to the access connector. When creating the to-be-created access connectors the tags of the resource group can be merged

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -140,16 +185,73 @@ def post(self, path: str, body: dict | None = None):
...
+        while len(url) > 0:
+            response = self._azure_mgmt.get(url)
+            for access_connector_raw in response["value"]:
+                access_connector = AccessConnector(
+                    id=access_connector_raw["id"],
+                    location=access_connector_raw["location"],
+                    tags=access_connector_raw.get("tags", {}),
+                )
+                access_connectors.append(access_connector)
+            url = response.get("nextLink", "").removeprefix(self._azure_mgmt.api_client._cfg.host)
```
JCZuurmond, line 228: @nfx : Pagination with nextLink fails with :


```
E               databricks.sdk.errors.base.DatabricksError: {'code': 'ProviderError', 'message': "Resource provider 'Microsoft.Databricks' failed to return collection response for type 'accessConnectors'."}

```

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -94,6 +115,18 @@ class AzureRoleAssignment:
     role_name: str
 
 
+@dataclass
+class AccessConnector:
+    id: str
+    name: str
+    location: str
+    type: str = "Microsoft.Databricks/accessConnectors"
+    identity: dict[str, str] = field(default_factory=dict)
```
nfx, line 124:
```suggestion
    principal_id: str  # user-assigned and system-assigned
    managed_identity_id: str | None = None # user-assigned
    client_id: str | None = None # user-assigned
    tenant_id: str | None = None # system-assigned
```

doing key-value for `identity` block is not ideal. let's do concrete fields with concrete types and match the `AzureManagedIdentityRequest` structure.

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -94,6 +115,18 @@ class AzureRoleAssignment:
     role_name: str
 
 
+@dataclass
+class AccessConnector:
+    id: str
+    name: str
+    location: str
+    type: str = "Microsoft.Databricks/accessConnectors"
```
nfx, line 123:
```suggestion
    identity_type: str
```
type can be 'SystemAssigned' or UserAssigned. let's not confuse the consumer

src/databricks/labs/ucx/azure/resources.py
nfx: user assigned - take the (first) key and add it to `managed_identity_id` property:
image

system assigned:
image

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -94,6 +115,18 @@ class AzureRoleAssignment:
...
+    identity: dict[str, str] = field(default_factory=dict)
+    tags: dict[str, str] = field(default_factory=dict)
+    properties: dict[str, str] = field(default_factory=dict)
+    system_data: dict[str, str] = field(default_factory=dict)
```
nfx, line 127:
```suggestion
```
no value in those two for us

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -374,3 +410,86 @@ def managed_identity_client_id(
                 return None
             return principal.client_id
         return None
+
+    def get_access_connector(self, subscription_id: str, resource_group_name: str, name: str) -> AccessConnector:
```
nfx, line 414: move these new methods to `AzureResources` class (higher level), which calls `AzureApi` class (lower level) and reduce copy-paste:
https://github.com/databrickslabs/ucx/blob/5563b661eafaab71fd6233153094c513c6ee76de/src/databricks/labs/ucx/azure/resources.py#L481-L512

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -94,6 +115,18 @@ class AzureRoleAssignment:
     role_name: str
 
 
+@dataclass
+class AccessConnector:
```
nfx, line 119: and you probably want to isolate the `RawResource -> AccessConnector` conversion to a single method.

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -94,6 +115,18 @@ class AzureRoleAssignment:
...
+    id: str
+    name: str
+    location: str
```
nfx, line 122:
```suggestion
    location: str
    resource_group: str
    subscription_id: str
```
just in case.

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -94,6 +115,18 @@ class AzureRoleAssignment:
     role_name: str
 
 
+@dataclass
+class AccessConnector:
+    id: str
```
nfx, line 120:
```suggestion
    id: AzureResource
```
so that we can extract resource group and subscription id easily.

nfx (changes requested):

src/databricks/labs/ucx/azure/resources.py
JCZuurmond: Can be easily retrieved using the id field as changed in the other comment to AzureResource

src/databricks/labs/ucx/azure/resources.py
JCZuurmond: The `AccessConnector.type` is "Microsoft.Databricks/accessConnectors". I'll remove it as it is redundant and solve the identity fields following your other comment.

src/databricks/labs/ucx/azure/resources.py
JCZuurmond: These methods are already in AzureResources added under `AzureResources.managed_identity_client_id`

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -80,6 +101,15 @@ class Principal:
     directory_id: str | None = None
 
 
+@dataclass
+class StorageAccount:
+    id: str
+    subscription_id: str
+    resource_group: str
+    name: str
+    location: str
```
JCZuurmond, line 110: Needed this object to access a storage account's location

src/databricks/labs/ucx/azure/access.py
```diff
@@ -179,22 +179,51 @@ def create_uber_principal(self, prompts: Prompts):
...
+                storage_account_info.append(storage)
+        for storage_account in storage_account_info:
+            access_connector = self._azurerm.create_or_update_access_connector(
+                storage_account.id.subscription_id,
+                storage_account.id.resource_group,
+                f"ac-{storage_account.name}",
+                storage_account.location,
+                tags={"CreatedBy": "ucx"},
+            )
+            self._apply_storage_permission(
```
JCZuurmond, line 211: Add logic to wait for the access connector to be provisioned fully

src/databricks/labs/ucx/azure/access.py
```diff
@@ -93,29 +93,29 @@ def save_spn_permissions(self) -> str | None:
...
-                f"spark_conf.fs.azure.account.oauth2.client.id.{storage.storage_account}.dfs.core.windows.net"
-            ] = self._policy_config(uber_principal.client.client_id)
-            policy_dict[
-                f"spark_conf.fs.azure.account.oauth.provider.type.{storage.storage_account}.dfs.core.windows.net"
-            ] = self._policy_config("org.apache.hadoop.fs.azurebfs.oauth2.ClientCredsTokenProvider")
-            policy_dict[
-                f"spark_conf.fs.azure.account.oauth2.client.endpoint.{storage.storage_account}.dfs.core.windows.net"
-            ] = self._policy_config(endpoint)
-            policy_dict[f"spark_conf.fs.azure.account.auth.type.{storage.storage_account}.dfs.core.windows.net"] = (
+            policy_dict[f"spark_conf.fs.azure.account.oauth2.client.id.{storage.name}.dfs.core.windows.net"] = (
```
nfx, line 104: diff in this method doesn't seem to be strictly necessary. can we avoid it in this PR?

nfx (changes requested):

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -245,15 +348,17 @@ def subscriptions(self):
                 continue
             yield subscription
 
-    def storage_accounts(self) -> Iterable[AzureResource]:
+    def storage_accounts(self) -> Iterable[StorageAccount]:
```
JCZuurmond, line 351: @nfx : Updated this return type to yield StorageAccount. Needed the `.location` when creating the access connector.

src/databricks/labs/ucx/azure/access.py
JCZuurmond: I needed to update the `AzureResource` -> `StorageAccount` to have a location for creating the access connector. That lead to this change

pyproject.toml
```diff
@@ -285,7 +285,7 @@ attr-naming-style = "snake_case"
 # Regular expression matching correct attribute names. Overrides attr-naming-
 # style. If left empty, attribute names will be checked with the set naming
 # style.
-attr-rgx = "[a-z_][a-z0-9_]{2,}$"
```
JCZuurmond, line 288: @nfx: Need this to match the `id` attribute on `RawAzureResource`, `StorageAccount` and `AccessConnector`. Could also restrict the regex to explicitly match `id`, not any two letter attribute

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -251,3 +253,17 @@ def run(self, prompts: Prompts, include_names: set[str] | None = None) -> list[S
...
         return execution_result
+
+    def run_access_connector_migration(self, prompts: Prompts) -> list[StorageCredentialValidationResult]:
+        plan_confirmed = prompts.confirm("Please confirm to create an access connector for each storage account.")
+        if plan_confirmed is not True:
+            return []
+
+        self._resource_permissions.create_access_connectors_for_storage_accounts()
+
+        return []
```
JCZuurmond, line 264: `run_service_principal_migration` creates storage credentials as well. We need a tactic for avoiding permission overlap when adding an access connector for each storage account

src/databricks/labs/ucx/azure/access.py
```diff
@@ -21,16 +24,23 @@
 @dataclass
 class StoragePermissionMapping:
```
JCZuurmond, line 25: @nfx : The change to this data class blew up the PR.

1. I will separate the PR to make it easier to comprehend and review.
2. I would like your input on if we can circumvent this change. Specifically the `role_name`, which is required to track the storage permission, so that we can set the appropriate permission level for the access connectors. The `StoragePermissionMapping` contains the Databricks privilege, however, this does not map to the Azure storage permissions. -> We could neglect the "Storage Blob Data Owner", give "Storage Blob Data Contributor" as highest permission as the access connectors should not be used to do POSIX access control

JCZuurmond (commented):
Will be split into smaller PR, see comment

JCZuurmond:
First merge: #1417

PR #1381, opened 2024-04-12
Handled new output format of `SHOW TBLPROPERTIES` command

nkvuong:
## Changes
- `test_revert_migrated_table` was failing, due to new output of `SHOW TBLPROPERTIES` - previously, it is blank if a tbl prop is missing, now it shows `Table <table> does not have property: <prop>`

### Tests

- [x] manually tested


nfx (approved):
lgtm

PR #1392, opened 2024-04-12
Remove `StaticTablesCrawler` in favor of created database tracking 

nfx:
This PR aims at speeding up real integration testing jobs and making them less flaky.


tests/integration/conftest.py
```diff
@@ -336,28 +303,94 @@ def config(self) -> WorkspaceConfig:
             warehouse_id=self._env_or_skip("TEST_DEFAULT_WAREHOUSE_ID"),
             inventory_database=self.inventory_database,
             connect=self.workspace_client.config,
+            renamed_group_prefix=f'tmp-{self.inventory_database}-',
+            include_group_names=self.created_groups,
+            include_databases=self.created_databases,
+        )
+
+    def save_tables(self):
```
nfx, line 311: it's clear why do we need `save_mount`, but why do we need `save_tables`?..

PR #1393, opened 2024-04-12
A notebook linter to detect DBFS references within notebook cells

### LINKED ISSUE #1108: [FEATURE]: Detect usages of DBFS Mount paths in SQL Notebook cells

### Problem statement

..

### Proposed Solution

Build on top of #1105

### Additional Context

Related issues:
- https://github.com/databrickslabs/ucx/issues/1105

jimidle:
## Changes

Provide a new linter for Notebooks that inspects all SQL and Python cells for references
to DBFS mount points/folders. The linter raises Advisory or Deprecated when it discovers
such references in the code cells.

### Linked issues

Resolves #1108

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table
- [x] adds a new linter

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -29,9 +29,18 @@
...
+        self._original_offset: int = 0
         self._original_code = source
         self._migrated_code = source
 
+    @property
+    def original_offset(self) -> int:
+        return self._original_offset
+
+    @original_offset.setter
+    def original_offset(self, value: int):
```
nfx, line 41: Can we inject it via constructor instead? This would prevent a class of bugs related to incompletely initialized state

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -232,37 +241,49 @@ def new_cell(self, source: str) -> Cell:
...
             if cell_language is None:
                 cell_language = self
             else:
+                before = len(cell_lines)
                 self._remove_magic_wrapper(cell_lines, cell_language)
+                start += before - len(cell_lines)
+
             cell_source = '\n'.join(cell_lines)
-            return cell_language.new_cell(cell_source)
+            cell = cell_language.new_cell(cell_source)
```
nfx, line 266: Yep, and probably you'll have to change a signature of this method as well to support constructor injection

src/databricks/labs/ucx/source_code/notebook_linter.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+
+    def __init__(self, langs: Languages, notebook: Notebook):
+        self._languages: Languages = langs
+        self._notebook: Notebook = notebook
+        self._cell_offsets: list[int] = []
+
+    @classmethod
+    def from_notebook(cls, langs: Languages, notebook: Notebook) -> 'NotebookLinter':
+        return cls(langs, notebook)
```
nfx, line 22: Why do we need a factory method that just passes exactly the same arguments to constructor? It has no benefit. Remove.

src/databricks/labs/ucx/source_code/notebook_linter.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @classmethod
+    def from_source(
+        cls, langs: Languages, source: str, default_language: Language = languages.Language.SQL
```
nfx, line 26:
```suggestion
        cls, langs: Languages, source: str, default_language: Language
```
So default language has a semantic meaning: "primary language for this given source code, while allowing for use of other languages wrapped up in magic comments". Having a default value for this argument would lead to wrong assumptions

nfx (changes requested):
Nice PR. Few nits about control flow to be addressed

src/databricks/labs/ucx/source_code/notebook_linter.py
jimidle: Sure - I htink that at some point I expected to do some additional processing, but clearly that was not necessary.

src/databricks/labs/ucx/source_code/notebook_linter.py
jimidle: Got you

src/databricks/labs/ucx/source_code/notebook.py
jimidle: Done

src/databricks/labs/ucx/source_code/notebook.py
jimidle: Signature changed

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -226,60 +231,80 @@ def read_cell_language(self, lines: list[str]) -> CellLanguage | None:
...
         header = f"{self.comment_prefix} {NOTEBOOK_HEADER}"
         if not lines[0].startswith(header):
             raise ValueError("Not a Databricks notebook source!")
 
-        def make_cell(cell_lines: list[str]):
+        def make_cell(cell_lines: list[str], start: int):
             # trim leading blank lines
-            while len(cell_lines) > 0 and len(cell_lines[0]) == 0:
-                cell_lines.pop(0)
+            new_cell_lines = cell_lines
```
ganeshgirase, line 245: This is bug as you seem to be using Len of cell_lines down. Whatever changes you're doing to new_cell_lines also getting reflected in cell_lines as well.
For e.g,

Screenshot 2024-04-14 at 12 34 53 PM

Also, I believe you just need length(number of items) in cell lines, so get length preserved in some variable.

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -226,60 +231,80 @@ def read_cell_language(self, lines: list[str]) -> CellLanguage | None:
...
             cell_language = self.read_cell_language(cell_lines)
             if cell_language is None:
                 cell_language = self
             else:
-                self._remove_magic_wrapper(cell_lines, cell_language)
-            cell_source = '\n'.join(cell_lines)
-            return cell_language.new_cell(cell_source)
+                before = len(cell_lines)
+                new_cell_lines = self._remove_magic_wrapper(new_cell_lines, cell_language)
+                start += before - len(cell_lines)
```
ganeshgirase, line 259: This will lead to bug. Seems you want to write as ```start += before - len(new_cell_lines)``` otherwise before - len(cell_lines) will always lead to zero.

ganeshgirase (commented):
Some bugs in code.

CC:- @nfx

src/databricks/labs/ucx/source_code/notebook.py
nfx, line 245:
```suggestion
            new_cell_lines = cell_lines[:]
```
You wanted a copy, right?

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -226,60 +231,80 @@ def read_cell_language(self, lines: list[str]) -> CellLanguage | None:
...
+            new_cell_lines = cell_lines
+            while len(new_cell_lines) > 0 and len(new_cell_lines[0]) == 0:
+                new_cell_lines.pop(0)
+                start += 1
+
             # trim trailing blank lines
-            while len(cell_lines) > 0 and len(cell_lines[-1]) == 0:
-                cell_lines.pop(-1)
+            while len(new_cell_lines) > 0 and len(new_cell_lines[-1]) == 0:
+                new_cell_lines.pop(-1)
```
nfx, line 252: Shouldn't it affect some offset as well? Otherwise a cell in the middle of the notebook with 7 trailing blank lines would produce incorrect start_line results, resulting in corruption of migration.

Add a unit test for this

src/databricks/labs/ucx/source_code/notebook.py
nfx, line 259:
```suggestion
                start += before - len(new_cell_lines)
```
Otherwise this is noop (start += 0)

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -226,60 +231,80 @@ def read_cell_language(self, lines: list[str]) -> CellLanguage | None:
...
+
         if len(cell_lines) > 0:
-            cell = make_cell(cell_lines)
+            cell = make_cell(cell_lines, next_cell_pos)
             cells.append(cell)
 
         return cells
 
-    def _remove_magic_wrapper(self, lines: list[str], cell_language: CellLanguage):
+    def _process_line(self, line: str, prefix: str, lang_prefix: str, cell_language: CellLanguage) -> list[str]:
```
nfx, line 286:
```suggestion
    def _process_line(self, line: str, prefix: str, lang_prefix: str, cell_language: CellLanguage) -> str:
```
Indeed we always return only single string here

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -226,60 +231,80 @@ def read_cell_language(self, lines: list[str]) -> CellLanguage | None:
...
+        if line.startswith(prefix):
+            line = line[len(prefix) :]
+            if cell_language.requires_isolated_pi and line.startswith(lang_prefix):
+                new_line = line[len(lang_prefix) :]
+                return (
+                    [f"{cell_language.comment_prefix} {LANGUAGE_PI}", new_line.strip()]
+                    if new_line.strip()
```
nfx, line 293: Ternary operators in python are readable only if they fit on one line under 32 characters. This one is not readable. Rewrite as a normal if condition

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -226,60 +231,80 @@ def read_cell_language(self, lines: list[str]) -> CellLanguage | None:
...
+                    else [f"{cell_language.comment_prefix} {LANGUAGE_PI}"]
+                )
+            return [line]
+        if line.startswith(self.comment_prefix):
```
nfx, line 297: Can you move this condition up in the method body and do early return, so that we have less nesting?

src/databricks/labs/ucx/source_code/notebook_linter.py
```diff
@@ -11,15 +11,31 @@ class NotebookLinter:
...
+
+    def lint(self) -> Iterable[Advice]:
+        for cell in self._notebook.cells:
             if not self._languages.is_supported(cell.language.language):
                 continue
             linter = self._languages.linter(cell.language.language)
-            yield from linter.lint(cell.original_code)
+            for advice in linter.lint(cell.original_code):
+                yield advice
+                self._cell_offsets.append(cell.original_offset)
```
nfx, line 32: Shouldn't it be the first statement in the loop? Otherwise there's a bug, where offsets are counted only for the bad cells

src/databricks/labs/ucx/source_code/notebook_linter.py
```diff
@@ -11,15 +11,31 @@ class NotebookLinter:
...
+
+    def adjust_advices(self, advices: list[Advice]) -> Iterable[Advice]:
+        for advice, offset in zip(advices, self._cell_offsets):
+            advice = advice.replace(start_line=advice.start_line + offset, end_line=advice.end_line + offset)
```
nfx, line 36: Do we even need this method? Why can't we replace line offsets before yielding?

tests/unit/source_code/test_notebook.py
```diff
@@ -82,8 +83,10 @@ def test_notebook_rebuilds_same_code(source: tuple[str, Language, list[str]]):
     assert notebook is not None
     new_source = notebook.to_migrated_code()
     # ignore trailing whitespaces
-    actual_purified = new_source.replace(' \n', '\n')
-    expected_purified = sources[0].replace(' \n', '\n')
+    actual_purified = re.sub(r'\s+$', '', new_source, flags=re.MULTILINE)
+    expected_purified = re.sub(r'\s+$', '', sources[0], flags=re.MULTILINE)
+    if actual_purified != expected_purified:
+        print("what have you done jim?")
```
nfx, line 89: What is this?

tests/unit/source_code/test_notebook_linter.py
```diff
@@ -1,39 +1,174 @@
...
-from databricks.labs.ucx.source_code.notebook import Notebook
 from databricks.labs.ucx.source_code.notebook_linter import NotebookLinter
 
+index = MigrationIndex([])
 
-def test_notebook_linter():
-    notebook_code = """-- Databricks notebook source
+
+@pytest.mark.parametrize(
+    "lang, source, expected, adjusted",
```
nfx, line 13:
```suggestion
    "lang, source, expected",
```
Yield already adjusted advice

nfx (changes requested):
This PR lacks unit tests with different leading and trailing whitespace in the middle of the notebook. There's high probability of source corruption due to noted issues in offset adjustment

nfx:
Also, we prefer git rebase to merge commits. This branch has irrelevant commit history now, for example

src/databricks/labs/ucx/source_code/notebook.py
```diff
@@ -226,60 +231,80 @@ def read_cell_language(self, lines: list[str]) -> CellLanguage | None:
...
+            return [f"{cell_language.comment_prefix} {COMMENT_PI}{line}"]
+        return [line]
```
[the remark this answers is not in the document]
jimidle: Sorry - this should have been a WIP to fix bugs in notebook processing. I split this out to reduce indents and did not return to the code yet.

src/databricks/labs/ucx/source_code/notebook.py
jimidle: This is indeed a bug introduced by refactor.

jimidle: I assumed that leading and trailing blank lines were being deleted for a reason. The new_cell lines thing was a bug left over from refactor, caused by needing to fix a bug in the notebook parser that indeed corrupted the source:

```
# MAGIC %sql  SELECT * FROM 
```

src/databricks/labs/ucx/source_code/notebook.py
jimidle: YEs - just a hangover from a refactor. FIxed now.

src/databricks/labs/ucx/source_code/notebook.py
jimidle: I will do that

src/databricks/labs/ucx/source_code/notebook.py
jimidle: I ocudl, but there are two conditionals and a default of neither of them passes. I could move the tests out of here perhaps. I will look. This was code I inherited of course.

src/databricks/labs/ucx/source_code/notebook_linter.py
jimidle: I only count them for bad cells. It is the only way to correlate the offsets wihtout storing the cell number in the advice. HOwever, we have now changed the requirement to just report the notebook source line anyway, so this will now become irrelevant

src/databricks/labs/ucx/source_code/notebook_linter.py
jimidle: You originally asked for reporting at cell offsets, then a layer to do the translation. If we do not need the original cell offsets, then I can remove this logic and just do it when the advice is first raised.

tests/unit/source_code/test_notebook.py
jimidle: Accident while debugging. Apologies - I was tired by then.

tests/unit/source_code/test_notebook_linter.py
jimidle: OK - originally I was asked to do this. It is a lot easier to yiled with adjustements done before

jimidle: Changed to yield only adjusted advisory

src/databricks/labs/ucx/source_code/notebook_linter.py
jimidle: Changed to only yield adjusted offsets

src/databricks/labs/ucx/source_code/notebook_linter.py
jimidle: No longer relevant

src/databricks/labs/ucx/source_code/notebook.py
jimidle: IN fact the existing tests all have trailing blank lines in cells and so this is already tested

src/databricks/labs/ucx/source_code/notebook.py
jimidle: I iwll change some of the tests to change leading and trailing blanks lines form other than one

src/databricks/labs/ucx/source_code/notebook.py
jimidle: This is no longer necessary.

src/databricks/labs/ucx/source_code/notebook.py
jimidle: All fixed

nfx (approved):
lgtm

PR #1395, opened 2024-04-13
Modifies dependency resolution logic to detect deprecated use of s3fs package

### LINKED ISSUE #443: feature: handle `s3fs` library, `s3fs.S3FileSystem` calls
One pattern I've seen is users using s3fs to access files on S3 buckets.
The the assessment tools, job migration tools, notebook migration tools should flag on this library.
Recommended upgrade strategy is to migrate to UC Volumes.

# Proposed solution

yield Deprecation for all `s3fs` imports

jimidle:
## Changes
The use of the s3fs import package is deprecated. A new check is implemented within the dependency walker to raise Deprecation warnings if imports of the s3fs package are detected in Python code.

### Linked issues

Resolves #443

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests


tests/unit/source_code/test_s3fs.py
```diff
@@ -0,0 +1,117 @@
+import pytest
+
+from databricks.labs.ucx.source_code.base import Advice, Deprecation
+from databricks.labs.ucx.source_code.s3fs import S3FSUsageLinter
+
+
+class TestDetectS3FS:
```
nfx, line 7: Our convention for pytest is NOT to group tests into classes, so that we keep the nesting to minimum. I don't see a valid reason to break this convention in this particular file.

src/databricks/labs/ucx/source_code/s3fs.py
```diff
@@ -0,0 +1,62 @@
...
+from databricks.labs.ucx.source_code.base import Advice, Linter, Deprecation
+
+
+class DetectS3fsVisitor(ast.NodeVisitor):
+    def __init__(self):
+        self._advices: list[Advice] = []
+
+    def visit_Import(self, node):  # pylint: disable=invalid-name
+        for alias in node.names:
+            if alias.name == 's3fs':
```
nfx, line 13: We have DependencyGraph object with dependencies method, that returns file/cell dependencies, including imports. 

Rewrite this to fit into that framework. There'll be more incompatible packages/subpackages we'll have to flag.

The good part is that you already have unit tests for this.

nfx (changes requested):
Rewrite to use dependency resolution framework : https://github.com/databrickslabs/ucx/commit/18f05b3025b99fc0965ab1c2dfdf3c66aa82c38a

nfx:
@jimidle See this commit, to be precise

src/databricks/labs/ucx/source_code/s3fs.py
jimidle: No problem.

tests/unit/source_code/test_s3fs.py
jimidle: class removed.

src/databricks/labs/ucx/source_code/s3fs.py
jimidle: I will do this in two separate commits. One to use the whitelist to detect s3fs and raise Advice, One will modify the processing so we can identify where the the import came for as this information is currently lost

src/databricks/labs/ucx/source_code/dependencies.py
```diff
@@ -104,16 +105,31 @@ def _load_source(self, object_info: ObjectInfo) -> str:
...
             return dependency
         compatibility = self._whitelist.compatibility(dependency.path)
         # TODO attach compatibility to dependency, see https://github.com/databrickslabs/ucx/issues/1382
         if compatibility is not None:
+            if compatibility == UCCompatibility.NONE:
+                self._advices.append(
+                    Deprecation(
+                        code="dependency-check",
+                        message=f"Use of dependency {dependency.path} is deprecated",
+                        start_line=0,
```
nfx, line 121: How can we modify it to highlight the exact code range where deprecated import happened?...

tests/unit/__init__.py
```diff
@@ -130,6 +131,23 @@ def _secret_not_found(secret_scope, _):
     msg = f"Secret Scope {secret_scope} does not exist!"
     raise NotFound(msg)
 
+    # can't remove **kwargs because it receives format=xxx
```
nfx, line 134: This comment seems to have wrong indentation

tests/unit/source_code/samples/s3fs-python-compatibility-catalog.yml
```diff
@@ -0,0 +1,29 @@
...
+      - some failure 1
+  - name: databricks.sdk.service.compute
+    compatibility: full
+    failures:
+      - some failure 1
+---
+identifier:
+  name: s3fs
+  # all versions are incompatible
+top_level: s3fs
```
nfx, line 20: This is a test resource, how this will be read by prod?..

tests/unit/source_code/test_s3fs.py
```diff
@@ -0,0 +1,152 @@
...
+            ],
+        ),
+        ("from leeds import path, s3fs", []),
+        (
+            "def func():\n    import s3fs",
+            [
+                Deprecation(
+                    code='dependency-check',
+                    message=S3FS_DEPRECATION_MESSAGE,
+                    start_line=0,
```
nfx, line 68: No start line information is bad :(

nfx (changes requested):

src/databricks/labs/ucx/source_code/dependencies.py
jimidle: Two things here. I think that it would be better to not report during graph building, but do a walk after. But add more information based on object type. So file, token etc. hence I wanted two phases to this. I talked briefly with @ericvergnaud about that.

tests/unit/__init__.py
jimidle: Ah yes.

tests/unit/source_code/samples/s3fs-python-compatibility-catalog.yml
jimidle: I can add it to pre-known packages. First the test I just copied existing test config.

tests/unit/source_code/test_s3fs.py
jimidle: Yes. Hence two phases in implementation. The graph/tree needs additional information. Possibly best to add parent to dependency but that's why I probably want to change to a pass after dependency graph is built. Tokens are not current preserved, but will be.

tests/unit/__init__.py
jimidle: Fixed.

tests/unit/source_code/test_s3fs.py
ericvergnaud: @nfx @jimidle I agree that it would be much simpler to proceed in at least 2 passes: a) build the dependency graph b) run the linters (might even need an intermediate one for dependency resolution)

src/databricks/labs/ucx/source_code/dependencies.py
jimidle: Exactly. But that info is not available right now. When the dependency is constructed, I need to store the origin information in the ObjectInfo. Then the Resolver can use it. Probably with another traversal.

tests/unit/source_code/test_s3fs.py
jimidle: @ericvergnaud Yes - that is what I intend. It all kind of happens at once at the moment, and as we said, I will need to make changes for tracking so we know how we end up at a particular node in the graph.

PR #1431, opened 2024-04-17
Propagate source location information within the import package dependency graph

### LINKED ISSUE #1418: Propagate file and start/end line/character information for incompatible dependencies
Modify to highlight the exact code range where deprecated import happened

_Originally posted by @nfx in https://github.com/databrickslabs/ucx/pull/1395#discussion_r1566448143_

jimidle:
## Changes

In order to report the location of imports that are either deprecated or will be migrated automatically, the location of the origin source code needs to be preserved. The graph of package dependencies now propagates location information such that reporting can indicate the hierarchy of dependencies that led to discovery of the deprecated import, or one that will be automatically migrated.

### Linked issues

Resolves #1418 

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table
- [x] modifies dependency graph build logic 

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/source_code/base.py
+++ b/src/databricks/labs/ucx/source_code/base.py
@@ -52,6 +52,9 @@ class Advice:
     def as_convention(self) -> 'Convention':
         return Convention(**self.__dict__)
 
+    def as_location(self) -> 'Location':
+        return Location(**self.__dict__)
+
 
 class Advisory(Advice):
     """A warning that does not prevent the code from running."""
@@ -67,6 +70,10 @@ class Deprecation(Advisory):
 
 class Convention(Advice):
     """A suggestion for a better way to write the code."""
+
+
+class Location(Advice):
+    """A location in the code, which is not yet an advice"""
 
 
 class Linter:
--- a/src/databricks/labs/ucx/source_code/dependencies.py
+++ b/src/databricks/labs/ucx/source_code/dependencies.py
@@ -1,13 +1,24 @@
 from __future__ import annotations
 
-import abc
+from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable
 
 from databricks.sdk.service.workspace import ObjectType, ObjectInfo, ExportFormat, Language
 from databricks.sdk import WorkspaceClient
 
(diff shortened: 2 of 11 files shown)


ericvergnaud:
@jimidle, @nfx  this collides significantly with PR #1440

jimidle:
> @jimidle, @nfx this PR collides significantly with PRs #1440 and #1447

I had a feeling it might when I heard what you were working on today. Perhaps it was unwise for us both to work on this. I guess one PR goes in first and the other gets reworked after those changes are in. Did you also solve the infinite visit and loss of dependencies that have a known status? It may be better for me to reapply this after your changes are applied?

ericvergnaud:
in my experience _the infinite visit_ is a side effect of using colliding names in testing i.e. if you name 'path' a sample that imports 'path', then you're in trouble (a python module cannot directly import itself) ?
not sure what you mean by _loss of dependencies that have a known status_ ?
Indeed I suspect it might be better to revisit this after https://github.com/databrickslabs/ucx/pull/1440 and https://github.com/databrickslabs/ucx/pull/1447 are merged (and also maybe #1448)

jimidle:
> in my experience _the infinite visit_ is a side effect of using colliding names in testing i.e. if you name 'path' a sample that imports 'path', then you're in trouble (a python module cannot directly import itself) ? not sure what you mean by _loss of dependencies that have a known status_ ? Indeed I suspect it might be better to revisit this after #1440 and #1447 are merged (and also maybe #1448)

In the current resolver, if it finds that a package is known, it does not return it, so it is not added to the dependency graph. My change returns it, adds it as a dependency, but then does not build its child dependencies.

The infinite walk was caused by the dependencies property adding itself to the list of dependencies. I did not see any code that needed to know about itself other than tests looking for paths. But of course we could have dependencies and child_dependencies as properties, and they can do different things. 

If @nfx wants us to do it in this order then we should bring in your PRs and then I can see what changes you have made and re-apply this PR.

nfx (approved):
okay for an increment

ericvergnaud:
@nfx the proposal was to do it the other way round
Can you roll this back ?

Issue #1124, opened 2024-03-26
[FEATURE]: cross-dbr migrations


nfx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

There are breaking changes sometimes

### Proposed Solution

Lint for breaking changes:
1. https://docs.databricks.com/en/release-notes/runtime/14.3lts.html#spark-catalog-apis-now-fully-supported-in-shared-access-mode
2. https://docs.databricks.com/en/archive/runtime-release-notes/14.0.html#default-current-working-directory-has-changed
3. https://docs.databricks.com/en/archive/runtime-release-notes/14.0.html#introducing-spark-connect-in-shared-cluster-architecture
4. https://docs.databricks.com/en/archive/runtime-release-notes/13.1.html#remove-start-version
5. https://docs.databricks.com/en/archive/runtime-release-notes/13.1.html#parquet-failonunknownfields-no-longer-silently-drop-data-on-type-mismatch
6. https://docs.databricks.com/en/archive/runtime-release-notes/13.0.html#pip-does-not-automatically-restart-python
7. https://docs.databricks.com/en/release-notes/runtime/13.3lts.html#block-schema-overwrite-when-using-dynamic-partition-overwrites
8. https://docs.databricks.com/en/archive/runtime-release-notes/13.2.html#unity-catalog-support-for-python-and-pandas-user-defined-functions-udfs
9. https://docs.databricks.com/en/release-notes/runtime/12.2lts.html#behavior-changes-with-the-new-lateral-column-alias-feature
10. https://docs.databricks.com/en/release-notes/runtime/12.2lts.html#implicit-lateral-column-aliasing-support
11. https://docs.databricks.com/en/release-notes/runtime/11.3lts.html#structured-streaming-trigger-once-is-deprecated
12. https://docs.databricks.com/en/release-notes/runtime/11.3lts.html#applyinpandaswithstate-throws-an-error-if-the-query-has-a-shuffle-after-the-operator
13. https://docs.databricks.com/en/release-notes/runtime/11.3lts.html#hadoop-cache-for-s3a-is-now-disabled
14. https://docs.databricks.com/en/archive/runtime-release-notes/11.0.html#https-scheme-now-enforced-when-abfs-client-uses-sas-token
15. https://docs.databricks.com/en/archive/runtime-release-notes/11.0.html#log4j-is-upgraded-from-log4j-1-to-log4j-2
[...]

PR #1435, opened 2024-04-17
Added Python linter for table creation with implicit format.

dumitrac:
## Changes
Starting with DBR v8.0, the default table format changed from 'parquet' to 'delta'.
This linter yields advice for situations where a 'writeTo', 'table', 'insertInto', or 'saveAsTable' method is invoked, and no '.format(...)' invocation is found in the same chain of calls.
For 'saveAsTable' in particular, the alternative of passing the 'format' as a direct argument is also supported.

Only linting is supported for this, no fixing.
If a DRM version is passed to the linter, it will skip linting when this version is >= 8.0. This is because the format default change from "parquet" to "delta" starting with v8.0.

Examples:
```
df.writeStream.table("c.db.table") # => linter yields Advice due to implicit format
df.writeStream.format("parquet").table("c.db.table") # => OK!
df.writeStream.saveToTable("c.db.table", format="parquet") # => OK!
```

The linter is implemented in the new file table_creation.py, and makes use of a few reusable AST utilities in the new file python_ast_util.py.
Unit-tests are added for both source files.

### Linked issues
Relates to  #1124 , specifically item 26 - https://docs.databricks.com/en/archive/runtime-release-notes/8.0.html#delta-is-now-the-default-format-when-a-format-is-not-specified

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [x] modified existing workflow: code migration
- [ ] added a new table

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests


tests/unit/source_code/test_python_ast_util.py
```diff
@@ -0,0 +1,100 @@
...
+
+@pytest.mark.parametrize(
+    "param",
+    [
+        {"stmt": "o.m1()", "expected": 0},
+        {"stmt": "o.m1(3)", "expected": 1},
+        {"stmt": "o.m1(first=3)", "expected": 1},
+        {"stmt": "o.m1(3, 3)", "expected": 2},
+        {"stmt": "o.m1(first=3, second=3)", "expected": 2},
+        {"stmt": "o.m1(3, second=3)", "expected": 2},
```
JCZuurmond, line 79: Maybe you could test for the python `*` wildcard

```suggestion
        {"stmt": "o.m1(3, *, second=3)", "expected": 2},
```

JCZuurmond (approved):
LGTM, some questions

tests/unit/source_code/test_python_ast_util.py
```diff
@@ -0,0 +1,100 @@
...
+        {"stmt": "o.m1(second=3, first=4)", "arg_index": None, "arg_name": "second", "expected": 3},
+        {"stmt": "o.m1(second=3)", "arg_index": 1, "arg_name": "second", "expected": 3},
+        {"stmt": "o.m1(4, 3, 2)", "arg_index": 1, "arg_name": "second", "expected": 3},
+    ],
+)
+def test_get_arg(migration_index, param):
+    call = get_statement_node(param["stmt"]).value
+    act = AstUtil.get_arg(call, param["arg_index"], param["arg_name"])
+    if param["expected"] is None:
+        assert param["expected"] is None
```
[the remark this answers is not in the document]
dumitrac: Good catch - last line should've been
`assert act is None`
Fixed. Thanks.

src/databricks/labs/ucx/source_code/table_creation.py
```diff
@@ -0,0 +1,107 @@
...
+        return Span(
+            call.lineno,
+            call.col_offset,
+            call.end_lineno or 0,
+            call.end_col_offset or 0,
+        )
+
+
+@dataclass
+class NoFormatPythonLinter:
```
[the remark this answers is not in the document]
dumitrac: removed @dataclass.

src/databricks/labs/ucx/source_code/python_ast_util.py
```diff
@@ -0,0 +1,68 @@
...
+    def extract_callchain(node: ast.AST) -> ast.Call | None:
+        """If 'node' is an assignment or expression, extract its full call-chain (if it has one)"""
+        call = None
+        if isinstance(node, ast.Assign):
+            call = node.value
+        elif isinstance(node, ast.Expr):
+            call = node.value
+        if not isinstance(call, ast.Call):
+            call = None
+        return call
```
nfx, line 28:
```suggestion
        if isinstance(node, ast.Assign):
            return node.value
        if isinstance(node, ast.Expr):
            return node.value
        return None
```
init with `None` is often source of bugs. always return early

src/databricks/labs/ucx/source_code/python_ast_util.py
```diff
@@ -0,0 +1,68 @@
+from __future__ import annotations
+
+import ast
+from dataclasses import dataclass
+
+
+@dataclass
+class Span:
```
nfx, line 8: LSP calls this datasctructure as `Range`.

src/databricks/labs/ucx/source_code/python_ast_util.py
```diff
@@ -0,0 +1,68 @@
...
+    """Represents a (possibly multiline) source code span."""
+
+    start_line: int
+    start_col: int
+    end_line: int
+    end_col: int
+
+
+class AstUtil:
```
nfx, line 17: can we move these methods to `ASTLinter`?.. 

this whole file largely overlaps with `python_linter.py`

src/databricks/labs/ucx/source_code/python_ast_util.py
```diff
@@ -0,0 +1,68 @@
...
+    @staticmethod
+    def extract_callchain(node: ast.AST) -> ast.Call | None:
```
nfx, line 19: isn't it the same as `databricks.labs.ucx.source_code.python_linter.ASTLinter.locate`?

src/databricks/labs/ucx/source_code/python_ast_util.py
```diff
@@ -0,0 +1,68 @@
...
+        if isinstance(node, ast.Assign):
+            call = node.value
+        elif isinstance(node, ast.Expr):
+            call = node.value
+        if not isinstance(call, ast.Call):
+            call = None
+        return call
+
+    @staticmethod
+    def extract_call_by_name(node: ast.Call, name: str) -> ast.Call | None:
```
nfx, line 31: can we reuse part of `databricks.labs.ucx.source_code.python_linter.ASTLinter.locate`? or at least place these methods there?

nfx (changes requested):
please see if you can reuse most of what's in `python_linter.py`

src/databricks/labs/ucx/source_code/python_ast_util.py
dumitrac: Ok, I tried to use it but it triggers a cyclic import, because lsp imports languages which imports my linter.

I think these foundational utilities should be importing much (if anything) from the codebase, instead they'd be at the bottom of the import chain (everything depends on them, they depend on nothing).

src/databricks/labs/ucx/source_code/python_ast_util.py
dumitrac: It's possible but very hard to tell, ASTLinter is undocumented and rather complex.
I'm also a bit worried about the over-use of class inheritance - we could easily do composition or simple static calls for utilities such as these.

src/databricks/labs/ucx/source_code/python_ast_util.py
dumitrac: It doesn't seem to work the same way - locate wants to be given the full form of the callchain, whereas we don't know that form here. We just know there is somewhere in the callchain a call to "saveToTable", but not what comes before it (could be anything).

src/databricks/labs/ucx/source_code/python_ast_util.py
dumitrac: It doesn't seem to be functionally equivalent, is it?
I can split it in two methods instead.

src/databricks/labs/ucx/source_code/python_ast_util.py
dumitrac: I moved everything over under ASTLinter in the new revision.

src/databricks/labs/ucx/source_code/python_ast_util.py
dumitrac: Please see my response above - it doesn't seem to do the same thing as far as I can tell.

PR #1448, opened 2024-04-18
Raise advice with missing or deprecated dependency

### LINKED ISSUE #1439: [FEATURE]: Create Advice alerts when failing to build a dependency graph

### Problem statement

When building a dependency graph, we may encounter situations where a dependency cannot be found. Rather than fail the entire discovery process we must raise appropriate Advices

### Proposed Solution

specify and raise Advices

### Additional Context

sub-ticket of #1202

ericvergnaud:
## Changes
Raise advice with missing or deprecated dependency

### Linked issues
#1202 

Resolves #1431 
Resolves #1439

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/source_code/base.py
```diff
@@ -15,8 +15,14 @@
 
 @dataclass
 class Advice:
+
+    MISSING_SOURCE_TYPE = "<MISSING_SOURCE_TYPE>"
+    MISSING_SOURCE_PATH = "<MISSING_SOURCE_PATH>"
+
     code: str
     message: str
+    source_type: str
```
nfx, line 24: we should keep WHAT (e.g. deprecation) problem we have from WHERE (e.g. notebook or a file) that problem is. 

```
@dataclass
class NotebookAdvice:
  path: str
  advice: list[Advice]

@dataclass
class FileAdvice:
  path: str
  advice: list[Advice]
```

that being said, we may need another datastructure to map onto a database table, but it should have Job ID and Job Name as well and most likely live within the `assessment` package:

```
@dataclass
class WorkloadAdvice:
  workflow_id: int
  workflow_name: str
  object_type: string
  object_path: string # alternatively, object_id, to make links easier
  code: str
  message: str
  line: int
```

nfx (changes requested):

src/databricks/labs/ucx/source_code/base.py
ericvergnaud: I'm afraid this proposal works for linting/fixing but not when building the dependency graph. 
In that scenario, there's more than one source container involved, and when we detect an issue (missing dependency, deprecation...) we don't know (and don't want to know) which one. We just want to yell: hey this dependency is deprecated!
Depending on how we plan to use the advices, we might indeed want to aggregate them the way you propose, but that would happen after (not during) collection.

nfx: then yield `DependencyProblem` and convert it to `Advice` in the `NotebookLinter`:

```
@dataclass
class DependencyProblem:
  path: Path
  node: ast.Node
  code: str # message code, not source code
  message: str
```

We cannot change the interfaces defined in this file, because it's not forward-compatible with LSP, which means that we can't allow for `class Advice` to get more fields.

nfx (changes requested):

src/databricks/labs/ucx/source_code/base.py
ericvergnaud: Ah, I wasn't aware of the LSP constraint. I like the `DependencyProblem` idea. Not sure `node` will make it though (no such node with `%run` cells), I guess location fields similar to Advice ones would be ok ?

nfx: > (no such node with %run cells)

there's a range for %run cells for sure

ericvergnaud: Yes it's already implemented in `_collect_advice` in  https://github.com/databrickslabs/ucx/pull/1448/files#diff-5c020cd42bbfe34d095f2440a98891b416f6097a94537838bfecbd98ce771ac3

ericvergnaud:
superseded by #1529

PR #1434, opened 2024-04-17
Migrate non delta dbfs tables using Create Table As Select (CTAS). Convert such tables to Delta tables.

FastLee:
Migrate Non-Delta DBFS root tables to Managed Delta tables.
closes #1340
This release enhances migration for non-delta DBFS tables, supporting more table types and configurations. New methods improve CTAS functionality with safer SQL statement generation. Additional table format compatibility, new creation methods, and updated integration tests are included. Grant assignments during migration are now supported.

PR #1454, opened 2024-04-19
Fix `test_describe_all_tables_in_databases`

nkvuong:
## Changes
- `make_table` now creates external tables that are identified as `DBFS_ROOT_NON_DELTA`. This was a change introduced in #1434

### Tests

- [x] manually tested


nfx (approved):
lgtm

PR #1464, opened 2024-04-19
Extended wait time for group checking during tests

nkvuong:
## Changes
Due to eventual consistency, wait up to 2 minutes before failed tests involved group APIs

### Tests

- [x] manually tested


nfx (approved):
two minutes so two minutes.

PR #1465, opened 2024-04-19
Adding seen tables feature

william-conti:
## Changes
Adding a filter to remove tables in mounts that have been identified by the assessment.

As the assessment job will identify tables in hive metastore, these tables can be considered as tables in mounts later by the scan tables in mounts job.
We should filter those tables as it will yields duplicates during the mapping and migration phase. Plus UCX yields better support to migrate tables that are registered in HMS.


JCZuurmond (approved):
LGTM

nfx (approved):
Lgtm

PR #1502, opened 2024-04-23
Added check for circular view dependency

### LINKED ISSUE #1158: [FEATURE]: Check circular dependencies when migrating views

### Problem statement

Although theoretically made impossible by the database server, there exists a scenario where views could refer to each other:
```
CREATE VIEW V1 AS SELECT * FROM V2
CREATE VIEW V2 AS SELECT * FROM V1
```
This could happen if the DDL is extracted at different points in time and consolidated outside the database.
This would lead to an infinite loop in the ViewMigrator, and an impossibility to migrate altogether.
It is therefore useful to check for this scenario.

### Proposed Solution

Enhance `src/databricks/labs/ucx/hive_metastore/views_migrator.py` such that it checks for circular dependencies and eventually raises an error 

### Additional Context

_No response_

jincejames:
## Changes
Added check for circular view dependency

### Linked issues
fixes #1158 

### Tests
Existing test for circular view dependency is satisfied, and a new test to check chained circular dependency (A->B, B->C, C->A) has been added.

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] verified on staging environment (screenshot attached)**

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/view_migrate.py
+++ b/src/databricks/labs/ucx/hive_metastore/view_migrate.py
@@ -41,18 +41,6 @@ class ViewToMigrate(TableToMigrate):
                 continue
             yield TableView("hive_metastore", src_db, old_table.name)
 
-    def check_circular_dependency(self, initial_view):
-        queue = []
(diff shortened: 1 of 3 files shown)


nfx (approved):
LGTM

PR #1511, opened 2024-04-24
Add CLI commands to trigger table migration workflow

### LINKED ISSUE #670: Add `databricks labs ucx migrate-tables` and a related workflow
databricks labs ucx migrate-tables to apply table mapping from CSV file with the following columns: src_workspace_name,dst_catalog,src_schema,dst_schema,src_table,dst_table . Consider the following stages:

**Initial assessment**
We scan all tables, generate an inventory, get all perms, etc. 
We provide a mapping file the customer can customize, but seed the UC table name with something like the workspace name or ID as the catalog.
Customer can choose to modify the mapping or not

**Initial migration**
We use the mapping file to sync tables, apply perms, and persist the mapping data in inventory

**Repeated migration** - this is probably a job that runs on schedule and the user should be instructed to configure accordingly
We use the mapping data in inventory to re-sync any tables, but not re-apply permissions as these probably changed.

Fine-grained:
- https://github.com/databrickslabs/ucx/issues/333

Upstream dependencies:
- https://github.com/databrickslabs/ucx/issues/341
- https://github.com/databrickslabs/ucx/issues/94
- https://github.com/databrickslabs/ucx/issues/669

Next quarter:
- https://github.com/databrickslabs/ucx/issues/334
- https://github.com/databrickslabs/ucx/issues/332
- https://github.com/databrickslabs/ucx/issues/319
- https://github.com/databrickslabs/ucx/issues/673

qziyuan:
## Changes
Add two CLI commands to trigger table migration workflows

### Linked issues
closes #670

### DIFF
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -400,19 +400,13 @@ def assign_metastore(
 
 
 @ucx.command
-def migrate_tables(w: WorkspaceClient):
+def migrate_tables(w: WorkspaceClient, *, ctx: WorkspaceContext | None = None):
     """Trigger the migrate-tables workflow"""
-    ctx = WorkspaceContext(w)
-    deployed_workflows = ctx.deployed_workflows
-    deployed_workflows.run_workflow("migrate-tables")
-
-
-@ucx.command
-def migrate_external_hiveserde_tables_in_place(w: WorkspaceClient, *, ctx: WorkspaceContext | None = None):
-    """Trigger the migrate-external-hiveserde-tables-in-place-experimental workflow"""
     if ctx is None:
         ctx = WorkspaceContext(w)
     deployed_workflows = ctx.deployed_workflows
+    deployed_workflows.run_workflow("migrate-tables")
+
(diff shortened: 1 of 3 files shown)


src/databricks/labs/ucx/cli.py
```diff
@@ -398,5 +398,21 @@ def assign_metastore(
...
+@ucx.command
+def migrate_tables(w: WorkspaceClient):
+    """Trigger the migrate-tables workflow"""
+    ctx = WorkspaceContext(w)
+    deployed_workflows = ctx.deployed_workflows
+    deployed_workflows.run_workflow("migrate-tables")
+
+
+@ucx.command
+def migrate_external_hiveserde_tables_in_place(w: WorkspaceClient):
```
nfx, line 410: Can we merge this into one command and prompt user to trigger hiveserde job only if there are such tables where we show some numbers? Also, how should we deal with CTAS? We have to prompt user for a choice.

nfx (changes requested):
Can we merge this into one command and prompt users?

nfx (approved):
Thank you

PR #1568, opened 2024-04-26
Remove hidden bugs from unit tests by enforcing mocks

nfx:
Enforce check defined in:
- https://github.com/databrickslabs/pylint-plugin/pull/44

### DIFF
--- a/tests/unit/aws/test_access.py
+++ b/tests/unit/aws/test_access.py
@@ -98,16 +98,11 @@ def backend():
 
 
 @pytest.fixture
-def mock_aws():
-    return create_autospec(AWSResources)
-
-
-@pytest.fixture
 def locations(mock_ws, backend):
     return ExternalLocations(mock_ws, backend, "ucx")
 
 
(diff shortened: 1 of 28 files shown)


tests/unit/aws/test_access.py
```diff
@@ -130,9 +126,10 @@ def test_create_external_locations(mock_ws, installation_multiple_roles, mock_aw
         call(mock.ANY, 's3://BUCKETX/FOLDERX', 'credx', skip_validation=True),
     ]
     mock_ws.external_locations.create.assert_has_calls(calls, any_order=True)
+    aws.get_role_policy.assert_not_called()
```
nkvuong, line 129: this should still work, without re-adding `aws = create_autospec(AWSResources)`
```suggestion
    mock_aws.get_role_policy.assert_not_called()
```

nfx: `mock_aws` was a bit difficult to read

PR #1576, opened 2024-04-28
Changes to identify service principal with custom roles on Azure storage account for principal-prefix-access

### LINKED ISSUE #886: [FEATURE]: Identity Service Principal with CustomRole access on storage account `databricks labs ucx save_azure_storage_accounts`

### Problem statement

save_azure_storage_accounts cmd currently looks for SPN with Built in blob roles (Blob Owner, Blob, Contributor and Blog Reader), but doesn't cover Custom Roles with read access on the storage used by clusters. Extend the save_azure_storage_accounts to include custom roles as well

### Proposed Solution

to be defined

### Additional Context

_No response_

mohanab-db:
## Changes
1. Changes to identify service principal with custom roles on Azure storage account
2. Unit tests added to increase overall coverage

### Linked issues
Resolves #886 

### Functionality 
- [ ] modified existing command: `databricks labs ucx principal-prefix-access`

### Tests
- [ ] manually tested
- [ ] added unit tests


JCZuurmond (commented):
Hi @mohanab-db, could you please look at my comments?

tests/unit/workspace_access/test_scim.py
```diff
@@ -204,6 +204,113 @@ def test_get_apply_task_should_ignore_groups_not_in_migration_state():
     assert sup.get_apply_task(item, MigrationState([mggrp])) is None
 
 
+def test_get_apply_task_target_group_not_found():
```
nkvuong, line 207: @mohanab-db can you split the additional unit tests into a separate PR? makes it easier to review & revert if necessary

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -435,27 +437,52 @@ def _role_assignment(
...
+    def _role_name(self, role_definition_id) -> dict[str, str | list[str] | None]:
         if role_definition_id not in self._role_definitions:
             role_definition = self._mgmt.get(role_definition_id, "2022-04-01")
             definition_properties = role_definition.get("properties", {})
             role_name: str = definition_properties.get("roleName")
             if not role_name:
-                return None
-            self._role_definitions[role_definition_id] = role_name
-        return self._role_definitions.get(role_definition_id)
+                return {
```
nkvuong, line 466: should make this a dataclass instead of a dict

src/databricks/labs/ucx/azure/access.py
```diff
@@ -54,6 +56,43 @@ def __init__(
...
+                elif privileges and privileges == Privilege.WRITE_FILES:
+                    higher_privilege = Privilege.WRITE_FILES
+                    break
+        return higher_privilege
+
+    def _get_role_privilege(self, role_assignment: AzureRoleAssignment) -> Privilege | None:
+        privilege = None
+        # Check for custom role permissions on the storage accounts
+        if role_assignment.role_permissions:
+            privilege = self._get_custom_role_privilege(role_assignment.role_permissions)
```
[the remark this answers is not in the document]
mohanab-db: Hi @JCZuurmond Thanks for reviewing the code. I am working on the comments, once done I will create a new PR and share the details with you. 
For this one, I have a question. How the existing functionality for built-in roles work is, if you have 2 role assignments (Blob Reader built-in role and Blob Contributor built-in role) to a service principal (sp1), then the output will be 2 rows -- one row for sp1 read files, one role for sp1 write files.
Similarly, with the new changes, if I have 2 role assignments -- one Blob Contributor built-in role, one read access through custom role, then the output will be 2 rows -- one row for sp1 write files, one role for sp1 read files.
Please let me know if I should handle both the scenarios (existing and new) to reflect only the higher privilege.

tests/unit/workspace_access/test_scim.py
mohanab-db: @nkvuong Thanks for reviewing. I have removed additional test from this PR and created another PR just with the unit tests - https://github.com/databrickslabs/ucx/pull/1586

src/databricks/labs/ucx/azure/resources.py
mohanab-db: @nkvuong - Done. Please check.

src/databricks/labs/ucx/azure/access.py
```diff
@@ -54,6 +56,43 @@ def __init__(
...
+            "Microsoft.Storage/storageAccounts/blobServices/containers/write": Privilege.WRITE_FILES,
+            "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/write": Privilege.WRITE_FILES,
+            "Microsoft.Storage/storageAccounts/blobServices/containers/read": Privilege.READ_FILES,
+            "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/read": Privilege.READ_FILES,
+        }
+
+    def _get_permission_level(self, permission_to_match: str) -> Privilege | None:
+        for each_level, privilege_level in self._permission_levels.items():
+            # Check for storage blob permission with regex to account for star pattern
+            match = re.search(permission_to_match, each_level)
```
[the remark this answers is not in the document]
mohanab-db: @JCZuurmond - Done, please review.

src/databricks/labs/ucx/azure/access.py
```diff
@@ -54,6 +56,43 @@ def __init__(
...
+            # If a write permission is found, no need to check for read permissions
```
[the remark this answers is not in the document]
mohanab-db: Done, please review.

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -435,27 +437,52 @@ def _role_assignment(
...
         if not scope:
             return None
-        role_name = self._role_name(role_definition_id)
-        if not role_name:
+        role_details = self._role_name(role_definition_id)
+        if not role_details['roleName']:
             return None
+        role_name = str(role_details['roleName'])
+        role_type = str(role_details['roleType'])
+        role_permissions = role_details['rolePermissions'] if isinstance(role_details['rolePermissions'], list) else []
```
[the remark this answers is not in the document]
mohanab-db: Done, please review.

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -435,27 +444,45 @@ def _role_assignment(
...
         if scope == "/":
             scope = resource_id
         return AzureRoleAssignment(
-            resource=AzureResource(resource_id), scope=AzureResource(scope), principal=principal, role_name=role_name
+            resource=AzureResource(resource_id),
+            scope=AzureResource(scope),
+            principal=principal,
+            role_name=role_name,
+            role_type=role_type,
+            role_permissions=role_permissions,
```
nkvuong, line 464: you can assign directly from the dataclass
```suggestion
            role_name=role_details.role_name,
            role_type=role_details.role_type,
            role_permissions=role_details.role_permissions,
```

src/databricks/labs/ucx/azure/access.py
```diff
@@ -54,6 +56,50 @@ def __init__(
...
+    def _get_custom_role_privilege(self, role_permissions: list[str]) -> Privilege | None:
+        # If both read and write privileges are found, only write privilege will be considered
+        higher_privilege = None
+        for each_permission in role_permissions:
+            privileges = self._get_permission_level(each_permission)
+            if privileges and privileges == Privilege.READ_FILES:
+                higher_privilege = Privilege.READ_FILES
+            elif privileges and privileges == Privilege.WRITE_FILES:
+                higher_privilege = Privilege.WRITE_FILES
+                break
```
nkvuong, line 92:
```suggestion
        for each_permission in role_permissions:
            privilege = self._get_permission_level(each_permission)
            if not privilege:
                continue
            # WRITE_FILES is the higher permission, don't need to check further    
            if privilege == Privilege.WRITE_FILES:
                return privilege
            if privilege == Privilege.READ_FILES:
                higher_privilege = Privilege.READ_FILES                
```

src/databricks/labs/ucx/azure/access.py
```diff
@@ -54,6 +56,50 @@ def __init__(
...
+            "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/write": Privilege.WRITE_FILES,
+            "Microsoft.Storage/storageAccounts/blobServices/containers/read": Privilege.READ_FILES,
+            "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/read": Privilege.READ_FILES,
+        }
+
+    def _get_permission_level(self, permission_to_match: str) -> Privilege | None:
+        # If string contains '*', construct a pattern with wildcard
+        if '*' in permission_to_match:
+            parts = re.escape(permission_to_match).split(r'\*')
+            pattern = '^' + parts[0] + '.*' + parts[1] + '$'
```
nkvuong, line 70: this one feels fragile to me, what if there are multiple * to match?

src/databricks/labs/ucx/azure/access.py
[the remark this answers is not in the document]
JCZuurmond: Ah, I now see that the existing code indeed creates multiple `StoragePermissionMapping` for the same principal when it is assigned multiple roles. So, let's keep that the same for your code changes.

However, we need to figure out if this goes well. Reading the code in `ServicePrincipalMigration` I don't see deduplication of the storage permissions and the create storage credentials are named after service principals. This creates a collision, with a chance of creating the credential with a lower permission (read instead of write). 

@nkvuong : Could you chime here? Do I read that correctly?

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -140,6 +140,15 @@ class AzureRoleAssignment:
...
     role_name: str
+    role_type: str
+    role_permissions: list
+
+
+@dataclass
+class AzureRoleDetails:
+    role_name: str | None
+    role_type: str
+    role_permissions: list
```
JCZuurmond, line 151:
```suggestion
    role_permissions: list[str]
```

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -140,6 +140,15 @@ class AzureRoleAssignment:
     scope: AzureResource
     principal: Principal
     role_name: str
+    role_type: str
+    role_permissions: list
```
JCZuurmond, line 144:
```suggestion
    role_permissions: list[str]
```

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -435,27 +444,45 @@ def _role_assignment(
...
         if role_definition_id not in self._role_definitions:
             role_definition = self._mgmt.get(role_definition_id, "2022-04-01")
             definition_properties = role_definition.get("properties", {})
             role_name: str = definition_properties.get("roleName")
             if not role_name:
-                return None
-            self._role_definitions[role_definition_id] = role_name
-        return self._role_definitions.get(role_definition_id)
+                return AzureRoleDetails(role_name=None, role_type='BuiltInRole', role_permissions=[])
+            role_type: str = definition_properties.get("type", "BuiltInRole")
```
JCZuurmond, line 474:
```suggestion
            role_type = definition_properties.get("type", "BuiltInRole")
```

src/databricks/labs/ucx/azure/access.py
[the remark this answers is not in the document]
nkvuong: that looks correct to me. the risk of collision was lower before as it is less likely for a principal to have both contributor & read access (it is redundant). if we now include custom roles, the collision risk is higher, so I would prefer if the changes are added in this PR

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -435,27 +444,45 @@ def _role_assignment(
...
+            role_permissions = []
+            if role_type == 'CustomRole':
+                role_permissions_list = definition_properties.get("permissions")
```
JCZuurmond, line 477:
```suggestion
                role_permissions_list = definition_properties.get("permissions", [])
```

src/databricks/labs/ucx/azure/access.py
JCZuurmond: Is the regex needed? Would a exact match not suffice?

JCZuurmond (changes requested):

src/databricks/labs/ucx/azure/resources.py
mohanab-db: @nkvuong - Thanks, I have added this.

src/databricks/labs/ucx/azure/access.py
mohanab-db: @nkvuong - Thanks, I have added this.

src/databricks/labs/ucx/azure/resources.py
mohanab-db: @JCZuurmond - Thanks, I have added this.

src/databricks/labs/ucx/azure/resources.py
mohanab-db: @JCZuurmond - Thanks, I have added this.

src/databricks/labs/ucx/azure/resources.py
mohanab-db: @JCZuurmond - Thanks, I have added this.

src/databricks/labs/ucx/azure/resources.py
mohanab-db: @JCZuurmond - Thanks, I have added this.

src/databricks/labs/ucx/azure/access.py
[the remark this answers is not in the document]
mohanab-db: @JCZuurmond @nkvuong - I have made changes to handle this scenario in the last commit. Please review.

src/databricks/labs/ucx/azure/access.py
```diff
@@ -54,28 +57,75 @@ def __init__(
...
+                higher_privilege = privilege
+        return higher_privilege
+
+    def _get_role_privilege(self, role_assignment: AzureRoleAssignment) -> Privilege | None:
+        privilege = None
+        # Check for custom role permissions on the storage accounts
+        if role_assignment.role_permissions:
+            privilege = self._get_custom_role_privilege(role_assignment.role_permissions)
+        elif role_assignment.role_name in self._levels:
+            privilege = self._levels[role_assignment.role_name]
```
nkvuong, line 99: this mean custom roles take higher priority than pre-defined role, which could be wrong. the cleanest approach is to collate all the output into a set and then emit WRITE_FILES if it is in the set, else emit READ_FILES.

mohanab-db: @nkvuong Thanks for reviewing. One role assignment can either be Custom, or Built-in. If one SP (sp1) has 2 role assignments (one custom read and one built-in write) for a container, then the first iteration in the loop will put READ in the principal_spm_mapping dict. The next iteration will find the write access from built-in and update the dict with the WRITE access.
I have added this scenario to unit test test_save_spn_permissions_custom_role_valid_azure_storage_account() for Custom_role_11. Please review and let me know if this doesn't seem correct.

src/databricks/labs/ucx/azure/access.py
```diff
@@ -98,28 +99,33 @@ def _get_role_privilege(self, role_assignment: AzureRoleAssignment) -> Privilege
...
 
-    def _map_storage(self, storage: StorageAccount) -> list[StoragePermissionMapping]:
+    def _map_storage(self, storage: StorageAccount) -> ValuesView[StoragePermissionMapping]:
         logger.info(f"Fetching role assignment for {storage.name}")
-        out = []
+        principal_spm_mapping: dict[str, StoragePermissionMapping] = {}
         for container in self._azurerm.containers(storage.id):
             for role_assignment in self._azurerm.role_assignments(str(container)):
+                # Skip the role assignments that already have WRITE_FILES privilege
+                spm_mapping_key = f"{container.container}_{role_assignment.principal.client_id}"
```
JCZuurmond, line 108: This could cause a collision when two storage accounts have the containers with the same name. 

``` suggestion
                spm_mapping_key = f"{container}_{role_assignment.principal.client_id}"
```

src/databricks/labs/ucx/azure/access.py
```diff
@@ -3,6 +3,8 @@
 import uuid
 from dataclasses import dataclass
 from functools import partial
+import re
+from collections.abc import ValuesView
```
JCZuurmond, line 7: Learned something new here :)

JCZuurmond, line 7:
``` suggestion
import re
import uuid
from collections.abc import ValuesView
from dataclasses import dataclass
from functools import partial
```

src/databricks/labs/ucx/azure/access.py
```diff
@@ -16,6 +18,7 @@
     AzureResources,
     PrincipalSecret,
     StorageAccount,
+    AzureRoleAssignment,
```
JCZuurmond, line 21: Also isort these

src/databricks/labs/ucx/azure/access.py
```diff
@@ -54,28 +57,75 @@ def __init__(
...
+
+    def _get_permission_level(self, permission_to_match: str) -> Privilege | None:
+        # String might contain '*', check for wildcard match
+        pattern = re.sub(r'\*', '.*', permission_to_match)
+        permission_compiled = re.compile(pattern)
+        for each_level, privilege_level in self._permission_levels.items():
+            # Check for storage blob permission with regex to account for star pattern
+            match = permission_compiled.match(each_level)
+            if match:
+                # If a write permission is found, no need to check for read permissions
```
JCZuurmond, line 75: Comment seems out of place

src/databricks/labs/ucx/azure/access.py
```diff
@@ -54,28 +57,75 @@ def __init__(
...
+                return privilege_level
+        return None
+
+    def _get_custom_role_privilege(self, role_permissions: list[str]) -> Privilege | None:
+        # If both read and write privileges are found, only write privilege will be considered
+        higher_privilege = None
+        for each_permission in role_permissions:
+            privilege = self._get_permission_level(each_permission)
+            if not privilege:
```
JCZuurmond, line 84:
```suggestion
            if privilege is None:
```

src/databricks/labs/ucx/azure/access.py
```diff
@@ -54,28 +57,75 @@ def __init__(
...
+        permission_compiled = re.compile(pattern)
+        for each_level, privilege_level in self._permission_levels.items():
+            # Check for storage blob permission with regex to account for star pattern
+            match = permission_compiled.match(each_level)
+            if match:
+                # If a write permission is found, no need to check for read permissions
+                return privilege_level
+        return None
+
+    def _get_custom_role_privilege(self, role_permissions: list[str]) -> Privilege | None:
```
JCZuurmond, line 79: It might make sense to let this return a list of privileges and select the highest privilege in `_map_storage`. Now the logic for selecting the highest priviliges is duplicated.

JCZuurmond (changes requested):
@mohanab-db : Thanks for implementing our feedback. Functionaly we are getting there, have some remarks about the implementation

src/databricks/labs/ucx/azure/access.py
mohanab-db: @JCZuurmond The _map_storage() method is called for every storage account separately and the dict principal_spm_mapping gets refreshed for every storage account. The dict will only contain container names for one storage account at a time.

src/databricks/labs/ucx/azure/access.py
mohanab-db: @JCZuurmond Thanks. Code updated with this.

src/databricks/labs/ucx/azure/access.py
mohanab-db: @JCZuurmond - Done

src/databricks/labs/ucx/azure/access.py
mohanab-db: @JCZuurmond  - Done.

src/databricks/labs/ucx/azure/access.py
mohanab-db: @JCZuurmond - Thanks. Changed.

src/databricks/labs/ucx/azure/access.py
mohanab-db: @JCZuurmond - The method _get_custom_role_privilege() gets the higher privilege for one custom role assignment. A custom role can contain multiple permissions, and this will get the higher privilege from these permissions. Method _map_storage() will get the higher privilege from a custom or builtin role if more than one role assignments exist for one principal. If I want to return a list of privileges for one custom role, then I will need to iterate through all permissions, whereas now I am exiting from the loop if I find one Write privilege. Please let me know if you think this makes sense.

JCZuurmond (approved):
LGTM!

src/databricks/labs/ucx/azure/access.py
JCZuurmond: yes, that is also true. I am a bit indifferent about this, therefore, I would not let it block this PR. 

Pros: exit loop early, a pattern you see more often in ucx
Cons: treating WRITE_FILES > READ_FILES in two location, speed gain is probably insignifficant

src/databricks/labs/ucx/azure/access.py
mohanab-db: @nkvuong Please let me know if this makes sense, or if some changes should be made.

src/databricks/labs/ucx/azure/access.py
mohanab-db: @nkvuong Please let me know if this makes sense, or if some changes should be made.

JCZuurmond (changes requested):
@mohanab-db : Don't forget to update the README as it now only mentions the built-in roles

mohanab-db:
> @mohanab-db : Don't forget to update the README as it now only mentions the built-in roles

@JCZuurmond - Sorry, I am unable to find the documentation that mentions about built-in roles. I can just see comments about the service principal in general.
image
image
Please point me to the correct documentation so that I can update it.

JCZuurmond:
Hi @mohanab-db, I saw it being mentioned in the `migrate-credentials` command. I think we should move this information to the `principal-prefix-access` command and reference the SP in `migrate-credentials` to be the ones created by `principal-prefix-access`

mohanab-db:
@JCZuurmond - I have updated README.md. Please review.

nkvuong (approved):

JCZuurmond (approved):
lgtm

PR #1487, opened 2024-04-22
Add support for migrating external location permissions from interactive cluster mounts

HariGS-DB:
## Changes
 - Gets the locations of interactive cluster from the AzureACL class
 - Gets the interactive cluster user mapping
 - for each location under each cluster grants CREATE EXTERNAL TABLE, CREATE EXTERNAL VOLUME and READ_FILES permission to each cluster principal

Linked issues #1192 #1193 
Resolves #1192 #1193

### Functionality 

- [ ] added relevant user documentation
- [ ] added new CLI command
- [X] modified existing command: `databricks labs ucx ...`
- [ ] added a new workflow
- [ ] modified existing workflow: `...`
- [ ] added a new table
- [ ] modified existing table: `...`

### Tests

- [X] manually tested
- [X] added unit tests
- [X] added integration tests
- [ ] verified on staging environment (screenshot attached)

PR #1583, opened 2024-04-29
Split AWSResourcePermission into separate classes

### LINKED ISSUE #1572: [FEATURE]: Split AWSResourcePermission into separate classes

### Problem statement

AWSResourcePermission is currently too big; this results in many classes referring to this and running into issues of circular dependency. 

### Proposed Solution

In order to avoid this, we need to split this class into smaller classes much like how it's done for Azure

### Additional Context

_No response_

nkvuong:
## Changes
- Split ExternalLocationsMigration related functions into a new class

### Linked issues

Resolves #1572

### Tests

- [x] manually tested


nkvuong:
need to wait for #1487 to be merged

src/databricks/labs/ucx/aws/access.py
```diff
@@ -35,7 +34,6 @@ def __init__(
         backend: SqlBackend,
```
nfx, line 34: also remove this one

src/databricks/labs/ucx/aws/access.py
```diff
@@ -44,7 +42,6 @@ def __init__(
         self._backend = backend
         self._ws = ws
         self._locations = external_locations
-        self._schema = schema
         self._aws_account_id = aws_account_id
         self._kms_key = kms_key
         self._filename = self.INSTANCE_PROFILES_FILE_NAMES
```
nfx, line 47: this one is not needed - this is the copy of the constant on the same class

src/databricks/labs/ucx/aws/locations.py
```diff
@@ -0,0 +1,88 @@
...
+class AWSExternalLocationsMigration:
+
+    def __init__(
+        self,
+        ws: WorkspaceClient,
+        external_locations: ExternalLocations,
+        aws_resource_permissions: AWSResourcePermissions,
+    ):
+        self._ws = ws
+        self._locations = external_locations
```
nfx, line 24:
```suggestion
        self._external_locations = external_locations
```

src/databricks/labs/ucx/aws/locations.py
```diff
@@ -0,0 +1,88 @@
...
+        external_location_num = 1
+        for path, role_arn in missing_paths:
+            if role_arn not in credential_dict:
+                logger.error(f"Missing credential for role {role_arn} for path {path}")
+                continue
+            while True:
+                external_location_name = f"{location_init}_{external_location_num}"
+                if external_location_name not in external_location_names:
+                    break
+                external_location_num += 1
```
nfx, line 47: can you turn these lines into a private method? there might be a list of infinite loop and it's always easier to find one with stack traces if it's in an isolated method

nfx (approved):

src/databricks/labs/ucx/aws/access.py
nkvuong: removed

src/databricks/labs/ucx/aws/access.py
nkvuong: removed

nfx:
@nkvuong what can be done to remove flakiness here?

nfx (approved):

PR #1426, opened 2024-04-17
Extend service principal migration to create storage credentials for access connectors created for each storage account

JCZuurmond:
## Changes
Extend service principal migration to create storage credentials for access connectors created for each storage account

### Linked issues

Resolves #1384 
Resolves #875

### To do
- [x] Update docs on creating access connectors. 
- [x] First prompt for creating credentials for access connectors, then for service princpals
- [x] Use the storage credentials in [external locations](https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/azure/locations.py#L65)

### Functionality 

- [ ] added relevant user documentation
- [ ] added new CLI command
- [x] modified existing command: `databricks labs ucx ...`
- [ ] added a new workflow
- [ ] modified existing workflow: `...`
- [ ] added a new table
- [ ] modified existing table: `...`

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [ ] verified on staging environment (screenshot attached)

PR #1589, opened 2024-04-30
Feature: Suggest users to create Access Connector(s) with Managed Identity to access Azure Storage Accounts behind firewall

### LINKED ISSUE #915: [FEATURE]: Suggest users to create MI for storage accounts with firewall

### Problem statement

UC credentials using SPN secrets do not work with storage accounts with firewall enabled. Customers need to switch to MI

### Proposed Solution

- Detect storage accounts with firewalls using AzureRM - https://github.com/databrickslabs/ucx/issues/886
- Flag that SPNs that have access to these storage accounts would stop working and users need to migrate to MI - https://github.com/databrickslabs/ucx/issues/339

### Additional Context

_No response_

JCZuurmond:
## Changes
Suggest users to create Access Connector(s) with Managed Identity to access Azure Storage Accounts behind firewall. Service principals can not be used to access storage accounts behind firewalls. We should warn users that this could break when migrating to UC

### Linked issues
Resolves #915

### Functionality 

- [ ] added new CLI command
- [x] modified existing command: `databricks labs ucx ...`
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/azure/access.py
+++ b/src/databricks/labs/ucx/azure/access.py
@@ -31,6 +31,7 @@ class StoragePermissionMapping:
     principal: str
     privilege: str
     type: str
+    default_network_action: str  # "Deny" or "Allow"
     # Need this directory_id/tenant_id when create UC storage credentials using service principal
     directory_id: str | None = None
 
@@ -72,6 +73,7 @@ class AzureResourcePermissions:
                         principal=role_assignment.principal.display_name,
                         privilege=privilege,
                         type=role_assignment.principal.type,
+                        default_network_action=storage.default_network_action,
                         directory_id=role_assignment.principal.directory_id,
                     )
                 )
--- a/src/databricks/labs/ucx/azure/access.py
+++ b/src/databricks/labs/ucx/azure/access.py
@@ -31,7 +31,7 @@ class StoragePermissionMapping:
     principal: str
     privilege: str
     type: str
-    default_network_action: str  # "Deny" or "Allow"
+    default_network_action: str = "Unknown"  # "Deny", "Allow" or "Unknown"
     # Need this directory_id/tenant_id when create UC storage credentials using service principal
     directory_id: str | None = None
 
--- a/src/databricks/labs/ucx/azure/credentials.py
+++ b/src/databricks/labs/ucx/azure/credentials.py
@@ -234,8 +234,17 @@ class ServicePrincipalMigration(SecretsMixin):
 
         execution_result = []
         for spn in sp_list_with_secret:
+            if spn.permission_mapping.default_network_action != "Allow":
+                logger.warning(
+                    f"Service principal '{spn.permission_mapping.principal}' accesses storage account "
+                    f"'{spn.permission_mapping.prefix}' with non-Allow network configuration, which might cause "
+                    "connectivity issues. We recommend using access connectors with managed identities instead "
+                    "(confirm with prompt below)."
+                )
+
             self._storage_credential_manager.create_with_client_secret(spn)
             execution_result.append(self._storage_credential_manager.validate(spn.permission_mapping))
+
         return execution_result
 
     def _create_access_connectors_for_storage_accounts(self) -> list[StorageCredentialValidationResult]:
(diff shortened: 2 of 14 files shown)


JCZuurmond:
> @JCZuurmond option a is most logical to me - what is the risk of adding a new column to the csv? Does blueprint throw an error if there are missing/additional columns in the file to be loaded?

Thanks @nkvuong for your input. I will go for that option. I will investigate if blueprint creates an error due to this. At the minimum, we need a default for this column when using it in the service principal migration as some users might not have updated the csv by rerunning the SP permission mapping.

src/databricks/labs/ucx/azure/access.py
```diff
@@ -31,6 +31,7 @@ class StoragePermissionMapping:
     principal: str
     privilege: str
     type: str
+    default_network_action: str = "Unknown"  # "Deny", "Allow" or "Unknown"
```
JCZuurmond, line 34: @nkvuong : It's confirmed that with a default we can read csv's missing this column. Could you please review the PR?

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -234,8 +234,17 @@ def _migrate_service_principals(
 
         execution_result = []
         for spn in sp_list_with_secret:
+            if spn.permission_mapping.default_network_action != "Allow":
+                logger.warning(
```
nkvuong, line 238: does this mean we still create the storage credential with service principal, and then will create another storage credential with managed identity? shouldn't it be either/or?

nkvuong (changes requested):
could you also add some documentation for this feature?

JCZuurmond:
> could you also add some documentation for this feature?

Done, docs on the creating access connectors will be added in #1426

src/databricks/labs/ucx/azure/credentials.py
JCZuurmond: @nkvuong : Thanks for your input - once more. 

I gave it a though:
1. I am not 100% sure that storage credentials with service principals that access storage behind firewalls will **always** be an issue. It's probably likely, though, it could not be the case. 
2. At least we want to notify the user about the risk of (1)
3. Additionally, we should nudge users to use access connectors as it avoids (1) and is a Databricks best practice

I updated the code to prompt the user once more before migrating **all** service principals if at least one service principal is used to access a storage accounts behind firewall. Additionally, we still warn the user with details on which service principal access which storage account behind a firewall, if the user whises to continue.

JCZuurmond: On the either/or, we decided to support creating both storage credentials for service principals and access connectors to provide flexibility to the user. Though, we (should) make it clear that we prefer access connectors.

nkvuong: my thinking is that you need to specify a storage credential per external location (~ a storage account). So the external location migration logic needs to pick one, our preference would be to pick the one with an access connector

nkvuong: the logic here is a bit fragile - https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/azure/locations.py#L65

JCZuurmond: let's cover that in #1426 as that PR creates the storage credentials for the access connectors with managed identities

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -246,16 +251,27 @@ def run(self, prompts: Prompts, include_names: set[str] | None = None) -> list[S
...
+                    "configuration, which might cause connectivity issues. We recommend using Databricks Access "
+                    "Connectors instead (next prompt). Would you like to continue with migrating the service "
+                    "principals?"
+                )
+            if plan_confirmed:
+                sp_results = self._migrate_service_principals(sp_migration_infos)
 
-        plan_confirmed = prompts.confirm("Please confirm to create an access connector for each storage account.")
+        plan_confirmed = prompts.confirm(
+            "[RECOMMENDED] Please confirm to create an access connector with a managed identity for each storage "
```
nkvuong, line 269: nit on this one - our default quota for storage credential is 200 versus 500 for external location, so creating one access connector per storage account might hit quota limit for customers - we could do address this in a follow up PR though

nkvuong (approved):
lgtm, just one thought about quota that we may need to revisit

nfx (approved):
lgtm

PR #1595, opened 2024-04-30
Modified permissions crawler to only emit permissions for groups specified in config

nkvuong:
## Changes
- Only emit permissions for groups specified in config. This avoids flakiness in integration tests such as `test_running_real_validate_groups_permissions_job`

### Tests

- [x] manually tested


src/databricks/labs/ucx/workspace_access/secrets.py
```diff
@@ -10,22 +10,26 @@
...
 logger = logging.getLogger(__name__)
 
 
 class SecretScopesSupport(AclSupport):
-    def __init__(self, ws: WorkspaceClient, verify_timeout: timedelta | None = None):
+    def __init__(
+        self,
+        ws: WorkspaceClient,
+        verify_timeout: timedelta | None = None,
+        include_object_permissions: list[str] | None = None,
```
JCZuurmond, line 24: Please add here that this parameter is intended for testing only

src/databricks/labs/ucx/workspace_access/redash.py
```diff
@@ -73,7 +73,7 @@ def _is_item_relevant(item: Permissions, migration_state: MigrationState) -> boo
     def get_crawler_tasks(self):
         if self._include_object_permissions:
             for item in StaticListing(self._include_object_permissions, self.object_types()):
-                yield partial(self._crawler_task, item.object_id, item.object_type)
+                yield partial(self._crawler_task, item.object_id, sql.ObjectTypePlural(item.object_type))
```
JCZuurmond, line 76: Does this change anything? Or is it "just" better practice to reuse the enum from the sdk?

src/databricks/labs/ucx/contexts/application.py
```diff
@@ -166,7 +166,9 @@ def scim_entitlements_support(self):
     @cached_property
     def secret_scope_acl_support(self):
         # Secret ACLs are not used much in tests, so skipping include_object_permissions
-        return SecretScopesSupport(self.workspace_client)
+        return SecretScopesSupport(
+            self.workspace_client, include_object_permissions=self.config.include_object_permissions
```
JCZuurmond, line 170: Should only the object permissions that start with `secrets:` be passed on? Or all?

JCZuurmond (approved):
LGTM, have some clarification questions

src/databricks/labs/ucx/workspace_access/secrets.py
nkvuong: this is used by other ACL support classes as well - added comments for all of them

src/databricks/labs/ucx/workspace_access/redash.py
nkvuong: `StaticListing` basically mocks what the listing would return to the crawler. The crawler expects an enum and would crash if a string is returned

src/databricks/labs/ucx/contexts/application.py
nkvuong: `StaticListing` handles the filtering, clarified in comments

nfx (approved):
thank you

PR #1598, opened 2024-05-01
Convert `account.py` into `account` package

nfx:
This PR splits `account.py` into few more modules, so that we can avoid circular dependencies the more work we do on the account level.

### DIFF
--- a/src/databricks/labs/ucx/account/workspaces.py
+++ b/src/databricks/labs/ucx/account/workspaces.py
@@ -3,43 +3,27 @@ from typing import ClassVar
 
 from databricks.labs.blueprint.installation import Installation
 from databricks.labs.blueprint.tui import Prompts
-from databricks.sdk import AccountClient, WorkspaceClient
-from databricks.sdk.errors import NotFound, ResourceConflict, PermissionDenied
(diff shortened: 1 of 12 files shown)


pritishpai (approved):
LGTM (before acceptance test completion)

src/databricks/labs/ucx/account/workspaces.py
```diff
@@ -3,43 +3,27 @@
...
-    }
-
     SYNC_FILE_NAME: ClassVar[str] = "workspaces.json"
 
-    def __init__(self, account_client: AccountClient, new_workspace_client=WorkspaceClient):
-        # TODO: new_workspace_client is a design flaw, remove it
-        self._new_workspace_client = new_workspace_client
+    def __init__(self, account_client: AccountClient, include_workspace_ids: list[int] | None = None):
         self._ac = account_client
+        self._include_workspace_ids = include_workspace_ids if include_workspace_ids else []
```
JCZuurmond, line 18:
```suggestion
        self._include_workspace_ids = include_workspace_ids or []
```

JCZuurmond (commented):
lgtm

Issue #1499, opened 2024-04-23
[FEATURE]: Implement LocalFileLoader and integration tests


ericvergnaud:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

https://github.com/databrickslabs/ucx/pull/1462 implements a dependency graph for local files, but since @ericvergnaud does not yet have access to DB infra, only unit tests have been implemented.
There is a need to implement `LocalFileLoader` and `LocalNotebookLoader`, and create proper integration tests.

### Proposed Solution

Omplement `LocalFileLoader` and `LocalNotebookLoader`, and create proper integration tests.

Issue #1287, opened 2024-04-05
[FEATURE]: Convert relative paths to full paths when building notebook dependency graph 


ericvergnaud:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Commands such as `%run ./some_notebook` refer to a notebook path that can be relative to the currently executing notebook.
Depending on where from the above command is invoked, the notebook path for the same notebook could vary significantly.
We need to support this variation of notebook paths when building the dependency graph

### Proposed Solution

The DependencyGraph already accepts a locator function. 
The locator used for unit testing only deals with local samples in the same folder.
This production locator should be context aware and convert relative paths to full paths, such that we only traverse notebooks once.

PR #1593, opened 2024-04-30
Resolve relative paths when building dependency graph

ericvergnaud:
@nfx can be reviewed but depends on #1560  

## Changes
Add support for cwd to SysPathProvider
Implement LocalNotebookLoader
Provide tests for local files and folders

### Linked issues
#1202 
Resolves #1499 
Resolves #1287

### Functionality 

- [ ] added relevant user documentation
- [ ] added new CLI command
- [ ] modified existing command: `databricks labs ucx ...`
- [ ] added a new workflow
- [ ] modified existing workflow: `...`
- [ ] added a new table
- [ ] modified existing table: `...`

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] verified on staging environment (screenshot attached)

PR #1608, opened 2024-05-02
Resolve relative paths when building dependency graph

ericvergnaud:
## Changes
Add support for cwd to SysPathProvider
Implement LocalNotebookLoader
Provide tests for local files and folders

### Linked issues
https://github.com/databrickslabs/ucx/issues/1202
Resolves https://github.com/databrickslabs/ucx/issues/1499
Resolves https://github.com/databrickslabs/ucx/issues/1287

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests
- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests

replaces #1593


src/databricks/labs/ucx/source_code/notebooks/base.py
```diff
@@ -0,0 +1 @@
+NOTEBOOK_HEADER = "Databricks notebook source"
```
nfx, line 1: perhaps the `cells.py` was not the best name for the purpose. `nbformat.py` night be a better name. could you bring it back to cells.py?

src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -16,10 +17,12 @@ def __init__(
         dependency: Dependency,
         parent: DependencyGraph | None,
         resolver: DependencyResolver,
+        syspath_provider: SysPathProvider,
     ):
         self._dependency = dependency
         self._parent = parent
         self._resolver = resolver
+        self._syspath_provider = syspath_provider
```
nfx, line 25:
```suggestion
        self._syspath_provider = syspath_provider
        syspath_provider.push_cwd(dependency.path.parent)
```

this resolves hidden downstream bugs and prevents the need to:

1. change the interface from source container.
2. push/pop cwd in graph building implementations, that we may miss. 

image

P.S. shouldn't `SysPathProvider` be renamed to `RelativePaths`, `PathLookup` or something like it?...

src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -192,7 +195,7 @@ def load(self) -> SourceContainer | None:
 class SourceContainer(abc.ABC):
 
     @abc.abstractmethod
-    def build_dependency_graph(self, parent: DependencyGraph) -> None:
+    def build_dependency_graph(self, parent: DependencyGraph, syspath_provider: SysPathProvider) -> None:
```
nfx, line 198: this change is unnecessary, see the comment on graph

tests/unit/source_code/test_syspath_simulation.py
```diff
@@ -0,0 +1,72 @@
...
+    file_path = Path(*elems)
+    whitelist = whitelist_mock()
+    provider = SysPathProvider.from_sys_path()
+    file_loader = VisitingFileLoader(provider, visited)
+    notebook_loader = VisitingNotebookLoader(provider, visited)
+    site_packages = SitePackages.parse(locate_site_packages())
+    resolvers = [
+        NotebookResolver(notebook_loader),
+        SitePackagesResolver(site_packages, file_loader, provider),
+        WhitelistResolver(whitelist),
```
nfx, line 67: only `NotebookResolver` and `LocalFileResolver` are per-se triggered, why do we need whitelist and sitepkgs?

src/databricks/labs/ucx/source_code/graph.py
ericvergnaud: I like `PathLookup`

ericvergnaud: Re `push_cwd`, I'm not sure if the traversal strictly follows the instantiation, so we'd have to test if it can work. And as discussed on slack I don't think it can work with `sys.path`. (I appreciate this PR does not restrict itself to changes required by `cwd`, I was heading for `sys.path` and discovered the need to handle `cwd` first).
I'd rather implement `sys.path` fully (next PR) and then simplify if that's ok with you.

tests/unit/source_code/test_syspath_simulation.py
ericvergnaud: because we want to ensure that NotebookResolver that other resolvers are not preventing the triggering. We're testing the DependencyResolver, not the NotebookResolver.

src/databricks/labs/ucx/source_code/notebooks/base.py
ericvergnaud: the file name isn't the problem here. This is about avoiding cyclic imports.

src/databricks/labs/ucx/source_code/site_packages.py
```diff
@@ -21,7 +22,7 @@ def __init__(
         self,
         site_packages: SitePackages,
         file_loader: FileLoader,
-        syspath_provider: SysPathProvider,
+        syspath_provider: PathLookup,
```
nfx, line 25: also rename the `syspath_provider` references to `path_lookup`

src/databricks/labs/ucx/source_code/path_lookup.py
```diff
@@ -0,0 +1,52 @@
...
+        paths = syspath.split(':')
+        return PathLookup([Path(path) for path in paths])
+
+    @classmethod
+    def from_sys_path(cls):
+        return PathLookup([Path(path) for path in sys.path])
+
+    def __init__(self, paths: list[Path]):
+        self._paths = paths
+        self._cwds = [Path(os.getcwd())]
```
nfx, line 22:
```suggestion
    def __init__(self, cwd: Path, sys_path: list[Path]):
        self._sys_path = paths
        self._cwd = cwd
    
    def cd(self, working_directory: Path):
        return PathLookup(working_directory, self._sys_path)
    
    @propertly
    def parent(self):
        return PathLookup(self._cwd.parent, self._sys_path)
```
why do we need multiple cwds?... don't use `os.getcwd()` - it'll lead to undefined behavior, as people won't read the sources to know it's there.

`sys.path` is shared within a notebook context:

image

nfx (changes requested):
remove `os.getcwd()`, as it'll lead to undefined behavior.

tests/unit/source_code/test_syspath_simulation.py
nfx: if we're testing `DependencyResolver`, why do we need concrete implementations of "sub-resolvers" then? ;)

src/databricks/labs/ucx/source_code/graph.py
nfx: it's incorrect to rely on `os.getcwd()` here, as it'll lead to undefined behavior.

https://github.com/databrickslabs/ucx/pull/1608#pullrequestreview-2036583040

src/databricks/labs/ucx/source_code/path_lookup.py
ericvergnaud: we don't have multiple cwds, only a stack for push/pop
if we can't use os.getcwd(), how should we populate the initial cwd ?

ericvergnaud: Re `sys.path`, thanks for confirming that it's shared within an entire notebook. But what happens when 'switching' notebook via `%run` or `dbutils.notebook.run` ? Does the target notebook benefit from a fresh `sys.path` or does it inherit the caller's ?

tests/unit/source_code/test_syspath_simulation.py
ericvergnaud: We're testing a _populated_ `DependencyResolver`.  Its behavior depends on its population.

src/databricks/labs/ucx/source_code/site_packages.py
ericvergnaud: done

src/databricks/labs/ucx/contexts/application.py
```diff
@@ -356,16 +358,17 @@ def site_packages(self):
         return SitePackages([])
 
     @cached_property
-    def syspath_provider(self):
-        return SysPathProvider.from_sys_path()
+    def path_lookup(self):
```
ericvergnaud, line 361: @nfx we might have a serious design issue/limitation here, worth discussing...

src/databricks/labs/ucx/source_code/path_lookup.py
```diff
@@ -0,0 +1,52 @@
...
+        self._cwds.append(path)
+
+    def pop_cwd(self):
+        result = self._cwds[0]
+        del self._cwds[0]
+        return result
+
+    @property
+    def cwd(self):
+        return self._cwds[-1] if len(self._cwds) > 0 else Path(os.getcwd())
```
nfx, line 52: we still have this implicit `os.getcwd` here

src/databricks/labs/ucx/source_code/path_lookup.py
```diff
@@ -0,0 +1,52 @@
...
+        del self._sys_paths[index]
+
+    def pop_path(self) -> Path:
+        result = self._sys_paths[0]
+        del self._sys_paths[0]
+        return result
+
+    @property
+    def paths(self) -> Iterable[Path]:
+        yield from self._sys_paths
```
nfx, line 40:
```suggestion
        yield self._cwd
        yield from self._sys_paths
```
missing cwd in front

src/databricks/labs/ucx/source_code/path_lookup.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def push_cwd(self, path: Path):
+        self._cwds.append(path)
+
+    def pop_cwd(self):
+        result = self._cwds[0]
+        del self._cwds[0]
+        return result
```
nfx, line 48:
```suggestion
```
why do we need to have a stack of cwds? can't it be just a single cwd?

src/databricks/labs/ucx/source_code/notebooks/sources.py
```diff
@@ -58,11 +61,13 @@ def to_migrated_code(self):
             sources.append('')  # following join will append lf
         return '\n'.join(sources)
 
-    def build_dependency_graph(self, parent: DependencyGraph) -> None:
+    def build_dependency_graph(self, parent: DependencyGraph, path_lookup: PathLookup) -> None:
```
nfx, line 64:
```suggestion
    def build_dependency_graph(self, parent: DependencyGraph) -> None:
```
don't change the signature.

src/databricks/labs/ucx/source_code/path_lookup.py
```diff
@@ -0,0 +1,52 @@
...
+        paths = syspath.split(':')
+        return PathLookup(cwd, [Path(path) for path in paths])
+
+    @classmethod
+    def from_sys_path(cls, cwd: Path):
+        return PathLookup(cwd, [Path(path) for path in sys.path])
+
+    def __init__(self, cwd: Path, sys_paths: list[Path]):
+        self._sys_paths = sys_paths
+        self._cwds = [cwd]
```
nfx, line 22:
```suggestion
        self._cwds = [cwd]
        
    def change_directory(new_cwd: Path):
        return PathLookup(new_cwd, self._sys_paths)
```

src/databricks/labs/ucx/source_code/files.py
```diff
@@ -153,15 +163,19 @@ def exists(self, path: Path) -> bool:
     def full_path(self, path: Path) -> Path | None:
         if path.is_file():
             return path
-        for parent in self._syspath_provider.paths:
+        child = Path(self._path_lookup.cwd, path)
```
nfx, line 166: this change is not needed, yield cwd from `paths`.

src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -16,10 +17,12 @@ def __init__(
         dependency: Dependency,
         parent: DependencyGraph | None,
         resolver: DependencyResolver,
+        path_lookup: PathLookup,
     ):
         self._dependency = dependency
         self._parent = parent
         self._resolver = resolver
+        self._path_lookup = path_lookup
```
nfx, line 25:
```suggestion
        self._path_lookup = path_lookup.change_directory(dependency.path.parent)
```
remove push_cwd/pop_cwd, they're not needed.

nfx (changes requested):
please simplify

src/databricks/labs/ucx/source_code/path_lookup.py
ericvergnaud: fixed

src/databricks/labs/ucx/source_code/path_lookup.py
ericvergnaud: mmm... can we leave this to the next PR (which will be about sys.path) ?

src/databricks/labs/ucx/source_code/path_lookup.py
ericvergnaud: With the current execution path, yes, because we need to restore the cwd after analyzing a child notebook or file

src/databricks/labs/ucx/source_code/files.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/path_lookup.py
ericvergnaud: that doesn't work, see below

src/databricks/labs/ucx/source_code/notebooks/sources.py
ericvergnaud: it's required, see below

src/databricks/labs/ucx/source_code/graph.py
nfx: > the instantiation path is different from the execution path.

I don't believe this is the case. can you address this comment and let's merge this?

ericvergnaud: This is not an opinion, it's an observation, see https://github.com/databrickslabs/ucx/tree/drop-cwd-stack where I did the work you asked for. Tests in test_path_lookup_simulation fail for the reason explained above. 
My wording is inaccurate, it should read: the instantiation sequence is different from the execution sequence.

src/databricks/labs/ucx/source_code/path_lookup.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/graph.py
nfx: tests are not correct - e.g. you either have to `from .sibling2_file import stuff` (there's a bug in `databricks.labs.ucx.source_code.python_linter.PythonLinter.list_import_sources` that ignores `ast.ImportFrom` level), or add that folder to a syspath.

nfx (approved):
this implementation still has bugs and is not very easy to follow, but fine, let's fix forward.

PR #1623, opened 2024-05-03
Improved reliability of table migration status refresher

### LINKED ISSUE #1622: Make migration status refresher more consistent
> Do you remember why you reset the migration status in-between batches?

It is reset because `self._view_can_be_migrated` uses the migration status to check if dependencies are migrated to decide if a view can be migrated or not. The reset introduces a new `crawl` (not a `fetch` from the inventory database) when the migration status is requested through the `index`.

@FastLee and @nfx : As code improvement // tech debt, I suggest to **always** crawl the migration status and update the migration status in the inventory database. (Not in this PR, but as separate work to be planned accordingly.)

When requesting the migration status in the code (only used in the `table_migrate`), we always want the up-to-date status. Now, we get the up-to-date version by resetting the migration status sometimes, creating implicit dependencies between where the migration status is `reset` and retrieved through `index` - spanning across methods.

It took me quite some time to figure out how/why the migration status was not refreshed in certain cases and why it was in others, and why my first changes let other test fail. IMO the code becomes clearer by removing the implicit dependency.

_Originally posted by @JCZuurmond in https://github.com/databrickslabs/ucx/pull/1599#discussion_r1587407113_

nkvuong:
## Changes
- Always reset table migration status when requesting the latest snapshot within `table_migrate`
- Handle `NotFound` error when refreshing migration status

### Linked issues

Resolves #1622, #1615

### Tests

- [x] manually tested

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/migration_status.py
+++ b/src/databricks/labs/ucx/hive_metastore/migration_status.py
@@ -70,7 +70,7 @@ class MigrationStatusRefresher(CrawlerBase[MigrationStatus]):
         seen_tables: dict[str, str] = {}
         for schema in self._iter_schemas():
             try:
-                tables = self._ws.tables.list(catalog_name=schema.catalog_name, schema_name=schema.name)
+                tables = list(self._ws.tables.list(catalog_name=schema.catalog_name, schema_name=schema.name))
             except NotFound:
                 logger.warning(
                     f"Schema {schema.catalog_name}.{schema.name} no longer exists. Skipping checking its migration status."
--- a/src/databricks/labs/ucx/hive_metastore/migration_status.py
+++ b/src/databricks/labs/ucx/hive_metastore/migration_status.py
@@ -70,6 +70,7 @@ class MigrationStatusRefresher(CrawlerBase[MigrationStatus]):
         seen_tables: dict[str, str] = {}
         for schema in self._iter_schemas():
             try:
+                # ws.tables.list returns Iterator[TableInfo], so we need to convert it to a list in order to catch the exception
                 tables = list(self._ws.tables.list(catalog_name=schema.catalog_name, schema_name=schema.name))
             except NotFound:
                 logger.warning(
--- a/src/databricks/labs/ucx/hive_metastore/table_migrate.py
+++ b/src/databricks/labs/ucx/hive_metastore/table_migrate.py
@@ -57,6 +57,10 @@ class TablesMigrator:
         self._principal_grants = principal_grants
 
     def index(self):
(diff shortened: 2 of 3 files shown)


src/databricks/labs/ucx/hive_metastore/migration_status.py
```diff
@@ -70,7 +70,7 @@ def get_seen_tables(self) -> dict[str, str]:
         seen_tables: dict[str, str] = {}
         for schema in self._iter_schemas():
             try:
-                tables = self._ws.tables.list(catalog_name=schema.catalog_name, schema_name=schema.name)
+                tables = list(self._ws.tables.list(catalog_name=schema.catalog_name, schema_name=schema.name))
```
JCZuurmond, line 73: Why adding this list?

JCZuurmond (changes requested):
The index is accessed directly:
- [here](https://github.com/databrickslabs/ucx/blob/79e228420f9a5e1330c81b2684ef8c487c90ac4d/src/databricks/labs/ucx/hive_metastore/table_migrate.py#L219)
- [here](https://github.com/databrickslabs/ucx/blob/79e228420f9a5e1330c81b2684ef8c487c90ac4d/src/databricks/labs/ucx/hive_metastore/table_migrate.py#L381)

Change those to use the index method on the table migration class.

src/databricks/labs/ucx/hive_metastore/migration_status.py
nfx: because of `yield`. @nkvuong it's best to add comments :)

src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -57,7 +57,7 @@ def __init__(
         self._principal_grants = principal_grants
 
     def index(self):
-        # TODO: remove this method
+        self._migration_status_refresher.reset()
```
nfx, line 60: don't reset it here. we use `.index` in CLI and it has to be executed fast.

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/migration_status.py
JCZuurmond: it is only used once in the code below, so yield should be fine

nkvuong: the SDK throws a NotFound exception if the schema does not exist - the try catch here won't work as the iterator is not consumed yet

nfx (approved):
lgtm

Issue #393, opened 2023-10-06
Prompt user if there's Terraform used for deploying of certain entities


nfx:
Infrastructure things, like storage, credentials, and sometimes grants are generally managed by Terraform. We have to prompt the installer to see if we defer certain parts of automation to TF (or change the TF AST). Potentially, we can also scan the TF AST and state for some quick wins.

PR #1664, opened 2024-05-08
Remove the "Is Terraform used" prompt

JCZuurmond:
## Changes
The outcome of the prompt is not used and might confuse users. (We received feedback that it does confuse an user.) Moreover, in the near future we don't expect to use the Terraform state for ucx migration.

### Linked issues
Was introduced for #393

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


nkvuong (approved):
it makes sense to remove the prompts for now, we can re-introduce them later

PR #1665, opened 2024-05-08
Assert if group name is in cluster policy

JCZuurmond:
## Changes
Assert if group name is in cluster policy

### Linked issues
Does not resolve the issue, but asserts the key to avoid the key error in #1616 

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/tests/integration/test_ext_hms.py
+++ b/tests/integration/test_ext_hms.py
@@ -170,4 +170,5 @@ def test_running_real_assessment_job_ext_hms(
     assert ext_hms_ctx.deployed_workflows.validate_step("assessment")
 
     after = ext_hms_ctx.generic_permissions_support.load_as_dict("cluster-policies", cluster_policy.policy_id)
+    assert ws_group_a.display_name in after, f"Group {ws_group_a.display_name} not found in cluster policy"
     assert after[ws_group_a.display_name] == PermissionLevel.CAN_USE


tests/integration/test_ext_hms.py
```diff
@@ -170,4 +170,5 @@ def test_running_real_assessment_job_ext_hms(
     assert ext_hms_ctx.deployed_workflows.validate_step("assessment")
 
     after = ext_hms_ctx.generic_permissions_support.load_as_dict("cluster-policies", cluster_policy.policy_id)
+    assert ws_group_a.display_name in after, f"Group {ws_group_a.display_name} not found in cluster policy"
     assert after[ws_group_a.display_name] == PermissionLevel.CAN_USE
```
JCZuurmond, line 174: @nkvuong : Do you remember why this assert is there? This does not look like something done during the assessment, but something done with the test setup (line 159 to 163). 

I expect this test to fail in the nightly run when we hit a limit for creating (account) groups and are throttled by Databricks.

nkvuong: this was a copy of `test_running_real_assessment_job`, except that it is set up with external hms. My reading of the test is that it is asserting that the permission is crawled correctly by the assessment job

nkvuong (approved):

PR #1669, opened 2024-05-08
Fix test failing due to groups not being created yet

JCZuurmond:
## Changes
(We think that) sometimes groups are not created before the test run, e.g. [this](https://github.com/databrickslabs/ucx/actions/runs/9002466231?pr=1532) and [that](https://github.com/databrickslabs/ucx/issues/1616). This PR extends the `make_group` with an optional flag to wait for the groups to be created before continuing with the test. To limit impact on performance, we don't wait for the groups to be created by default.  

### Linked issues
Hopefully esolves #1616

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/mixins/fixtures.py
+++ b/src/databricks/labs/ucx/mixins/fixtures.py
@@ -642,7 +642,6 @@ def _make_group(name, cfg, interface, make_random):
         else:
             logger.info(f"Workspace group {group.display_name}: {cfg.host}#setting/accounts/groups/{group.id}")
 
-        # When
         @retried(on=[NotFound], timeout=timedelta(minutes=1))
         def _wait_for_provisioning() -> None:
             interface.get(group.id)
--- a/src/databricks/labs/ucx/mixins/fixtures.py
+++ b/src/databricks/labs/ucx/mixins/fixtures.py
@@ -625,6 +625,7 @@ def _make_group(name, cfg, interface, make_random):
         roles: list[str] | None = None,
         entitlements: list[str] | None = None,
         display_name: str | None = None,
+        wait_for_provisioning: bool = False,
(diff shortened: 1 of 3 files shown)


src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -638,6 +639,14 @@ def create(
             logger.info(f"Account group {group.display_name}: {cfg.host}/users/groups/{group.id}/members")
         else:
             logger.info(f"Workspace group {group.display_name}: {cfg.host}#setting/accounts/groups/{group.id}")
+
+        @retried(on=[NotFound], timeout=timedelta(minutes=1))
```
JCZuurmond, line 643: @nkvuong : What do you think a reasonable time should be? One, two, five minutes?

nkvuong: Two minutes should be enough for most situations

tests/integration/conftest.py
```diff
@@ -569,7 +569,7 @@ def __init__(  # pylint: disable=too-many-arguments
         self._make_acc_group = make_acc_group_fixture
         self._make_user = make_user_fixture
 
-    def make_ucx_group(self, workspace_group_name=None, account_group_name=None):
+    def make_ucx_group(self, workspace_group_name=None, account_group_name=None, wait_for_provisioning=False):
```
nkvuong, line 572: we should always wait for the group to be provisioned right?

nkvuong: that makes sense to me 👍

nkvuong (approved):

PR #1679, opened 2024-05-10
Improved error handling for all queries executed during table migration

### LINKED ISSUE #1676: [BUG]: Error when running migrate-tables workflow. Views in an invalid state break the migrate-views task

### Current Behavior

Workflow: migrate-tables
task: migrate_views
Error: When the View is in an Invalid state in the hive_metastore and all dependencies have been successfully migrated, the "migration_view" task attempts to create the view in the target database, however, the view cannot be compiled due to a structural error, such as : referencing a column that does not exist in a table.

image

image

### Expected Behavior

I was expecting that the process would not fail completely.
Maybe a better approach would be log a message like: "The view XXXX cannot be migrated."

### Steps To Reproduce

_No response_

### Cloud

Azure

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

nkvuong:
## Changes

- As we are migrating multiple tables & views in parallel, any error in SQL execution should not fail the workflow. Added error handling to all `sql_backend.execute` commands

### Linked issues

Resolves #1676

### Tests

- [x] manually tested
- [x] added unit tests

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/table_migrate.py
+++ b/src/databricks/labs/ucx/hive_metastore/table_migrate.py
@@ -217,7 +217,7 @@ class TablesMigrator:
                 src_view.src.sql_alter_from(src_view.rule.as_uc_table_key, self._ws.get_workspace_id())
             )
         except DatabricksError as e:
-            logger.warning(f"Failed to migrate view {src_view.src.key}: {e}")
+            logger.warning(f"Failed to migrate view {src_view.src.key} to {src_view.rule.as_uc_table_key}: {e}")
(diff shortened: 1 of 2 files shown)


JCZuurmond (changes requested):
We should adapt the ViewSequencer to handle the situation of a view not being migrated. 

For example, it might confuse the users when they see a `View {src_view.src.key} is not supported for migration` message for the views that are dependent on an invalid view as it does not explain why they are not supported for migration

nfx (approved):
lgtm

Issue #1642, opened 2024-05-06
Workflow linter to analyse PyPI library


nfx:
Would share significant codebase with:
- https://github.com/databrickslabs/ucx/issues/1640

PR #1692, opened 2024-05-14
Add pip cell installs to dependency graph

JCZuurmond:
## Changes
Logic for registering import installed in a pip cell

### Linked issues
#1642

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -44,13 +44,6 @@ def dependency(self):
     def path(self):
         return self._dependency.path
 
-    def register_library(self, name: str) -> MaybeGraph:
```
JCZuurmond, line 47: @ericvergnaud : For now, shall we stick with the `import`, not have the `library`, and add it to the `DependencyGraph` right away without requiring an `import ...`

ericvergnaud: mmm... I suspect that would only increase confusion when discovering an already complex framework...

PR #1715, opened 2024-05-17
Automatically retrieved `aws_account_id` from aws profile instead of prompting

nkvuong:
## Changes
- Automatically retrieved `aws_account_id` from aws profile instead of prompting users for this.
- Documented the optional kms-key flag for create roles
cc: @mohanab-db 

### Linked issues

Resolves #1714

### Tests

- [x] manually tested
- [x] verified on staging environment (screenshot attached)

PR #1718, opened 2024-05-19
Getting aws account id from the profile.

### LINKED ISSUE #1714: [BUG]: `aws_account_id` is None when creating uber principal

### Current Behavior

I am getting an error on running `databricks labs ucx create-uber-principal` as the `aws_account_id` is being set as None. I see in the code that this value is fetched from `named_parameters`. How to pass the `aws_account_id` as a `named_parameter` via CLI?

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

FastLee:
closes #1714
Added lookup for AWS account id from the AWS profile.

### DIFF
--- a/src/databricks/labs/ucx/contexts/workspace_cli.py
+++ b/src/databricks/labs/ucx/contexts/workspace_cli.py
@@ -146,12 +146,15 @@ class WorkspaceContext(CliContext):
 
     @cached_property
     def aws_resource_permissions(self):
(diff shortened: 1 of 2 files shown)


nkvuong:
I've already opened the PR #1715 :)

PR #1709, opened 2024-05-17
WhitelistResolver must be invoked before other import resolvers

ericvergnaud:
## Changes
Ensure that WhitelistResolver, if present, is always imported first

### Linked issues
Resolves #1707 

### Functionality 

- [ ] added relevant user documentation
- [ ] added new CLI command
- [ ] modified existing command: `databricks labs ucx ...`
- [ ] added a new workflow
- [ ] modified existing workflow: `...`
- [ ] added a new table
- [ ] modified existing table: `...`

### Tests
- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] verified on staging environment (screenshot attached)

PR #1746, opened 2024-05-22
[internal] Simplify import resolvers

### LINKED ISSUE #1707: [BUG]: Linting raises library-not-found problem when import `type_extensions`

### Current Behavior

We raise a library-not-found problem when import type_extentions

```
if TYPE_CHECKING:
    from typing_extensions import ParamSpec

    P = ParamSpec("P")
```

### Expected Behavior

No problem should be raised. Maybe we should white list this library

### Steps To Reproduce

1. Create unit test for above code snippet

### Cloud

Azure

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

ericvergnaud:
## Changes
Remove chain of import resolvers in favor of a single resolver.

Supersedes #1709 

Resolves #1707

### DIFF
--- a/src/databricks/labs/ucx/source_code/files.py
+++ b/src/databricks/labs/ucx/source_code/files.py
@@ -125,7 +125,7 @@ class LocalFileResolver(BaseImportResolver, BaseFileResolver):
         return MaybeDependency(None, [problem])
 
     def resolve_import(self, path_lookup: PathLookup, name: str) -> MaybeDependency:
-        maybe = self._whitelist_resolver.resolve_import(path_lookup, name)
+        maybe = self._whitelist_resolver.resolve_import(name)
         if maybe is not None:
             return maybe
         maybe = self._resolve_import(path_lookup, name)
--- a/src/databricks/labs/ucx/source_code/files.py
+++ b/src/databricks/labs/ucx/source_code/files.py
@@ -4,6 +4,7 @@ import logging
 from pathlib import Path
 
 from databricks.labs.ucx.source_code.path_lookup import PathLookup
+from databricks.labs.ucx.source_code.whitelist import Whitelist, WhitelistResolver
 from databricks.sdk.service.workspace import Language
 
 from databricks.labs.ucx.source_code.languages import Languages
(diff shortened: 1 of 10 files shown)


src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -277,21 +277,9 @@ def _fail(code: str, message: str) -> MaybeDependency:
 
 class BaseImportResolver(abc.ABC):
```
nfx, line 278: redundant: we no longer need `BaseImportResolver`

src/databricks/labs/ucx/source_code/files.py
```diff
@@ -111,13 +112,11 @@ def __repr__(self):
 
 class LocalFileResolver(BaseImportResolver, BaseFileResolver):
```
nfx, line 113:
```suggestion
class ImportResolver(BaseFileResolver):
```
and get rid of BaseImportResolver.

src/databricks/labs/ucx/source_code/whitelist.py
```diff
@@ -22,16 +20,13 @@
...
+        super().__init__()
         self._whitelist = whitelist
 
-    def with_next_resolver(self, resolver: BaseImportResolver) -> BaseImportResolver:
-        return WhitelistResolver(self._whitelist, resolver)
-
-    def resolve_import(self, path_lookup: PathLookup, name: str) -> MaybeDependency:
+    def resolve_import(self, name: str) -> MaybeDependency | None:
         # TODO attach compatibility to dependency, see https://github.com/databrickslabs/ucx/issues/1382
         compatibility = self._whitelist.compatibility(name)
```
nfx, line 31: move the body of this method as private method to `ImportResolver` class and get rid of `WhitelistResolver`.

nfx (changes requested):
simplify even further.

src/databricks/labs/ucx/source_code/whitelist.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/files.py
ericvergnaud: renamed to ImportFileResolver

PR #1770, opened 2024-05-24
Changes to create access connector if secret is not found for spn

### LINKED ISSUE #892: [FEATURE]: Create spn secret if it does not exist

### Problem statement

When migrating Azure SPNs to UC storage credentials, SPNs without client secrets are not migrated

### Proposed Solution

Attempt to create a short-lived secret for SPNs without client secrets. Log any that errors

Relevant issue: https://github.com/databrickslabs/ucx/issues/339

### Additional Context

_No response_

mohanab-db:
## Changes
Changes to create access connector if secret is not found for Azure SPN

### Linked issues

Resolves #892 

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


tests/unit/azure/test_credentials.py
```diff
@@ -542,3 +544,68 @@ def test_create_access_connectors_for_storage_accounts(sp_migration):
     assert len(validation_results) == 1
     assert validation_results[0].name.startswith("ac")
     assert len(validation_results[0].failures) == 0
+
+
+@pytest.fixture
+def sp_migration_1(installation, credential_manager):
```
nfx, line 550: Don't create a fixture for something that is used only once, as it makes it harder to debug. Rewrite

tests/unit/azure/test_credentials.py
```diff
@@ -542,3 +544,68 @@ def test_create_access_connectors_for_storage_accounts(sp_migration):
...
+    ws = create_autospec(WorkspaceClient)
+    ws.secrets.get_secret.return_value = GetSecretResponse(
+        value=base64.b64encode("hello world".encode("utf-8")).decode("utf-8")
+    )
+
+    mock_get_secret = Mock()
```
nfx, line 556: Technically, you don't need this variable. We're also preventing untyped mocks, because it's more difficult to refactor

nfx (changes requested):

JCZuurmond (changes requested):
Hi @mohanab-db, thank you for opening another PR 😄 

Do I understand correctly that you want the user to be able to use access connectors to access the storage accounts for which the service principal can not be found? Is **precisely** this a request from a customer? 

I would prefer if we nudge the user to using access connectors for **all** storage accounts. We could warn to use access connectors when a SP secret can not be fetched. Moreover, it would simplify your code changes as it does not require to keep track of the `set`s you introduced

nfx:
@mohanab-db @JCZuurmond What remains to be done here?

JCZuurmond:
@mohanab-db : Could you reply to my comment above?

mohanab-db:
Hi @JCZuurmond - Sorry for the delayed response. This is not a request from my customer. I saw that there is this open P2 feature request https://github.com/databrickslabs/ucx/issues/892 where it says that only SPs which have passwords saved will be migrated to a storage credential. There are comments here saying that if the password is not found then a managed identity should be created and used to create a Storage Credential. I committed my changes for this, since I had some time and wanted to contribute to the project. As far as I know, there is already a feature to create access connectors for all storage credentials. I can close the PR if you think that this is not needed.

JCZuurmond:
Hi @mohanab-db, thanks for your reply - even though it is late 😋

I see, this indeed implements what that PBI describes. Still, I would prefer to nudge the users to use access connectors for **all** storage accounts. Or mention that if they want to migrate a service principal, that they should create a secret for it which can be found by ucx. 

Could you trim the PR down to the log messages that telling the user this? I will add comments in the code

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -183,7 +183,9 @@ def __init__(
         self._sp_crawler = service_principal_crawler
         self._storage_credential_manager = storage_credential_manager
 
-    def _fetch_client_secret(self, sp_list: list[StoragePermissionMapping]) -> list[ServicePrincipalMigrationInfo]:
+    def _fetch_client_secret(
+        self, sp_list: list[StoragePermissionMapping], storage_acc_set: set[str] | None = None
```
JCZuurmond, line 187: revert the signature

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -204,9 +207,22 @@ def _fetch_client_secret(self, sp_list: list[StoragePermissionMapping]) -> list[
...
+                    logger.info(
+                        f"Cannot fetch the service principal client_secret for {sp_info.application_id}. "
+                        f"An access connector will be used for this storage account {sp_info.storage_account}."
+                    )
+                    storage_acc_no_secret.add(sp_info.storage_account)
             else:
                 logger.info(
-                    f"Cannot fetch the service principal client_secret for {sp_info.application_id}. "
+                    f"Storage account information missing for service principal {sp_info.application_id}. "
                     f"This service principal will be skipped for migration"
```
JCZuurmond, line 226: explain how to resolve this:
Rerun migration,
i) Confirm to create access connector instead
ii) After creating SP secret

JCZuurmond (changes requested):
See comments

src/databricks/labs/ucx/azure/credentials.py
mohanab-db: @JCZuurmond  This 'else' block would only be reached if a service principal (SP) doesn't have any storage account (SA) associated with it. Which I am thinking will never happen. If it does, then the SP is of no use, as there is no SA that needs to be connected with a storage credential in this case. I can change the message saying that "skipping this SP as there is no SA associated to it".

The logic is like this -- 
1. First the user will be prompted if they would like create an access connector with a managed identity for each storage account (line 325, existing functionality).
2. If they say yes, then all access connectors will be created (line 297, existing functionality) and a set will be created to keep a track of the SAs which are already configured with access connector (line 312, change in PR).
3. Next it will check the list of SPs in azure_storage_account_info.csv and azure_service_principals table and check if a secret is found (line 206, existing functionality). If not, it will get the SA associated with the SP, check that the access connector is not already added, and then list all new SAs for which access connector needs to be created (line 222, change in PR).
4. It will then prompt the user to accept creating of storage credentials from the SP and access connectors (line 354, change in PR).
5. Create the storage credentials (line 345, existing, and line 358, new).

I can revery back all of my changes if you think it is not needed as we already have a functionality to create access connector for all storage accounts (line 325).

JCZuurmond: > This 'else' block would only be reached if a service principal (SP) doesn't have any storage account (SA) associated with it. Which I am thinking will never happen. If it does, then the SP is of no use, as there is no SA that needs to be connected with a storage credential in this case. I can change the message saying that "skipping this SP as there is no SA associated to it".

Agreed. If I read the code correctly, the following would be reached if the SP has a storage account associated with it, but its credentials are not found:

``` python
                    logger.info(
                        f"Cannot fetch the service principal client_secret for {sp_info.application_id}. "
                        f"An access connector will be used for this storage account {sp_info.storage_account}."
                    )
```

That message can be updated to tell the user to create access connectors instead OR create the SP credentials for ucx to find

> I can revery back all of my changes if you think it is not needed as we already have a functionality to create access connector for all storage accounts (line 325).

yes, to keep the code flow more unidirectional, I prefer to separate the access connector creation from the service principal migration (as much as possible)

mohanab-db: Thanks, closing this PR as it is no longer needed.

PR #1815, opened 2024-05-31
Handle legacy ACL `DENY` permission in group migration

### LINKED ISSUE #1803: [FEATURE]: legacy Deny ACL is not handled

### Problem statement

In our legacy table ACLs, there are `DENY` operations. However, the current UCX version 0.24.0 only supports `GRANT` and `REVOKE` operations. This discrepancy causes errors on `apply_permissions_to_account_groups` step when running the Group Migration job.

The following error occurs:

```
Unknown: Action Unknown ActionType DENIED_CREATE cannot be granted on database `xxxx`
```

On our HMS there are 6 denial cases.

- `DENIED_READ_METADATA`
- `DENIED_MODIFY`
- `DENIED_CREATE_NAMED_FUNCTION`
- `DENIED_SELECT`
- `DENIED_USAGE`
- `DENIED_CREATE`

This issue is blocking the successful execution of Group Migration tasks and needs to be addressed to ensure smooth migration and permission application processes.

### Proposed Solution

I think it can be resolved just checking `DENIED_` prefix exists on actions.

```py
def hive_grant_sql(self) -> list[str]:
        object_type, object_key = self.this_type_and_key()
        # See https://docs.databricks.com/en/sql/language-manual/security-grant.html
        statements = []
        actions = self.action_type.split(", ")
        if "OWN" in actions:
            actions.remove("OWN")
            statements.append(self._set_owner_sql(object_type, object_key))
        if actions:
            for action in actions:
                if "DENIED" in action:
                    deny_action = action.replace("DENIED_", "")
                    statements.append(self._apply_deny_sql(deny_action, object_type, object_key))
                else:
                    statements.append(self._apply_grant_sql(action, object_type, object_key))
        return statements

...

def _apply_deny_sql(self, action_type, object_type, object_key):
        return f"DENY {action_type} ON {object_type} {escape_sql_identifier(object_key)} TO `{self.principal}`"
```

Please refer my commit on forked repo.

https://github.com/Seokyun-Ha/ucx/commit/d5b26c9d7aa55b8d30422924bc560e2a92e962ea

Thanks! 😄 

### Additional Context

_No response_

nkvuong:
## Changes

- In our legacy table ACLs, there are DENY operations. These permissions are shown with `DENIED` prefix. Apply them correctly when applying permissions for group migrations

### Linked issues

Resolves #1803

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/grants.py
+++ b/src/databricks/labs/ucx/hive_metastore/grants.py
@@ -137,6 +137,7 @@ class Grant:
     def _apply_grant_sql(self, action_type, object_type, object_key):
         if "DENIED" in action_type:
             action_type = action_type.replace("DENIED_", "")
+            # need to wrap the action type in backticks to avoid syntax errors with DENY SELECT
             return f"DENY `{action_type}` ON {object_type} {escape_sql_identifier(object_key)} TO `{self.principal}`"
         return f"GRANT {action_type} ON {object_type} {escape_sql_identifier(object_key)} TO `{self.principal}`"
 
--- a/src/databricks/labs/ucx/hive_metastore/grants.py
+++ b/src/databricks/labs/ucx/hive_metastore/grants.py
@@ -135,6 +135,9 @@ class Grant:
         return f"ALTER {object_type} {escape_sql_identifier(object_key)} OWNER TO `{self.principal}`"
 
     def _apply_grant_sql(self, action_type, object_type, object_key):
+        if "DENIED" in action_type:
+            action_type = action_type.replace("DENIED_", "")
+            return f"DENY `{action_type}` ON {object_type} {escape_sql_identifier(object_key)} TO `{self.principal}`"
         return f"GRANT {action_type} ON {object_type} {escape_sql_identifier(object_key)} TO `{self.principal}`"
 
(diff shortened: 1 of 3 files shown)


src/databricks/labs/ucx/hive_metastore/grants.py
```diff
@@ -135,6 +135,9 @@ def _set_owner_sql(self, object_type, object_key):
         return f"ALTER {object_type} {escape_sql_identifier(object_key)} OWNER TO `{self.principal}`"
 
     def _apply_grant_sql(self, action_type, object_type, object_key):
+        if "DENIED" in action_type:
+            action_type = action_type.replace("DENIED_", "")
+            return f"DENY `{action_type}` ON {object_type} {escape_sql_identifier(object_key)} TO `{self.principal}`"
```
nkvuong, line 140: we need to wrap the action type in DENY, because `DENY SELECT` throws an error. ``DENY `SELECT` `` works

nfx: Why it's not a code comment? 😉 add, otherwise someone would remove backticks

nfx (approved):

PR #1816, opened 2024-05-31
Update linting checks for Spark table methods

### LINKED ISSUE #1779: [BUG]: `register` triggers too many false positives

### Current Behavior

Currently, if we lint local code or add new known libraries, a lot of false positives appear in reports:
```
      {
        "code": "table-migrate",
        "message": "Can't migrate 'register' because its table name argument is not a constant"
      },
```

### Expected Behavior

Improve detection logic to trigger warnings only on spark-related code

### Steps To Reproduce

in ucx codebase:
1. `rm src/databricks/labs/ucx/source_code/known.json`
2. `make known`
3. view the contents of known.json file and seek false positives

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

asnare:
## Changes

This PR:

- Refactors most of the pyspark linter unit tests to be functional tests using sample files.
- Updates the sample for each detected method to use code that is valid Spark code.
- Removes a spurious matcher for `.table()` on Spark Catalog; this doesn't exist.
- Expands the pyspark linter tests to check the various things for every method instead of just a few.
- Marks some false-positive warnings as such.
- Notes some warnings that should be present but aren't yet supported.

### Linked issues

- Relates to #1779

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [X] added unit tests
- [ ] added integration tests


tests/unit/source_code/linters/test_pyspark.py
```diff
@@ -209,10 +209,10 @@ def test_spark_table_variable_arg(migration_index):
     ftf = FromTable(migration_index, CurrentSessionState())
     sqf = SparkSql(ftf, migration_index)
     old_code = """
-spark.read.csv("s3://bucket/path")
+df = spark.read.csv("s3://bucket/path")
 for i in range(10):
-    df = spark.saveAsTable(name)
-    do_stuff_with_df(df)
+    df = do_stuff_with_df(df)
+    df = spark.write.saveAsTable(name)
```
nfx, line 215: This isn't valid spark code afaik

asnare: Indeed; fixed.

tests/unit/source_code/linters/test_pyspark.py
```diff
@@ -98,25 +98,33 @@ def test_spark_sql_match_named(migration_index):
     "recoverPartitions",
     "refreshTable",
     "uncacheTable",
-    "table",
-    "insertInto",
-    "saveAsTable",
-]
+}
+TABLE_METHOD_NAMES = {
```
nfx, line 102: Please also add some tests to prevent false positives, eg "foo.register()"

asnare:
Depends on #1823 to resolve some false positives.

tests/unit/source_code/samples/functional/pyspark/dataframe-write-save-as-table.py
```diff
@@ -0,0 +1,13 @@
+# ucx[direct-filesystem-access:+1:0:+1:0] The use of direct filesystem references is deprecated: s3://bucket/path
+df = spark.read.csv("s3://bucket/path")
+for i in range(10):
```
nfx, line 3: i wonder if we want to have a loop in every file. some files might get functions. or methods. etc.

tests/unit/source_code/samples/functional/pyspark/dataframe-write-insert-into.py
```diff
@@ -0,0 +1,13 @@
...
+
+    ## Check a literal reference to a known table that is migrated.
+    # ucx[table-migrate:+1:0:+1:0] Table old.things is migrated to brand.new.stuff in Unity Catalog
+    df.write.insertInto("old.things")
+
+    ## Check a literal reference to an unknown table (that is not migrated); we expect no warning.
+    df.write.insertInto("table.we.know.nothing.about")
+
+    ## Check that a call with too many positional arguments is ignored as (presumably) something else; we expect no warning.
+    df.write.insertInto("old.things", None, "extra-argument")
```
nfx, line 13: looks like a bug in UCX analysis

tests/unit/source_code/samples/functional/pyspark/dataframe-write-save-as-table.py
```diff
@@ -0,0 +1,13 @@
...
+
+    ## Check a literal reference to a known table that is migrated.
+    # ucx[table-migrate:+1:0:+1:0] Table old.things is migrated to brand.new.stuff in Unity Catalog
+    df.write.format("delta").saveAsTable("old.things")
+
+    ## Check a literal reference to an unknown table (that is not migrated); we expect no warning.
+    df.write.format("delta").saveAsTable("table.we.know.nothing.about")
+
+    ## Check that a call with too many positional arguments is ignored as (presumably) something else; we expect no warning.
+    df.write.format("delta").saveAsTable("old.things", None, None, None, "extra-argument")
```
nfx, line 13: looks like you found a bug :)

tests/unit/source_code/samples/functional/pyspark/spark-catalog-cache-table.py
```diff
@@ -0,0 +1,13 @@
+# ucx[direct-filesystem-access:+1:0:+1:0] The use of direct filesystem references is deprecated: s3://bucket/path
+spark.read.csv("s3://bucket/path")
+for i in range(10):
+
```
nfx, line 4:
```suggestion
```
irrelevant

tests/unit/source_code/samples/functional/pyspark/spark-catalog-cache-table.py
```diff
@@ -0,0 +1,13 @@
...
+    ## Check a literal reference to a known table that is migrated.
```
nfx, line 5: i'd suggest to put these files into `spark.catalog` folder or something like it - .e.g. if there are more than 5 files in the same group, they may deserve own folder

tests/unit/source_code/samples/functional/pyspark/dataframe-write-save-as-table.py
asnare: Indeed, some variety does no harm.

tests/unit/source_code/samples/functional/pyspark/dataframe-write-insert-into.py
asnare: As written, it's intentional behaviour (covered by the `Matcher` `max_arg` property). 

There are two issues with this behaviour that I've got in the back of my mind:

 - I'm not sure what we need to cope with in terms of future compatibility, but the Spark APIs tend to add new positional arguments over time with default values. The current `max_args` spec instantly means that future versions of these APIs won't match. This drift has already happened over the Spark 3.x evolution, and it maintains source/binary compatibility (on the Python side) so I would expect this to continue. As such I think it might be worth defaulting to treating methods as having an implicit `*args` even if they don't, with `max_args` limited to cases where it's needed to disambiguate or eliminate false positives.
 - The current matcher specs don't really cover the way python arguments can be specified. Some limitations include missing `*args` (workaround: max=1000) and `**kwargs` (no workaround) support. It also doesn't understand that all positional arguments have a name, and often these are used. (Style guides and linting rulesets often mandate using named syntax for int/boolean parameters, so I expect to find some quirks in this area that we need to address.)

tests/unit/source_code/linters/test_pyspark.py
asnare: Done in the functional tests.

nfx (approved):
Lgtm to unblock

PR #1820, opened 2024-05-31
Improve make known devex

asnare:
## Changes

Improve the devex around `known.json`:

 - Improve logging so that it's clearer what it's doing, in particular when there aren't any new distributions to scan.
 - Handle building from scratch when `known.json` has been removed.
 - Only rewrite the known.json file if it has changed.
 - Use `importlib` instead of `pkgutil` when loading `known.json` (the latter is deprecated from python 3.12).

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [X] manually tested
- [X] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/source_code/known.py
+++ b/src/databricks/labs/ucx/source_code/known.py
@@ -2,9 +2,9 @@ from __future__ import annotations
 
 import collections
 import email
+import importlib.resources
 import json
 import logging
-import pkgutil
 import re
 import sys
 from dataclasses import dataclass
@@ -56,13 +56,15 @@ class Whitelist:
             self._module_problems[name] = []
 
     @staticmethod
-    def _get_known():
-        module = __name__
-        if __name__ == "__main__":  # code path for UCX developers invoking `make known`
-            module = "databricks.labs.ucx.source_code.known"
-        # load known.json from package data, because we may want to use zipapp packaging
-        data = pkgutil.get_data(module, "known.json")
-        return json.loads(data)
+    def _get_known() -> dict[str, dict[str, list[dict[str, str]]]]:
+        this_module = __package__
+        if this_module is None:
+            # When this file is invoked directly as a script, __package__ is None.
+            # Note: can get confused if this package is _also_ installed in site-packages.
+            from databricks.labs.ucx import source_code as this_module  # pylint: disable=import-outside-toplevel
+        known_json = importlib.resources.files(this_module) / "known.json"
+        with known_json.open("rb") as f:
+            return json.load(f)
 
     def module_compatibility(self, name: str) -> Compatibility:
         if not name:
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/source_code/known.py
```diff
@@ -56,13 +56,22 @@ def __init__(self):
...
-        return json.loads(data)
+    def _get_known() -> dict[str, dict[str, list[dict[str, str]]]]:
+        """Loads the set of known distributions (and false positives), included with this module.
+
+        Returns:
+            A dictionary (keyed by distribution name) containing the known problems for each module.
+        Raises:
+            FileNotFoundError: If the builtin set of known distribution is missing.
+        """
+        this_module = __package__
```
nfx, line 67: 1. does it work for `make known`? 
2. does it work when the script is debugged directly?

src/databricks/labs/ucx/source_code/known.py
```diff
@@ -56,13 +56,22 @@ def __init__(self):
...
+        if this_module is None:
+            # When this file is invoked directly as a script, __package__ is None.
+            # Note: can get confused if this package is _also_ installed in site-packages.
+            from databricks.labs.ucx import source_code as this_module  # pylint: disable=import-outside-toplevel
+        known_json = importlib.resources.files(this_module) / "known.json"
```
nfx, line 72: why do we need this change? does it work with zipapps? does it work with PEX setup? `importlib.resources` had issues when invoked from zip archives - we don't know yet if UCX will be deployed to Databricks runtime as ZIP or not.

src/databricks/labs/ucx/source_code/known.py
```diff
@@ -103,13 +112,24 @@ def _cleanup_name(name):
     def rebuild(cls, root: Path):
         """rebuild the known.json file by analyzing the source code of installed libraries. Invoked by `make known`."""
         path_lookup = PathLookup.from_sys_path(root)
-        known_distributions = cls._get_known()
+        try:
```
nfx, line 115: this is a good change :) let's limit this PR to it.

nfx (changes requested):

src/databricks/labs/ucx/source_code/known.py
asnare: Yes, and yes.

src/databricks/labs/ucx/source_code/known.py
asnare: We don't need this change; `pkgutil` is just something I try to avoid, partly because it's deprecated, partly because in ancient times I also had issues it.

Given your reservations and complexity of testing I'll revert this.

src/databricks/labs/ucx/source_code/known.py
asnare: Thanks; will do.

nfx (approved):
lgtm

PR #1824, opened 2024-06-03
Deduplicate errors emitted by Spark Connect linter

vsevolodstep-db:
## Changes
This PR adds deduplication of errors emitted by Spark Connect linter. It also adds more functional tests using the new framework

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests
- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/source_code/linters/spark_connect.py
+++ b/src/databricks/labs/ucx/source_code/linters/spark_connect.py
@@ -22,6 +22,15 @@ class SharedClusterMatcher:
     def lint(self, node: ast.AST) -> Iterator[Advice]:
         pass
 
+    def lint_tree(self, tree: ast.AST) -> Iterator[Advice]:
+        reported_locations = set()
+        for node in ast.walk(tree):
+            for advice in self.lint(node):
+                loc = (advice.start_line, advice.start_col)
+                if loc not in reported_locations:
+                    reported_locations.add(loc)
+                    yield advice
+
 
 class JvmAccessMatcher(SharedClusterMatcher):
     _FIELDS = [
@@ -195,6 +204,5 @@ class SparkConnectLinter(Linter):
 
     def lint(self, code: str) -> Iterator[Advice]:
         tree = ast.parse(code)
-        for node in ast.walk(tree):
-            for matcher in self._matchers:
-                yield from matcher.lint(node)
+        for matcher in self._matchers:
+            yield from matcher.lint_tree(tree)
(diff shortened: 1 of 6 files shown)


tests/unit/source_code/test_functional.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, path: Path):
 
     def verify(self):
         expected_problems = list(self._expected_problems())
-        actual_problems = list(self._lint())
+        actual_problems = sorted(list(self._lint()), key=lambda a: (a.start_line, a.start_col))
```
asnare, line 50: Sometimes we generate more than one piece of advice with the same starting position. To be stable I think we need to:
1. Include the `code` and the `message` in the sorting key.
2. Also sort the expected problems by the same criteria (although this is less important because the order in the samples can be adjusted.)

tests/unit/source_code/samples/functional/spark-logging.py
```diff
@@ -0,0 +1,15 @@
+# ucx[legacy-context-in-shared-clusters:+1:0:+1:14] sc is not supported on UC Shared Clusters. Rewrite it using spark
+# ucx[spark-logging-in-shared-clusters:+1:0:+1:22] Cannot set Spark log level directly from code on UC Shared Clusters. Remove the call and set the cluster spark conf 'spark.log.level' instead
+sc.setLogLevel("INFO")
+setLogLevel("WARN")
```
nfx, line 4:
```suggestion
# ucx[legacy-context-in-shared-clusters:+1:0:+1:14] sc is not supported on UC Shared Clusters. Rewrite it using spark
sc.setLogLevel("INFO")

# ucx[spark-logging-in-shared-clusters:+1:0:+1:22] Cannot set Spark log level directly from code on UC Shared Clusters. Remove the call and set the cluster spark conf 'spark.log.level' instead
setLogLevel("WARN")
```

annotated comments have to be right above the line they annotate. at the moment the framework doesn't check the line offsets, as there's a bug in notebook cell parsing.

src/databricks/labs/ucx/source_code/linters/spark_connect.py
```diff
@@ -22,6 +22,15 @@ def _cluster_type_str(self) -> str:
...
         pass
 
+    def lint_tree(self, tree: ast.AST) -> Iterator[Advice]:
+        reported_locations = set()
+        for node in ast.walk(tree):
+            for advice in self.lint(node):
+                loc = (advice.start_line, advice.start_col)
+                if loc not in reported_locations:
+                    reported_locations.add(loc)
+                    yield advice
```
nfx, line 32:
```suggestion
                if loc in reported_locations:
                    continue
                reported_locations.add(loc)
                yield advice
```
break the loops early for a more readable flow.

nfx (approved):
few nits, otherwise lgtm

tests/unit/source_code/samples/functional/spark-logging.py
vsevolodstep-db: This one is actually correct: the first code line should emit two warnings, while the second line is a valid code (assuming that `setLogLevel` is defined as a user function somewhere).

One thing worth considering is whether we want to show both these errors or only the more specific one (`spark-logging-in-shared-clusters` in this case). Right now I feel that both errors are useful, but we can change it later

PR #1843, opened 2024-06-05
Check offsets of expected warnings in functional linter tests

asnare:
## Changes

This PR is stacked on #1855 and includes the following changes:

 - Updates the fixtures for the functional linter tests so that the expected offsets are correct.
 - Updates the functional tests to also validate the region offsets are as expected.
 - The test itself has also been updated so that test failures are reported as lists of missing/unexpected comments in the form used by the functional tests.

### Tests

- [ ] manually tested
- [X] added unit tests
- [ ] added integration tests


nfx (commented):
👍🏻

nfx (approved):
lgtm

PR #1848, opened 2024-06-05
Fix flaky test test_instance_pools

pritishpai:
## Changes
The test did not fail with the timeout error in the original issue. Added a retry on timeout error to reduce flakiness.

### Linked issues
Adds retried to avoid #1829

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/tests/integration/workspace_access/test_generic.py
+++ b/tests/integration/workspace_access/test_generic.py
@@ -26,7 +26,7 @@ from . import apply_tasks
 
 
 @pytest.mark.parametrize("is_experimental", [True, False])
-@retried(on=[NotFound], timeout=timedelta(minutes=3))
+@retried(on=[NotFound, TimeoutError], timeout=timedelta(minutes=3))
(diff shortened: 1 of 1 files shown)


tests/integration/workspace_access/test_generic.py
```diff
@@ -26,7 +26,7 @@
 
 
 @pytest.mark.parametrize("is_experimental", [True, False])
-@retried(on=[NotFound], timeout=timedelta(minutes=3))
+@retried(on=[NotFound, TimeoutError], timeout=timedelta(minutes=3))
```
JCZuurmond, line 29: For which timeout is this intended? This is looks like a way to extend the timeout, which would like to avoid

JCZuurmond (changes requested):
See comment

PR #1853, opened 2024-06-06
Complete support for pip install command

### LINKED ISSUE #1203: [FEATURE]: Detect the usage of external OSS and private libraries and analyse them for UC compatibility

### Problem statement

Databricks supports a variety of library sources - https://docs.databricks.com/en/libraries/index.html:
image

### Proposed Solution

Analyse the code in those dependent libraries, regardless of their location.

### Additional Context

_No response_

JCZuurmond:
## Changes
Complete support for pip install command by passing all pip install arguments to install the libaries.

### Linked issues

Resolves #1203

### Functionality 

- [x] modified existing command: `databricks labs ucx lint-local-code`
- [x] added a new workflow
- [x] modified existing workflow: `experimental-workflow-linter`

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests
- [x] Add integration test for pip install with `--index-url`


src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -261,7 +265,9 @@ def __repr__(self):
 
 class LibraryResolver(abc.ABC):
     @abc.abstractmethod
-    def register_library(self, path_lookup: PathLookup, library: Path) -> list[DependencyProblem]:
+    def register_library(
+        self, path_lookup: PathLookup, *libraries: str, installation_arguments: list[str] | None = None
```
JCZuurmond, line 269: There are two changes in this API:
1. Multiple libraries can be installed at once. Motivation: with pip magic multiple installation can be installed at once. To remain as close as possible to the installation command defined as possible, thus minimizing the chance of mimicking the install incorrectly, we support installing multiple libraries at once using the installation arguments.
2. libraries are strings. Motivatioin: i) practical reason, when a library is defined as a relative path "./path/to/lib.whl" a conversion to `Path` loses the leading `./`, which makes that we can not replace the relatives anymore when resolving them with the path lookup. ii) here, libraries are strings, e.g. when `pip install pytest`, only later they are maybe a path when resolving them with the path lookup

src/databricks/labs/ucx/source_code/notebooks/cells.py
```diff
@@ -2,12 +2,17 @@
...
 import logging
 import re
+import shlex
 from abc import ABC, abstractmethod
 from ast import parse as parse_python
 from enum import Enum
 from pathlib import Path
 from sqlglot import parse as parse_sql, ParseError as SQLParseError
 
+from pip._internal.cli.main_parser import parse_command
```
JCZuurmond, line 12: Note this is introduces a connection with pip internals, which introduce a (higher) risk of breakage. However, I still introduced it because the reusing pips `parse_command` is preferred over reimplementing. Also, I tried to minimize the number of dependencies on pip internals.

src/databricks/labs/ucx/source_code/notebooks/cells.py
```diff
@@ -183,16 +188,37 @@ def language(self):
...
+        except PipError as e:
+            return [DependencyProblem("library-install-failed", str(e))]
+        if cmd_name != "install":
+            return [DependencyProblem("library-install-failed", f"Unsupported %pip command: {cmd_name}")]
+        if len(cmd_args) == 0:
+            return [DependencyProblem("library-install-failed", "Missing arguments in '%pip install'")]
+        install = InstallCommand(name="install", summary="Install packages.")
+        _, args = install.parse_args(cmd_args)
+        # Installation arguments are needed to pass installation flags
+        return graph.register_library(*args, installation_arguments=cmd_args if args != cmd_args else [])
```
JCZuurmond, line 221: If a pip command has additional arguments defined besides the to be installed libraries, then these arguments are passed to the `register_library` so that it can be reused in the pip install

tests/integration/source_code/test_libraries.py
```diff
@@ -1,15 +1,39 @@
...
+def test_build_notebook_dependency_graphs_installs_wheel_with_pip_cell_in_notebook(simple_ctx, notebook):
+    ctx = simple_ctx.replace(path_lookup=MockPathLookup())
+
+    maybe = ctx.dependency_resolver.build_notebook_dependency_graph(Path(notebook))
+
+    assert not maybe.problems
+    assert maybe.graph.all_relative_names() == {f"{notebook}.py", "thingy/__init__.py"}
+
+
+@pytest.mark.xfail(reason="Spaces in path are not handled by subprocess when quoted")
```
JCZuurmond, line 31: A pip install with spaces in a path does not work in subprocess despite quoting them. It works when running the same install in a terminal

tests/unit/source_code/notebooks/test_cells.py
```diff
@@ -105,4 +107,33 @@ def test_pip_cell_build_dependency_graph_handles_multiline_code():
...
+            ["%pip", "install", "./distribution/dist/thingy-0.0.1-py2.py3-none-any.whl"],
+        ),
+        (
+            "%pip install distribution/dist/thingy-0.0.1-py2.py3-none-any.whl",
+            ["%pip", "install", "distribution/dist/thingy-0.0.1-py2.py3-none-any.whl"],
+        ),
+    ],
+)
+def test_pip_cell_split(code, split):
+    assert PipCell._split(code) == split  # pylint: disable=protected-access
```
JCZuurmond, line 139: @nfx : We discussed this. Here the unit test for a protected method was useful to unity test many different pip install versions. It breaks the "no cheat" CI

JCZuurmond (commented):
Please have a look at the PR and my comments. This PR introduces some API changes about; I would like to have feedback if the motivation is sufficient to motivate these changes

src/databricks/labs/ucx/source_code/notebooks/cells.py
nfx: this means we need to bundle `pip` with the other wheels when installing in no-internet workspaces, which adds a layer of complexity. depending on `_internal` package is a slippery slope. If you want to pass down more arguments to `pip`, why not just `def register_library(self, path_lookup: PathLookup, *arguments: str)`?.. that way, both `x.register_library(lookup, "tensorflow")` and `x.register_library(lookup, "/Workspace/x/y/x.whl", "--index-url", "https://...")` work. and we don't have to reimplement anything.

src/databricks/labs/ucx/source_code/python_libraries.py
```diff
@@ -46,41 +58,81 @@ def _temporary_virtual_environment(self):
...
+        problems = []
+        venv = self._temporary_virtual_environment
+        install_commands = []
+        if len(installation_arguments) == 0:
+            for library in libraries:
+                install_command = f"pip install {shlex.quote(library)} -t {venv}"
+                install_commands.append(install_command)
+        else:
+            # pip allows multiple target directories in its call, it uses the last one, thus the one added here
+            install_command = f"pip install {shlex.join(installation_arguments)} -t {venv}"
```
nfx, line 98: `install_command = f"pip install {shlex.quote(library)} -t {venv}"` and `install_command = f"pip install {shlex.join(installation_arguments)} -t {venv}"` are the same

nfx (changes requested):

src/databricks/labs/ucx/source_code/python_libraries.py
JCZuurmond: They aren't always the same: i) installation arguments are empty for the pip libraries defined on a job and ii) the libraries don't contain the installation flags when those are defined in a `%pip` command

src/databricks/labs/ucx/source_code/notebooks/cells.py
JCZuurmond: > this means we need to bundle pip with the other wheels when installing in no-internet workspaces, which adds a layer of complexity.

Isn't that also the case when using `pip` in a subprocess?

> depending on _internal package is a slippery slope

Agree!

The alternative you provided does not work well with the `Whitelist` as that requires to separate the (to-be-installed) libraries and the installation arguments. For simplification, we could only use the `Whitelist` for `len(arguments) == 0`, then the library will be installed at the cost of some overhead, but it will not be linted when an `import <library>` is defined

src/databricks/labs/ucx/source_code/graph.py
asnare: I think it's possible to make these changes while retaining compatibility:
```python
def register_library(self, path_lookup: PathLookup, library: Path, *more_libraries: Path, installation_arguments: list[str] | None = None) -> list[DependencyProblem]:
    ...
```
This type signature ensures that at least one argument is provided.

Can you elaborate a bit more on why there's a problem with the leading `./` being stripped? I'm a little confused because [anything without a leading `/` is considered relative](https://docs.python.org/3.10/library/pathlib.html#pathlib.PurePath.is_absolute), and if it's a problem why wasn't it a problem before?

JCZuurmond: With current changes, I am replacing the resolved path in the installation arguments which uses an equality to [replace the argument](https://github.com/databrickslabs/ucx/blob/9be7b05a4b5d9016b824ceead254d5a9752b99e8/src/databricks/labs/ucx/source_code/python_libraries.py#L84)

```
args = [maybe_library.as_posix() if arg == library else arg for arg in args]
```

Before the library path was "just" resolved, the path lookup can handle both the relative paths with and without the ./

JCZuurmond: I could split the `*libraries` into `library` and `libraries` to enforce at least defining one library. Though, it adds complexity to the code; the current code works; and allows users to unpack a possibly empty list of libraries. Though, it nudges users to define at least one library, which is can be a good thing. IMO only this change does not add much value

nfx: this gets too complicated - just pass `list[str]` down to `pip` and let it handle all the work. also `shlex` them and download any `pathlib.Path`-like files

src/databricks/labs/ucx/source_code/notebooks/cells.py
nfx: > Isn't that also the case when using pip in a subprocess?

external Public API of `pip` subprocess is way more stable than any internals. e.g. what if in the next minor version they decide to move packages around?

nfx (changes requested):

tests/integration/source_code/test_libraries.py
JCZuurmond: https://github.com/databrickslabs/ucx/issues/1859

src/databricks/labs/ucx/source_code/notebooks/cells.py
JCZuurmond: I removed the dependency on `pip._internal`

nfx (approved):
Lgtm

PR #1855, opened 2024-06-06
Consistent 0-based line tracking for linters

asnare:
## Changes

This PR includes the following changes with respect to how line information is tracked and reported by the linting code:

 - [X] We now use 0-based positions everywhere, instead of a mix of 0-based and 1-based. This particular choice is because that's also the [convention used by the LSP protocol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position).
 - [X] An off-by-one error has been fixed in the way the notebook parser tracks cell position offsets.
 - [X] Existing tests have been checked and updated to verify the 0-based line positions.
 - [X] A new test for cell parsing has been introduced.

For notebooks, there is an issue of what the position information means: for now the line position will refer to the 0-based line number in the underlying notebook file. (There is a related issue with respect to tracking column positions for `MAGIC` cells, but that is unaddressed by this PR.)

### Linked issues

Blocks #1843.


nfx (changes requested):
everything has to be zero-based

src/databricks/labs/ucx/source_code/base.py
```diff
@@ -21,6 +21,7 @@
 class Advice:
     code: str
     message: str
+    # Lines are 1-based, columns are 0-based.
```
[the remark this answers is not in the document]
asnare: Got it, I'll switch everything over to 0-based.

nfx (approved):
Lgtm

src/databricks/labs/ucx/source_code/notebooks/cells.py
```diff
@@ -24,13 +24,20 @@
...
+            raise ValueError("Cells must have a positive offset within the original file.")
         self._original_offset = original_offset
         self._original_code = source
         self._migrated_code = source
 
     @property
     def original_offset(self) -> int:
+        """Line offset of this cell within the original notebook file.
+
+        Example: if the line offset is 5, then line 10 within this cell corresponds to line 5 of the original file.
```
JCZuurmond, line 39: Shouldn't this be the other way around? 

> then line 10 within this cell corresponds to line 5 of the original file.

should be

> then line 5 within this cell corresponds to line 10 of the original file.

JCZuurmond (commented):
one comment

src/databricks/labs/ucx/source_code/notebooks/cells.py
asnare: See: #1874

PR #1863, opened 2024-06-07
Support inferred values when linting sys path

ericvergnaud:
## Changes
Infer value of sys.path.xxx() calls and raise advice if unresolved

### Linked issues
Progresses #1205

### Functionality 

- [ ] added relevant user documentation
- [ ] added new CLI command
- [ ] modified existing command: `databricks labs ucx ...`
- [ ] added a new workflow
- [ ] modified existing workflow: `...`
- [ ] added a new table
- [ ] modified existing table: `...`

### Tests
- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests
- [ ] verified on staging environment (screenshot attached)

Issue #1205, opened 2024-04-01
[FEATURE]: If a code computes a value dynamically, do value inference, at least at the state of linting


nfx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

If arguments to the relevant functions are string literals, resolve them and perform linting:

- https://github.com/databrickslabs/ucx/issues/1133
- https://github.com/databrickslabs/ucx/issues/1118
- https://github.com/databrickslabs/ucx/issues/1104
- https://github.com/databrickslabs/ucx/issues/1082
- https://github.com/databrickslabs/ucx/issues/391
- https://github.com/databrickslabs/ucx/issues/315

e.g. in cases like:

```python
db = 'foo'
t = f'{db}.bar'
display(spark.table(t))
```

we have to check if `foo.bar` table is migrated. 

### Proposed Solution

infer values from the available scope.

ericvergnaud:
migration to astroid is successful
simple inference works

What doesn't work at all (no inference):
 - fstrings i.e.` dbutils.notebook.run(f"prefix {value} suffix")` -> see test_raises_advice_when_dbutils_notebook_run_is_too_complex

What is problematic:
 - loops:
```
paths = [ "p1", "p2" ] 
for path in paths:
    dbutils.notebook.run(path)
```
the value is inferred only once (for the first item)

ericvergnaud:
@nfx I believe value inference is done for all currently implemented scenarios, see #1835, #1860, #1866, #1868 and #1870
Not sure if missing linters should be implemented as part of this issue (I guess not) ? Rather value inference should be done as part of their dedicated implementation ?

ericvergnaud:
@nfx as of writing, the only known issue is with fstrings (see above). I've created corresponding ticket #1871

nfx:
@ericvergnaud lets solve the f-string issue, as it's very common

PR #1866, opened 2024-06-09
Support inferred values when linting sys path

ericvergnaud:
## Changes
Infer value of sys.path.xxx() calls and raise advice if unresolved

### Linked issues

Progresses #1205
Replaces PR #1863

### DIFF
--- a/src/databricks/labs/ucx/source_code/linters/imports.py
+++ b/src/databricks/labs/ucx/source_code/linters/imports.py
@@ -233,6 +233,7 @@ class SysPathChangesVisitor(TreeVisitor):
     def _visit_inferred(self, changed: NodeNG, inferred: NodeNG, is_relative: bool, is_append: bool):
         if inferred is Uninferable or not isinstance(inferred, Const):
             self.sys_path_changes.append(UnresolvedPath(changed, changed.as_string(), is_append))
+            return
         if is_relative:
             self.sys_path_changes.append(RelativePath(changed, inferred.value, is_append))
         else:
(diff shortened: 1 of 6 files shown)


ericvergnaud:
@nfx only change from #1863 is test coverage

tests/unit/source_code/test_dependencies.py
```diff
@@ -20,55 +22,47 @@
 from tests.unit.conftest import MockPathLookup
 
 
+def dependency_resolver(notebook_resolver, path_lookup):
```
JCZuurmond, line 25: Make this a fixture

src/databricks/labs/ucx/source_code/linters/imports.py
```diff
@@ -211,10 +217,27 @@ def visit_call(self, node: Call):
...
+        relative = False
+        if isinstance(changed, Call):
+            if not self._match_aliases(changed.func, ["os", "path", "abspath"]):
+                return
+            relative = True
+            changed = changed.args[0]
+        try:
+            all_inferred = changed.inferred()
+            for inferred in all_inferred:
+                self._visit_inferred(changed, inferred, relative, is_append)
```
JCZuurmond, line 229: This will not raise an `InferenceError`, right? I prefer to move it out of the try

JCZuurmond (changes requested):
Two minor remarks

src/databricks/labs/ucx/source_code/linters/imports.py
ericvergnaud: That would require `all_inferred` to be available outside the try-except clause which would increase complexity

tests/unit/source_code/test_dependencies.py
ericvergnaud: done

PR #1868, opened 2024-06-10
Support inferred values when linting DBFS mounts

ericvergnaud:
## Changes
Implement value inference
Improve consistency of generated advices

### Linked issues
Progresses #1205

### DIFF
--- a/src/databricks/labs/ucx/source_code/linters/dbfs.py
+++ b/src/databricks/labs/ucx/source_code/linters/dbfs.py
@@ -1,11 +1,15 @@
+import logging
 from collections.abc import Iterable
 
-from astroid import Call, Const  # type: ignore
+from astroid import Call, Const, InferenceError, NodeNG  # type: ignore
 import sqlglot
 from sqlglot.expressions import Table
 
-from databricks.labs.ucx.source_code.base import Advice, Linter, Advisory, Deprecation
+from databricks.labs.ucx.source_code.base import Advice, Linter, Deprecation
 from databricks.labs.ucx.source_code.linters.python_ast import Tree, TreeVisitor
+
+
+logger = logging.getLogger(__file__)
 
 
 class DetectDbfsVisitor(TreeVisitor):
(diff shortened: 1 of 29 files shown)


src/databricks/labs/ucx/source_code/linters/dbfs.py
```diff
@@ -21,30 +25,36 @@ def __init__(self):
...
-                    self._reported_locations.add((arg.lineno, arg.col_offset))
+            self._visit_arg(arg)
+
+    def _visit_arg(self, arg: NodeNG):
+        try:
+            for inferred in arg.inferred():
+                if isinstance(inferred, Const) and isinstance(inferred.value, str):
+                    self._check_str_constant(arg, inferred)
+        except InferenceError:
+            logger.debug(f"Could not infer value of {str(arg)}")
```
asnare, line 36: In format strings I don't think you need to wrap arguments in `str()` because that will happen anyway.

Also, might it be worth adding in `exc_info=True` here?

src/databricks/labs/ucx/source_code/linters/dbfs.py
```diff
@@ -21,30 +25,36 @@ def __init__(self):
...
+        if any(value.startswith(prefix) for prefix in self._fs_prefixes):
+            advisory = Deprecation.from_node(
+                code='dbfs-usage', message=f"Deprecated file system path: {value}", node=source_node
+            )
+            self._advices.append(advisory)
+
+    def _already_reported(self, *nodes: NodeNG):
+        reported = any((node.lineno, node.col_offset) in self._reported_locations for node in nodes)
+        for node in nodes:
+            self._reported_locations.add((node.lineno, node.col_offset))
```
asnare, line 56: As an aside, I'm interested in other opinions on this: it looks like we're only ever going to report a single problem for a specific start location, when there might be multiple (independent) problems. Is that we want?

The linting model is to report as many problems as possible during a given run, and this seems to go against that. If we're suppressing issues like this, then I expect users will end up in a pattern of "I fixed this, and now there's a new problem in the same location?"

asnare:
I think this PR might also fix an off-by-one problem with the ending column for (at least some) linters.

src/databricks/labs/ucx/source_code/linters/dbfs.py
ericvergnaud: done

ericvergnaud:
> I think this PR might also fix an off-by-one problem with the ending column for (at least some) linters.
Yes I believe there was an issue where the code was missing the enclosing quotes of constant strings

src/databricks/labs/ucx/source_code/linters/dbfs.py
asnare: Okay, I can understand that. How do we ensure the "right" problem is reported when this sort of thing happens?

ericvergnaud: There's only 1 problem type, so the risk of reporting the wrong problem is inexistant

nfx (approved):
lgtm

PR #1870, opened 2024-06-10
Support inferred values when linting spark.sql

ericvergnaud:
## Changes
Infer value of tables spark.sql() calls and raise advice if unresolved

### Linked issues
Progresses #1205


nfx (approved):
Lgtm

PR #1872, opened 2024-06-10
Propagate widget parameters and data security mode to `CurrentSessionState`

nfx:
This PR adds propagation of named parameters from jobs to current session state. Additionally, we also check for DBFS paths as named and positional parameters.


nfx:
Test

Issue #1869, opened 2024-06-10
[FEATURE]: Report `DENY` ACLs during assessment


asnare:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

As noted during #1803, legacy ACLs may contain `DENY` rules interspersed amongst `GRANT` permissions. These are [not supported by UC](https://learn.microsoft.com/en-us/azure/databricks/sql/language-manual/security-deny) and cannot be migrated. As of #1815 these are handled during group migration, but during table migration the ACLs will be dropped and a warning generated.

This may be problematic for customers because it means additional users will end up having permissions on migrated tables, and it's not a 'fail-safe' situation.

Users should be warned earlier than table migration about this being a problem because fixing this might be quite time-consuming.

### Proposed Solution

During assessment we should detect and report on the presence of `DENY` ACLs.

PR #1890, opened 2024-06-12
Upgrade the `grants` schema to include assessment columns.

asnare:
## Changes

This PR upgrades the `grants` inventory table to include columns for tracking assessments.

### Linked issues

Relates: #1869 

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table
- [X] modified existing table: Additional `success` and `failures` columns for the `grants` table.

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/grants.py
+++ b/src/databricks/labs/ucx/hive_metastore/grants.py
@@ -59,6 +59,8 @@ class Grant:
     udf: str | None = None
     any_file: bool = False
     anonymous_function: bool = False
+    success: int = 1
+    failures: str = "[]"
 
     @staticmethod
     def type_and_key(
--- a/src/databricks/labs/ucx/upgrades/v0.27.0_add_assessment_to_grants.py
+++ b/src/databricks/labs/ucx/upgrades/v0.27.0_add_assessment_to_grants.py
@@ -1,0 +1,18 @@
+# pylint: disable=invalid-name
+
+import logging
+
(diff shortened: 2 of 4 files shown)


src/databricks/labs/ucx/hive_metastore/grants.py
```diff
@@ -59,6 +59,8 @@ class Grant:
     udf: str | None = None
     any_file: bool = False
     anonymous_function: bool = False
+    success: int = 1
+    failures: str = "[]"
```
nfx, line 63: why do we need to modify this schema? How can we do this without it?

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/grants.py
asnare: At the moment there is no location to record assessment findings for grants/permissions, and for most entities this is how they are recorded.

Two options for doing it without modifying the schema:

 - Use a table alongside, referring to the grant via a business key. (This doesn't make much sense because all the columns are part of the business key, so it effectively duplicates the table.)
 - Perform the findings logic entirely via views, in the style of `code_patterns` and `misc_patterns`.

Of these, the views approach would preferable even though it means relying on integration tests.

asnare:
Superseded by #1903.

Issue #1901, opened 2024-06-13
[BUG]: solacc is noisy and reports false positives


ericvergnaud:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

when running `make solacc`, a number of failures occur, and many advices are false positives

### Expected Behavior

All errors should be caught and all advices should be relevant

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

ericvergnaud:
I suggest closing this once all PRs that reference this issue are closed. The remaining noise should be handed on a per case basis.

PR #1891, opened 2024-06-12
Infer linted values that resolve to dbutils.widgets.get

ericvergnaud:
## Changes
When linting `dbutils.widget.get(...)` or `dbutils.widget.getAll()`, use named parameters from current session state to resolve the value

### Linked issues
Resolves #1880

PR #1905, opened 2024-06-13
Improved support for magic commands in python cells

ericvergnaud:
## Changes
support parsing python cells that contain magic lines
support execution of %pip/!pip magic command

### Linked issues
Progresses #1901 

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests
- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests


ericvergnaud:
@nfx 17:42:57  INFO [verify-accelerators] Parseable: 95% (1100/1155), missing imports: 1554
(with PR #1891 we should be getting very close to 1155)

src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -197,16 +208,31 @@ def build_graph_from_python_source(self, python_code: str) -> list[DependencyPro
...
         return problems
 
+    @classmethod
+    def convert_magic_commands(cls, python_code: str):
+        lines = python_code.split("\n")
+        for i, line in enumerate(lines):
+            if not line.startswith("%"):
+                continue
+            content = base64.b64encode(line.encode())  # use base64 to simplify escaping
+            lines[i] = f"magic_command('{content.decode('ascii')}')"
```
nfx, line 218: yep, this is roughly how jupyter implements it. `%pip x` is translated into `get_ipython().run_cell_magic('pip', 'x')`

look at `IPython.core.tests.test_inputsplitter.CellMagicsCommon` and surrounding classes

src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -480,3 +506,49 @@ class MaybeGraph:
     @property
     def failed(self):
         return len(self.problems) > 0
+
+
+T = TypeVar("T")
+
+# non-top-level import is required to avoid cyclic dependency
+# pylint: disable=wrong-import-position, cyclic-import
```
nfx, line 514: not allowed to disable this. move `MagicCommand` to commands.py

src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -480,3 +506,49 @@ class MaybeGraph:
...
+    @classmethod
+    def _make_commands_for_magic_command_call_nodes(cls, nodes: list[Call]):
+        for node in nodes:
+            arg = node.args[0]
+            if isinstance(arg, Const):
+                yield MagicCommand(node, arg.value)
+
+    def __init__(self, node: NodeNG, command: str):
+        super().__init__(node)
+        data = base64.b64decode(command)
```
nfx, line 545: do we _really need_ to base64 encode this, btw?..

src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -480,3 +506,49 @@ class MaybeGraph:
...
+        self._command = data.decode("ascii")
+
+    def execute(self, return_type: T, **kwargs) -> T:
```
nfx, line 548: if you move it to commands.py, then you don't need to add generic type here.

src/databricks/labs/ucx/source_code/notebooks/commands.py
```diff
@@ -0,0 +1,48 @@
+import re
```
nfx, line 1: why it's not in the cells.py? `%pip` is one of the line magic types.

src/databricks/labs/ucx/source_code/notebooks/commands.py
```diff
@@ -0,0 +1,48 @@
...
+import shlex
+from typing import TypeVar, cast
+
+from databricks.labs.ucx.source_code.graph import DependencyGraph, DependencyProblem
+
+T = TypeVar("T")
+
+
+class PipCommand:
```
nfx, line 10:
```suggestion
class PipMagic:
```
command is something different in the Jupyter protocol.

nfx (changes requested):

src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -197,16 +208,31 @@ def build_graph_from_python_source(self, python_code: str) -> list[DependencyPro
                 problems.append(problem)
         return problems
 
+    @classmethod
+    def convert_magic_commands(cls, python_code: str):
```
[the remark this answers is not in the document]
ericvergnaud: done

src/databricks/labs/ucx/source_code/graph.py
ericvergnaud: The command could contain _any_ character so we need to either escape the problematic ones (and unescape them later somewhat blindly), or encode the entire command to a charset known to not require escaping .
I believe base64 to be a safe and low cost option but I'm open to suggestions.

src/databricks/labs/ucx/source_code/graph.py
ericvergnaud: Done but unfortunately that doesn't remove the need for a late import, open to solutions...

src/databricks/labs/ucx/source_code/graph.py
ericvergnaud: noted

src/databricks/labs/ucx/source_code/graph.py
ericvergnaud: Dropping for now but I suspect it will come back when we execute from a linter, which will require a different return type.

src/databricks/labs/ucx/source_code/notebooks/commands.py
ericvergnaud: because there are dozens of magic lines (run `%lsmagic` in. notebook), not sure we want to pollute cells.py with all that stuff ?

src/databricks/labs/ucx/source_code/notebooks/commands.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/graph.py
ericvergnaud: I tried moving `DependencyGraph.build_graph_from_python_source` to a dedicated file, didn't help...

src/databricks/labs/ucx/source_code/notebooks/magic.py
```diff
@@ -0,0 +1,93 @@
...
+
+class MagicCommand(NodeBase):
+
+    @classmethod
+    def convert_magic_lines_to_magic_commands(cls, python_code: str):
+        lines = python_code.split("\n")
+        for i, line in enumerate(lines):
+            if not line.startswith("%"):
+                continue
+            lines[i] = f"magic_command({line.encode()!r})"
```
JCZuurmond, line 21: This `magic_command` "method" is added to use `python_ast.tree` I understand?

src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -482,3 +489,8 @@ class MaybeGraph:
     @property
     def failed(self):
         return len(self.problems) > 0
+
+
+# non-top-level import is required to avoid cyclic dependency
+# pylint: disable=wrong-import-position, cyclic-import
```
nfx, line 495: Why can't magic command be part of this file? Cyclic import disable is not allowed, as it leads to a bigger mess 🚫

src/databricks/labs/ucx/source_code/linters/imports.py
```diff
@@ -116,6 +116,11 @@ def __init__(self, session_state: CurrentSessionState):
         self._session_state = session_state
 
     def lint(self, code: str) -> Iterable[Advice]:
+        # can't find a way to avoid this cyclic import
+        # pylint: disable=import-outside-toplevel, cyclic-import
+        from databricks.labs.ucx.source_code.notebooks.magic import MagicCommand
```
nfx, line 121: Disabling cyclic import checker is not allowed

nfx (changes requested):
Disabling cyclic import checker is not allowed

src/databricks/labs/ucx/source_code/notebooks/commands.py
nfx: so `%lsmagic foo bar` gets translated into `get_ipython().run_line_magic('lsmagic', '', 'foo', 'bar')`. `%md` is one of the magics as well. `%scala`, `%r`, `%sql`, `%python` - they're all magics. Databricks REPL doesn't make a difference between cell magics and line magics, so keep it all in cells.py.

src/databricks/labs/ucx/source_code/graph.py
nfx: see comment about how magics get translated under the hood

src/databricks/labs/ucx/source_code/notebooks/magic.py
ericvergnaud: not exactly, we're using astroid, and we need magic commands to be converted to regular python code

src/databricks/labs/ucx/source_code/graph.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/notebooks/commands.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/linters/imports.py
ericvergnaud: done

ericvergnaud:
@nfx not sure why `no-cheat` fails...

ericvergnaud:
> @nfx not sure why `no-cheat` fails...

actually a `# pylint: disable` was moved from a previous version

asnare:
Some notes on the way magic works:

 - If a line starts with `!` then it's just run as a system command. Arguably `!foo` should translated to `os.system("foo");` during pre-processing.
 - If a line starts with `!!` then it's translated into `%sx` magic: `!!foo` -> `%sx foo`<br />
   Arguably this should also be handled during pre-processing.
 - `%sx` is magic that uses (deprecated) `commands.getoutput(…)` to run a shell command and capture the output as a list of lines.
 - Databricks seems to only recognise cell magic (`%%foo`) if it's the first line in a cell. Looking at the internal dataset only `%%time` has been used, which can be ignored for us.
 - Line-magic can also be used as an expression, in the form of:
   ```python
   all_files = %sx find -type f`
   ```
   This isn't something I've found being used yet.
 - Automagic is something that is being used in a handful of places. This means that this:
   ```
   pip freeze
   ```
   Is translated into this:
   ```
   %pip freeze
   ```
   … but ONLY if there is no `pip` variable in scope.
 - There's a semi-related feature _autocall_ feature, which allows for Scala-style omission of brackets when making function calls. This:
   ```
   list "what?"
   ```
   is translated into this:
   ```python
   list("what?")
   ```
   This isn't enabled by default, and I haven't seen it being used anywhere.

asnare:
Of the line-magics that are supported, by far the most common are:

 - `%config`: Irrelevant for us.
 - `%matplotlib`: Also irrelevant for us.
 - `%pip`: Already handled.
 - `%reload_ext`: Normally only used with the `autoreload` extension, which isn't relevant for us.
 - `%run`: This is a little funky. Databricks treats it like cell-magic if it's the first line, and line-magic otherwise. We partially handle this already (by chasing dependencies) but don't handle it in terms of value inference. (A common pattern on the way to proper library development is to define a bunch of common settings and functions in a notebook and then %run them to import them.)

 -

asnare:
> * `%run`: This is a little funky. Databricks treats it like cell-magic if it's the first line, and line-magic otherwise. We partially handle this already (by chasing dependencies) but don't handle it in terms of value inference. (A common pattern on the way to proper library development is to define a bunch of common settings and functions in a notebook and then `%run` them to import them.)

I've created #1927 for this.

src/databricks/labs/ucx/source_code/linters/python_ast.py
```diff
@@ -27,6 +27,16 @@ def parse(code: str):
         root = parse(code)
         return Tree(root)
 
+    @classmethod
+    def convert_magic_lines_to_magic_commands(cls, python_code: str):
```
nfx, line 31: technically, it's better to make it part of the `parse` method

nfx (approved):
lgtm

PR #1903, opened 2024-06-13
Include findings on `DENY` grants during assessment

asnare:
## Changes

This PR adds support for flagging DENY permissions on objects, which cannot be migrated to UC.

It also adds some new integration tests for existing grant-scanning.

### Linked issues

Resolves #1869, supersedes #1890.

### Functionality 

- [x] added relevant user documentation
- [ ] added new CLI command
- [ ] modified existing command: `databricks labs ucx ...`
- [ ] added a new workflow
- [X] modified existing workflow: `assessment`
- [ ] added a new table
- [ ] modified existing table: `...`
- [x] modified existing view: `grant_detail`

### Tests

- [x] manually tested
- [ ] added unit tests
- [X] added integration tests
- [x] verified on staging environment (screenshot attached)

PR #1911, opened 2024-06-14
Fix grant visibility and classification

asnare:
## Changes

This PR follows on from #1903 and fixes some problems in the underlying inventory for grants and their details:

 - The `grant_details` view does not cover grants at the catalog level.
 - The object type for UDF and view grants is misclassified.

(These problems in the underlying inventory were present prior to #1903 but latent because they weren't directly exposed by the dashboard.)

### Functionality 

- [ ] added new CLI command
- [ ] added a new workflow
- [X] modified existing workflow: `assessment`
- [ ] added a new table
- [X] modified existing view: `grant_detail`

### Tests

- [X] manually tested
- [X] added unit tests
- [X] added integration tests
- [X] verified on staging environment (screenshot attached)

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/grants.py
+++ b/src/databricks/labs/ucx/hive_metastore/grants.py
@@ -246,8 +246,12 @@ class GrantsCrawler(CrawlerBase[Grant]):
                 tasks.append(partial(self.grants, catalog=catalog, database=database))
         for table in self._tc.snapshot():
             fn = partial(self.grants, catalog=catalog, database=table.database)
-            # views are recognized as tables
-            tasks.append(partial(fn, table=table.name))
+            # Views are treated as a type of table and enumerated by the table crawler.
+            if table.view_text is None:
+                task = partial(fn, table=table.name)
+            else:
+                task = partial(fn, view=table.name)
+            tasks.append(task)
         for udf in self._udf.snapshot():
             fn = partial(self.grants, catalog=catalog, database=udf.database)
             tasks.append(partial(fn, udf=udf.name))
@@ -268,6 +272,17 @@ class GrantsCrawler(CrawlerBase[Grant]):
         for grant in self.grants(catalog=schema.catalog_name, database=schema.name):
             principal_permissions[grant.principal].add(grant.action_type)
         return principal_permissions
+
+    # The reported ObjectType for grants doesn't necessary match the type of grants we asked for. This maps
+    # those that aren't themselves.
+    _grants_reported_as = {
+        # SHOW type: RESULT type
+        "SCHEMA": "DATABASE",
+        "CATALOG": "CATALOG$",
+        "VIEW": "TABLE",
+        "ANY FILE": "ANY_FILE",
+        "ANONYMOUS FUNCTION": "ANONYMOUS_FUNCTION",
+    }
 
     def grants(
         self,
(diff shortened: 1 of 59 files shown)


asnare:
image

src/databricks/labs/ucx/hive_metastore/grants.py
```diff
@@ -269,6 +273,17 @@ def for_schema_info(self, schema: SchemaInfo):
             principal_permissions[grant.principal].add(grant.action_type)
         return principal_permissions
 
+    # The reported ObjectType for grants doesn't necessary match the type of grants we asked for. This maps
+    # those that aren't themselves.
+    _grants_reported_as = {
+        # SHOW type: RESULT type
+        "SCHEMA": "DATABASE",
+        "CATALOG": "CATALOG$",
```
JCZuurmond, line 281: Why the dollar sign?

src/databricks/labs/ucx/hive_metastore/grants.py
```diff
@@ -318,18 +333,13 @@ def grants(
...
-                "ANONYMOUS_FUNCTION": "ANONYMOUS FUNCTION",
-            }
+        expected_grant_object_type = self._grants_reported_as.get(on_type, on_type)
+        grants = []
+        try:
             for row in self._fetch(f"SHOW GRANTS ON {on_type} {escape_sql_identifier(key)}"):
                 (principal, action_type, object_type, _) = row
-                object_type = object_type_normalization.get(object_type, object_type)
-                if on_type != object_type:
+                # This seems to be here to help unit tests that don't sufficiently narrow the target of mocked queries. -ajs
```
JCZuurmond, line 341: Can we verify this comment and update it to be affirmative instead of questioning

JCZuurmond (changes requested):
Have some clarifying points

src/databricks/labs/ucx/hive_metastore/grants.py
asnare: This table maps what we ask for in the `SHOW` statement to the object type as reported in the result-set. In the case of Catalog-scoped permissions it goes like this:
```python
spark.sql("show grants on catalog hive_metastore").show(truncate=False)
```
```
+--------------+----------+----------+---------+
|Principal     |ActionType|ObjectType|ObjectKey|
+--------------+----------+----------+---------+
|ajs_test_group|SELECT    |CATALOG$  |None     |
+--------------+----------+----------+---------+
```

src/databricks/labs/ucx/hive_metastore/grants.py
asnare: So that's interesting. As written (prior to this PR, and preserved as such) this piece of code filters out grants for object types that weren't requested while dealing with the fact that sometimes grants are reported slightly differently to how they were requested.

The current code is from here, squashed from #246: https://github.com/databrickslabs/ucx/blob/cbc26542f2c7edfcfe0e9d193dfe01803acb1333/src/databricks/labs/ucx/hive_metastore/grants.py#L231

And there was some discussion about it [here](https://github.com/databrickslabs/ucx/pull/246#discussion_r1331907633) where @william-conti wrote:
> It's a corner case where the object type isn't the same as the one requested. For example you request the grant for a view and you get the grants for a table, it's not a normal behaviour and should be highlighted

However this only addresses expected cases where the requested/reported types don't match: what we don't know is when in practice this actually occurs (and why the guard is in place).

Since the original code was written, it turns out that the main place this check is used is with unit tests written in the meantime that don't mock the statements correctly. Essentially these mock `SHOW GRANTS ON .*` and therefore the mocks report table grants when we query the catalog grants and stuff like that. Here's an example of this:
https://github.com/databrickslabs/ucx/blob/a8e823f1bcff971f7bfa266bb4233216fb8dc4ac/tests/unit/hive_metastore/test_grants.py#L205

As things stand instead of this check I think I'd prefer to fix the unit tests (they're unintentionally mocking "broken" behaviour) and remove the check entirely.

JCZuurmond: Agreed, but out-of-scope for this PR though. Could you create a tech debt issue for this?

JCZuurmond (approved):

src/databricks/labs/ucx/hive_metastore/grants.py
asnare: > Could you create a tech debt issue for this?

Done; #1932.

PR #1912, opened 2024-06-14
Improve solacc script

ericvergnaud:
## Changes
Improve solacc script as follows:
 - skip malformed files
 - record list of unparseable files
 - allow linting just 1 file

### Linked issues
Progresses #1901 

### Functionality 
- [ ] modified existing command: `make solacc`

### Tests
- [x] manually tested

### DIFF
--- a/tests/integration/source_code/solacc.py
+++ b/tests/integration/source_code/solacc.py
@@ -85,7 +85,9 @@ def lint_all(file_to_lint: str | None):
                     f.write(file.relative_to(dist).as_posix())
                     f.write("\n")
     parseable_pct = int(parseable / len(all_files) * 100)
-    logger.info(f"Skipped: {len(skipped or [])}, parseable: {parseable_pct}% ({parseable}/{len(all_files)}), missing imports: {missing_imports}")
+    logger.info(
+        f"Skipped: {len(skipped or [])}, parseable: {parseable_pct}% ({parseable}/{len(all_files)}), missing imports: {missing_imports}"
+    )
     if parseable_pct < 100:
         sys.exit(1)
 
@@ -104,4 +106,3 @@ def main(args: list[str]):
 
 if __name__ == "__main__":
     main(sys.argv)
-
--- a/tests/integration/source_code/solacc.py
+++ b/tests/integration/source_code/solacc.py
@@ -45,8 +45,29 @@ def clone_all():
         run_command(f'git clone {url} {dst}')
 
 
+def lint_one(file: Path, ctx: LocalCheckoutContext, unparsed: Path | None) -> tuple[int, int]:
+    try:
+        missing_imports = 0
+        for located_advice in ctx.local_code_linter.lint_path(file):
+            if located_advice.advice.code == 'import-not-found':
+                missing_imports += 1
+            message = located_advice.message_relative_to(dist.parent, default=file)
+            sys.stdout.write(f"{message}\n")
+        return missing_imports, 1
+    except Exception as e:  # pylint: disable=broad-except
+        # here we're most likely catching astroid & sqlglot errors
+        if unparsed is None:  # linting single file, log exception detqils
+            logger.error(f"Error during parsing of {file}: {e}".replace("\n", " "), exc_info=e)
+        else:
+            logger.error(f"Error during parsing of {file}: {e}".replace("\n", " "))
+            # populate solacc-unparsed.txt
+            with unparsed.open(mode="a", encoding="utf-8") as f:
+                f.write(file.relative_to(dist).as_posix())
+                f.write("\n")
+        return 0, 0
+
+
 def lint_all(file_to_lint: str | None):
-    # pylint: disable=too-many-nested-blocks
     ws = WorkspaceClient(host='...', token='...')
     ctx = LocalCheckoutContext(ws).replace(
         linter_context_factory=lambda session_state: LinterContext(MigrationIndex([]), session_state)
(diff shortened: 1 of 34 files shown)


tests/integration/source_code/solacc.py
```diff
@@ -44,16 +45,27 @@ def clone_all():
...
     parseable = 0
     missing_imports = 0
-    all_files = list(dist.glob('**/*.py'))
+    all_files = list(dist.glob('**/*.py')) if file_to_lint is None else [Path(dist, file_to_lint)]
+    if file_to_lint is None:
+        unparsed = Path(Path(__file__).parent, "solacc-unparsed.txt")
+        if unparsed.exists():
+            os.remove(unparsed)
+    skipped: set[str] | None = None
+    malformed = Path(Path(__file__).parent, "solacc-malformed.txt")
```
asnare, line 62:
```suggestion
    malformed = Path(__file__).parent / "solacc-malformed.txt"
```

tests/integration/source_code/solacc.py
```diff
@@ -63,17 +75,34 @@ def lint_all():
             parseable += 1
         except Exception as e:  # pylint: disable=broad-except
             # here we're most likely catching astroid & sqlglot errors
-            logger.error(f"Error during parsing of {file}: {e}".replace("\n", " "), exc_info=e)
+            if file_to_lint:
+                logger.error(f"Error during parsing of {file}: {e}".replace("\n", " "), exc_info=e)
+            else:
+                logger.error(f"Error during parsing of {file}: {e}".replace("\n", " "))
```
asnare, line 81: Why are these branches needed? 🤔

I think here we can also use `logger.exception()` which will automatically include the exception stack trace, etc?

tests/integration/source_code/solacc.py
```diff
@@ -44,16 +45,27 @@ def clone_all():
...
         linter_context_factory=lambda session_state: LinterContext(MigrationIndex([]), session_state)
     )
     parseable = 0
     missing_imports = 0
-    all_files = list(dist.glob('**/*.py'))
+    all_files = list(dist.glob('**/*.py')) if file_to_lint is None else [Path(dist, file_to_lint)]
+    if file_to_lint is None:
+        unparsed = Path(Path(__file__).parent, "solacc-unparsed.txt")
+        if unparsed.exists():
+            os.remove(unparsed)
```
asnare, line 60: Instead of removing the file and then adding a line at a time as things run, what do you think about:
1. Starting with an empty list;
2. Filling the list as things run;
3. Replacing the file once at the end (with `.open("wt"`).

tests/integration/source_code/solacc.py
```diff
@@ -44,16 +45,27 @@ def clone_all():
         run_command(f'git clone {url} {dst}')
 
 
-def lint_all():
+def lint_all(file_to_lint: str | None):
```
asnare, line 48: I'd be tempted to allow a list of files to be passed in here.

tests/integration/source_code/solacc.py
```diff
@@ -63,17 +75,34 @@ def lint_all():
...
     install_logger()
     logging.root.setLevel(logging.INFO)
-    logger.info("Cloning...")
-    clone_all()
+    file_to_lint = args[1] if len(args) > 1 else None
+    if not file_to_lint:
+        # don't clone if linting just one file, assumption is we're troubleshooting
+        logger.info("Cloning...")
+        clone_all()
     logger.info("Linting...")
```
asnare, line 103: Would like to see information about what's going to be linted here: there might not be any other feedback about what's happening until it's finished.

tests/integration/source_code/solacc.py
```diff
@@ -63,17 +75,34 @@ def lint_all():
...
-    logger.info(f"Parseable: {parseable_pct}% ({parseable}/{len(all_files)}), missing imports: {missing_imports}")
+    logger.info(
+        f"Skipped: {len(skipped or [])}, parseable: {parseable_pct}% ({parseable}/{len(all_files)}), missing imports: {missing_imports}"
+    )
     if parseable_pct < 100:
         sys.exit(1)
 
 
-if __name__ == "__main__":
+def main(args: list[str]):
```
asnare, line 95: 👍

tests/integration/source_code/solacc.py
ericvergnaud: we only want the stack trace when debugging a single file otherwise it pollutes the output

tests/integration/source_code/solacc.py
ericvergnaud: I think it's clearer to remove the file upfront

tests/integration/source_code/solacc.py
ericvergnaud: that doesn't correspond to a usage scenario i.e. either you're checking the entire data set or you're troubleshooting 1 file. But feel free to improve later if you fond it useful.

tests/integration/source_code/solacc.py
ericvergnaud: This is a tool for us, so not sure how useful such feedback would be, but feel free to enhance

ericvergnaud:
@nfx there's no `#pylint: disable` in the submitted code, so not sure why no-cheat fails ? (this seems to happen frequently...)

ericvergnaud:
> @nfx there's no `#pylint: disable` in the submitted code, so not sure why no-cheat fails ? (this seems to happen frequently...)

actually it was moved from the previous version

PR #1915, opened 2024-06-17
catch sqlglot exceptions and convert them to advices

ericvergnaud:
## Changes
Generate Failure when sqlglot fails to parse query

### Linked issues
Progresses #1901

### Functionality 
None

### Tests
- [x] added unit tests

### DIFF
--- a/src/databricks/labs/ucx/source_code/queries.py
+++ b/src/databricks/labs/ucx/source_code/queries.py
@@ -49,7 +49,8 @@ class FromTable(Linter, Fixer):
                 if not statement:
                     continue
                 yield from self._lint_statement(statement)
-        except SqlParseError:
+        except SqlParseError as e:
+            logger.debug(f"Failed to parse SQL: {code}", exc_info=e)
             yield Failure(
                 code='table-migrate',
                 message=f"SQL query is not supported yet: {code}",
--- a/src/databricks/labs/ucx/source_code/queries.py
+++ b/src/databricks/labs/ucx/source_code/queries.py
@@ -52,7 +52,7 @@ class FromTable(Linter, Fixer):
         except SqlParseError:
             yield Failure(
                 code='table-migrate',
-                message=f"SQL query {code} is not supported yet",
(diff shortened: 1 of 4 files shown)


src/databricks/labs/ucx/source_code/queries.py
```diff
@@ -43,41 +43,56 @@ def schema(self):
...
+        try:
+            statements = parse_sql(code, read='databricks')
+            for statement in statements:
+                if not statement:
                     continue
+                yield from self._lint_statement(statement)
+        except SqlParseError:
+            yield Failure(
+                code='table-migrate',
+                message=f"SQL query {code} is not supported yet",
```
asnare, line 55: The code being (not) parsed could be quite long: tens or even hundreds of lines. At the very least for this I'd prefer to see it at the end of the message rather than in the middle.
```suggestion
                message=f"SQL query is not supported yet: {code}",
```

src/databricks/labs/ucx/source_code/queries.py
```diff
@@ -43,41 +43,56 @@ def schema(self):
...
-                    # Sqlglot captures the schema name in the Create statement as a Table, with
-                    # the schema  as the db name.
-                    self._session_state.schema = table.db
+        try:
+            statements = parse_sql(code, read='databricks')
+            for statement in statements:
+                if not statement:
                     continue
+                yield from self._lint_statement(statement)
+        except SqlParseError:
```
asnare, line 52: At the moment for people running into this there's no way for them to figure out why something is failing to parse; at the very least we should log the cause of the SQL error. To keep the noise down I'd be okay with `DEBUG` (and `exc_info=True`); it's really about ensuring we can get the information if we need it without having to reproduce it manually ourselves.

asnare (approved):
Some minor suggestions but lgtm.

src/databricks/labs/ucx/source_code/queries.py
ericvergnaud: done

JCZuurmond (approved):

PR #1919, opened 2024-06-19
Improve catch-all handling and avoid some pylint suppressions

asnare:
## Changes

This PR is stacked on top of #1905; changes include:

 - Avoid some pylint-suppressions that aren't really necessary.
 - Improve logging when catch-all handlers are hit so we can better understand what went wrong.
 - When an internal error occurs during magic-processing, work already completed is returned instead of abandoned.

Incidental changes include some more type annotations amongst the unit tests.

### Tests

- manually tested
- added unit tests


tests/unit/source_code/notebooks/test_cells.py
```diff
@@ -206,7 +253,14 @@ def test_parses_python_cell_with_magic_commands(simple_dependency_resolver, mock
     ],
 )
 def test_pip_command_split(code, split):
-    assert PipMagic._split(code) == split  # pylint: disable=protected-access
+
+    # Avoid direct protected access to the _split method.
```
JCZuurmond, line 257: Oke, that is an interesting workaround. I am not settled on the protected access in test cases, I like them sometimes, when it is clear unit testable functionality, like here

JCZuurmond (approved):
See comments

src/databricks/labs/ucx/source_code/notebooks/cells.py
```diff
@@ -477,11 +479,11 @@ def extract_from_tree(
             nodes = tree.locate(Call, [("magic_command", Name)])
             for command in cls._make_commands_for_magic_command_call_nodes(nodes):
                 commands.append(command)
-            return commands, problems
         except Exception as e:  # pylint: disable=broad-except
```
JCZuurmond, line 482: Can we make this one more narrow?

tests/unit/source_code/notebooks/test_cells.py
asnare: When code has been well-designed with testing in mind and makes good use of functional composition, I agree it can often be avoided.

However in practice most codebases don't function that way and the fixtures for unit tests are (by necessity) tightly coupled to the implementation of whatever is being tested. This means better-than-public access is frequently required.

My preferred approach here would be to relax the linting rules so that unit tests are allowed to access protected attributes. The current situation is that:
 - There are reasons for doing it.
 - The linting rules don't allow it, and there is no effective override due to the CI checks.
 - Combined, these lead to code such as this that is worse than allowing it in the first place.

JCZuurmond: I agree with allowing the protected access in the tests. I ran into the same problem. Though, I think it good practice to nudge the test to test the public API's in the first place, but also have the option to narrowly test non public API's

nfx (approved):
lgtm

nfx (approved):
lgtm

PR #1778, opened 2024-05-27
EditorConfig specification for the project and minor tooling improvements

asnare:
## Changes

This PR includes some trivial repository updates:

 - An [EditorConfig](https://editorconfig.org) profile to set basic code formatting, set to match the files already in the repo. (IntelliJ/PyCharm should pick this up out of the box.)
 - A warning when running the `make verify` target is eliminated.
 - The `Makefile` targets are now marked as [phony](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html), which is more of a hygiene thing than anything else.

### Functionality 

- [ ] added relevant user documentation
- [ ] added new CLI command
- [ ] modified existing command: `databricks labs ucx ...`
- [ ] added a new workflow
- [ ] modified existing workflow: `...`
- [ ] added a new table
- [ ] modified existing table: `...`

### Tests

- [X] manually tested
- [ ] added unit tests
- [ ] added integration tests
- [ ] verified on staging environment (screenshot attached)

PR #1923, opened 2024-06-20
Tidy up whitespace

asnare:
## Changes

This PR follows on from #1778 by cleaning up files that use tabs, have trailing whitespace, or are missing EOL at the end of files. By fixing these in one go other PRs become less noisy. (Sample files used in tests have been left alone and are not affected by this PR.)


JCZuurmond (changes requested):
Is this the effect of changing a linter setting somewhere? The Changelog changes are a bit tricky, they will come. back

asnare:
> Is this the effect of changing a linter setting somewhere?

It's a side-effect of the editor settings in #1778. Historically we have quite a few files with trailing whitespace and missing the EOL at EOF. These aren't enforced by linting (which only covers python code), but with the new editor settings in place any time you modify a file trailing whitespace elsewhere in the file will be stripped. This was intended as a fix-them-all-for-now PR to avoid spurious updates on other PRs, particularly in Markdown.

> The Changelog changes are a bit tricky, they will come back

Agreed; I'm happy to drop those given that the file is generated.

JCZuurmond (approved):

LICENSE
```diff
@@ -1,69 +1,69 @@
                                Databricks License
                         Copyright (2023) Databricks, Inc.
 
-    Definitions. 
-    
-    Agreement: The agreement between Databricks, Inc., and you governing 
-    the use of the Databricks Services, as that term is defined in 
-    the Master Cloud Services Agreement (MCSA) located at 
+    Definitions.
```
nfx, line 4: please revert changes to this file

nfx (approved):
lgtm

PR #1925, opened 2024-06-20
Support run assessment as a collection

HariGS-DB:
## Changes
 - Support running eligible CLI commands as a collection
 - Added initially for assessment run cli command 
 - added collection_workspace_id parameter
 - If its None then use workspace client and run current installation workflow
 - if collection_workspace_id is passed create account context and iterate through all workspaces of the collection and run assessment workflow

### Linked issues

Resolves #..

### Functionality 

- [ ] added new CLI command
- [X] modified existing command: `databricks labs ucx ...`
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [X] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/cli.py
```diff
@@ -147,12 +148,22 @@ def validate_external_locations(w: WorkspaceClient, prompts: Prompts):
 
 
 @ucx.command
-def ensure_assessment_run(w: WorkspaceClient):
+def ensure_assessment_run(
+    w: WorkspaceClient, collection_workspace_id: int | None = None, a: AccountClient | None = None
```
nfx, line 152: let's make simpler knobs. we have `w.get_workspace_id()` already in the SDK.

src/databricks/labs/ucx/install.py
```diff
@@ -676,13 +675,40 @@ def _get_workspace(workspace_id: int, ids_to_workspace: dict[int, Workspace]) ->
             msg = "Current workspace is not known, Please run as account-admin: databricks labs ucx sync-workspace-info"
             raise KeyError(msg) from None
 
-    def _get_workspace_info(self, current_workspace_id: int):
+    def get_workspace_info(self, current_workspace_id: int):
```
nfx, line 678: keep it private

nfx (changes requested):

src/databricks/labs/ucx/cli.py
```diff
@@ -147,12 +148,23 @@ def validate_external_locations(w: WorkspaceClient, prompts: Prompts):
 
 
 @ucx.command
-def ensure_assessment_run(w: WorkspaceClient):
+def ensure_assessment_run(w: WorkspaceClient, run_as_collection: bool = False, a: AccountClient | None = None):
```
ericvergnaud, line 151: Can we use a more expressive name for AccountClient ? For example `account_client` ?

src/databricks/labs/ucx/install.py
```diff
@@ -676,17 +675,39 @@ def _get_workspace(workspace_id: int, ids_to_workspace: dict[int, Workspace]) ->
...
-        installation = Installation.current(current_workspace_client, self.product_info.product_name())
-        workspace_info = WorkspaceInfo(installation, current_workspace_client)
-        return workspace_info.load_workspace_info()
+        workspaces = account_client.workspaces.list()
+        ids_to_workspace = {}
+        for workspace in workspaces:
+            ids_to_workspace[workspace.workspace_id] = workspace
+        return ids_to_workspace
+
+    def get_workspaces_context(self, collection_workspace_id: int) -> list[WorkspaceContext]:
```
ericvergnaud, line 686: since this is returning a collection, maybe rename to `get_workspace_contexts` ?

src/databricks/labs/ucx/install.py
```diff
@@ -676,17 +675,39 @@ def _get_workspace(workspace_id: int, ids_to_workspace: dict[int, Workspace]) ->
             msg = "Current workspace is not known, Please run as account-admin: databricks labs ucx sync-workspace-info"
             raise KeyError(msg) from None
 
-    def _get_workspace_info(self, current_workspace_id: int):
+    def _get_workspace_info(self):
```
ericvergnaud, line 678: maybe rename to `_get_workspace_ids` ?

src/databricks/labs/ucx/cli.py
```diff
@@ -147,12 +148,23 @@ def validate_external_locations(w: WorkspaceClient, prompts: Prompts):
...
     """ensure the assessment job was run on a workspace"""
-    ctx = WorkspaceContext(w)
-    deployed_workflows = ctx.deployed_workflows
-    if not deployed_workflows.validate_step("assessment"):
-        deployed_workflows.run_workflow("assessment")
+    if run_as_collection:
+        if not a:
+            a = AccountClient(product='ucx', product_version=__version__)
+        account_installer = AccountInstaller(a)
+        workspaces_context = account_installer.get_workspaces_context(w.get_workspace_id())
```
ericvergnaud, line 157: rename to `workspace_contexts`

src/databricks/labs/ucx/cli.py
```diff
@@ -473,10 +485,11 @@ def revert_dbsql_dashboards(w: WorkspaceClient, dashboard_id: str | None = None)
 
 
 @ucx.command(is_account=True)
-def join_collection(a: AccountClient, workspace_id: int, join_workspace_id: int):
+def join_collection(a: AccountClient, workspace_ids: str, target_workspace_id: int | None = None):
```
ericvergnaud, line 488: rename 'a'

tests/unit/test_cli.py
```diff
@@ -218,6 +219,13 @@ def test_ensure_assessment_run(ws):
     ws.jobs.list_runs.assert_called_once()
 
 
+def test_ensure_assessment_run_collection(ws):
+    acc_client = create_autospec(AccountClient)
```
ericvergnaud, line 223: don't we have a fixture for AccountClient ?

ericvergnaud (changes requested):

src/databricks/labs/ucx/cli.py
nfx: it's the convention/blueprint workings.

src/databricks/labs/ucx/cli.py
nfx: not in `@XXX.command`

tests/unit/test_cli.py
HariGS-DB: updated to use the fixture

src/databricks/labs/ucx/install.py
HariGS-DB: done

src/databricks/labs/ucx/install.py
HariGS-DB: done

src/databricks/labs/ucx/install.py
```diff
@@ -676,17 +675,39 @@ def _get_workspace(workspace_id: int, ids_to_workspace: dict[int, Workspace]) ->
...
+            workspace_client = account_client.get_workspace_client(workspace)
+            ctx = WorkspaceContext(workspace_client)
+            workspace_contexts.append(ctx)
+        return workspace_contexts
 
     def join_collection(
         self,
-        current_workspace_id: int,
+        workspace_ids: list[int],
         target_workspace_id: int | None = None,
```
nfx, line 711: signature is a bit strange - if we have `target_workspace_id` and `workspace_ids` - what's the difference?... how would `ctx.account_workspaces.get_accessible_workspaces()` semantics change?

src/databricks/labs/ucx/install.py
```diff
@@ -676,17 +675,39 @@ def _get_workspace(workspace_id: int, ids_to_workspace: dict[int, Workspace]) ->
...
+        if not acct_ctx.account_workspaces.can_administer(collection_workspace):
+            logger.error(f"User is not workspace admin of collection workspace {collection_workspace_id}")
+            return []
+        installer = self._get_installer(collection_workspace)
+        if installer.config.installed_workspace_ids is None:
+            logger.error(f"No collection info found in the workspace {collection_workspace_id}")
+            return []
+        for workspace_id in installer.config.installed_workspace_ids:
+            workspace = account_client.workspaces.get(workspace_id)
+            if not acct_ctx.account_workspaces.can_administer(workspace):
```
nfx, line 700: shouldn't this move/be integrated with `ctx.account_workspaces.get_accessible_workspaces()`? e.g. you get accessible workspaces once and you assume you can administer those.

src/databricks/labs/ucx/install.py
```diff
@@ -676,17 +675,39 @@ def _get_workspace(workspace_id: int, ids_to_workspace: dict[int, Workspace]) ->
...
-        installation = Installation.current(current_workspace_client, self.product_info.product_name())
-        workspace_info = WorkspaceInfo(installation, current_workspace_client)
-        return workspace_info.load_workspace_info()
+        workspaces = account_client.workspaces.list()
+        ids_to_workspace = {}
+        for workspace in workspaces:
+            ids_to_workspace[workspace.workspace_id] = workspace
+        return ids_to_workspace
+
+    def get_workspace_contexts(self, collection_workspace_id: int) -> list[WorkspaceContext]:
```
nfx, line 686:
```suggestion
    def get_workspace_contexts(self, other_workspace_id: int) -> list[WorkspaceContext]:
```

i wonder if `other workspace` is a better name here to reason about

src/databricks/labs/ucx/install.py
```diff
@@ -676,17 +675,39 @@ def _get_workspace(workspace_id: int, ids_to_workspace: dict[int, Workspace]) ->
...
 
-    def _get_workspace_info(self, current_workspace_id: int):
+    def _get_workspace_ids(self):
         account_client = self._get_safe_account_client()
-        workspace = account_client.workspaces.get(current_workspace_id)
-        current_workspace_client = account_client.get_workspace_client(workspace)
-        installation = Installation.current(current_workspace_client, self.product_info.product_name())
-        workspace_info = WorkspaceInfo(installation, current_workspace_client)
-        return workspace_info.load_workspace_info()
+        workspaces = account_client.workspaces.list()
```
nfx, line 680: looks like a duplicate logic to ctx.account_workspaces.get_accessible_workspaces()

nfx (changes requested):

src/databricks/labs/ucx/install.py
HariGS-DB: So join collection is invoked from cli command. You can supply a list of workspace id in the workspace_ids variable which you want to join to a collection. target_workspace_id is the workspace to which's collection you want to join.
Ex if target_workspace_id (111) has a collection with workspaces ids 111,222,333. You want to join new set of workspace id to the collection (555,666) Then you pass [555,666] to workspace_ids and 111 to target_workspace_id and the new collection will be [111,222,333,444,555].

The above works when you are passing a target_workspace_id. If you are not passing, then code will run the get_accessible_workspace and get the list of eligible workspace(which you are admin on) which you can join as a collection.

nfx: > target_workspace_id is the workspace to which's collection you want to join.

do we need a "target workspace"? isn't the supplied list of workspaces a target? the point of this command is to make sure that all workspaces in a collection know about other workspaces. effectively modifying IDs everywhere.

isn't the intended flow as:

1. get list of every other workspaces
2. for every other workspace, get a list of what they know of other workspaces
3. combine all these workspace ids
4. for every workspace id, modify configuration of UCX and put the same workspace ids everywhere

HariGS-DB: ok. so in my example above, workspace 555 and 666 are not part of any collection. workspaces 111,222,333 are part of a collection. so we run a command saying join collection with workspace_ids as [111,555,666] and resultant will be all 5 workspaces will be part of the collection.?

nfx: > resultant will be all 5 workspaces will be part of the collection

yes, isn't that the goal? the less ambiguous arguments the function has - the better.

HariGS-DB: Updated the logic to reflect this. when run from cli it will pass the list of workspace id. Run doing it as part of the ucx installation pass a boolean value it tell users to prompt for joining collection and also lists the existing installation to join as collection

src/databricks/labs/ucx/install.py
HariGS-DB: have used account_workspaces.get_accessible_workspace() to get list of accessible workspace and using that to validate if user can join a collection

src/databricks/labs/ucx/install.py
HariGS-DB: updated

src/databricks/labs/ucx/install.py
HariGS-DB: done

src/databricks/labs/ucx/install.py
```diff
@@ -691,135 +692,102 @@ def install_on_account(self):
...
-        except KeyError:
-            msg = "Current workspace is not known, Please run as account-admin: databricks labs ucx sync-workspace-info"
-            raise KeyError(msg) from None
+    def _validate_collection(workspace_ids: list[int], ids_to_workspace: dict[int, Workspace]) -> bool:
+        # validate if the passed list of workspace ids, the user has workspace admin access on
+        # if not, then cant join a collection
+        for workspace_id in workspace_ids:
+            try:
+                ids_to_workspace[workspace_id]
+            except KeyError:
```
nfx, line 724:
```suggestion
            if workspace_id not in ids_to_workspace:
```
why do we need try-catch?..

nfx (approved):
lgtm

labs.yml
```diff
@@ -59,6 +59,9 @@ commands:
 
   - name: ensure-assessment-run
     description: ensure the assessment job was run on a workspace
+    flags:
+      - name: run-as-collection
```
JCZuurmond, line 63: @HariGS-DB : Did you verify the command with this flag? I want a boolean flag in lsql too, but expect this to be a flag requiring an argument

JCZuurmond: I suspect this does not work as expected, see https://github.com/databrickslabs/lsql/pull/235

PR #1936, opened 2024-06-26
Improve the retry strategy used during integration tests

asnare:
## Changes

Integration tests can be decorated with conditions under which they should be retried due to problems with the integration infrastructure. However the existing mechanism defaults to no retries at all if the initial attempt fails but takes longer than the timeout. This is almost certainly not the desired behaviour: being marked for retry is an indication that the first attempt is known to sometimes and that at least one retry should be made.

Th focus of this PR is replacing the (existing) SDK mechanism with a custom version that includes the following changes:

 - In addition to a timeout the number of attempts can be configured:

    - The minimum number of attempts defaults to 2 (=initial attempt + 1 retry).
    - The maximum number of attempts can also be specified. (There is no default.)
 
 - The timeout is now optional, so long as a maximum number of attempts is specified.
 - Improved logging, including:

    - Errors when a successful attempt takes longer than the timeout; this indicates the timeout is too low.
    - Warnings when the remaining time after successful attempt is within 20% of the timeout; this indicates that the timeout probably has insufficient margin.

 - Timeouts are handled using the Python clocks intended for measuring elapsed durations. 

### Tests

- [X] added integration tests
- [X] verified on staging environment (screenshot attached)


tests/integration/retries.py
```diff
@@ -0,0 +1,214 @@
+"""Temporary replacement for databricks.sdk.retries.
+
+After evaluation this can either be removed or moved upstream.
```
JCZuurmond, line 3: Let's merge this with upstream (after agreeing on the implementation)

tests/integration/retries.py
```diff
@@ -0,0 +1,214 @@
...
+        return self._clock.time() + timeout.total_seconds() if timeout is not None else None
+
+    def _check_attempt_duration(self, func: Callable[P, T], attempt_duration: dt.timedelta) -> None:
+        timeout = self._timeout
+        if timeout is not None:
+            if timeout <= attempt_duration:
+                logger.error(
+                    f"Retryable call ({func.__name__}) attempt took {attempt_duration}, longer than retry timeout ({timeout}). Adjust timeout accordingly!"
+                )
+            elif (timeout * 0.80) <= attempt_duration:
```
JCZuurmond, line 131: Maybe it does not really matter, but do we want to log this? We can keep it though

tests/integration/retries.py
```diff
@@ -0,0 +1,214 @@
...
+                logger.warning(
+                    f"Retryable call ({func.__name__}) attempt took {attempt_duration}, within 20% of the specified timeout ({timeout}); please adjust timeout accordingly."
```
JCZuurmond, line 133:
```suggestion
                    f"Retryable call ({func.__name__}) attempt took {attempt_duration}, within 80% of the specified timeout ({timeout}); please adjust timeout accordingly."
```

tests/integration/retries.py
```diff
@@ -0,0 +1,214 @@
...
+                            # raise if exception is not retryable
+                            raise err
+                        retry_reason = maybe_retry_reason
+                        sleep = min(10, attempt)
+
+                attempt += 1
+                logger.debug(f'Retrying (attempt={attempt}): {retry_reason} (sleeping ~{sleep}s)')
+
+                # Save some time upon failure by failing early if the deadline will be exceeded after sleeping.
+                sleep_duration = sleep + random()
```
JCZuurmond, line 183: Maybe move this into the `_check_retries_exhausted`

JCZuurmond: Btw this check if the timeout will happen in less than a second right? How likely is that? I recon very unlikely

tests/integration/retries.py
```diff
@@ -0,0 +1,214 @@
...
+            msg = f'max_attempts=x must be equal or higher to min_attempts ({min_attempts})'
+            raise SyntaxError(msg)
+
+        self._on = on
+        self._is_retryable = is_retryable
+
+        dilated_timeout = timeout * timeout_dilation if timeout is not None else None
+        self._timeout = dilated_timeout
+        self._clock = clock
+        self._min_attempts = min_attempts
```
JCZuurmond, line 102:
```suggestion
        self._min_attempts = max(min_attempts, 1)
```

JCZuurmond (changes requested):
I see this as the main change:

> The prior mechanism defaults to no retries at all if the first attempt fails but takes longer than the timeout. (This is not the desired behaviour: being marked for retry is an indication that the first attempt is know to sometimes fail.)

Could you make that clearer in the PR and code?

tests/integration/retries.py
asnare: I want to log it as a hint that a timeout is too tightly specified; there needs to be a large margin to allow for natural variation the runtime of tests. It's a warning that the attempt was within the timeout _this_ time but there's a good chance it won't always be.

tests/integration/retries.py
asnare: The sleep itself can be up to 11 seconds, but can be higher if the attempt is due to (upstream) throttling by the platform.

This check is about not bothering to do the sleep if we know we're to fail immediately afterwards. The alternative here, and what the original code did, is to do the `sleep()` call here and then check afterwards if the we're past the deadline.

tests/integration/retries.py
asnare: Hmm. That's funny, because 20% is what I intended here and I'm pretty sure it's the correct usage in conjunction with _within_. But when I saw this suggestion I definitely had to think about it, and I think it's ambiguous what the 'within 20%' refers to. I've adjusted the wording to (hopefully!) make it clearer.

tests/integration/retries.py
asnare: Good point; given that this is specified manually I've decide to treat it as an error to force it to be fixed instead of just allowing (and ignoring) and incorrect value.

asnare:
> I see this as the main change:
> […]
> Could you make that clearer in the PR and code?

I've adjust the PR description; does this help?

tests/integration/retries.py
JCZuurmond: But does the `random()` increase the likelihood of early catching a timeout significantly?

JCZuurmond:
> > I see this as the main change:
> > […]
> > Could you make that clearer in the PR and code?
> 
> I've adjust the PR description; does this help?

LGTM, but not approving as I want to merge it with upstream and not add this to the tests only

tests/integration/retries.py
asnare: In the literal sense: given a timeout _could_ be caught early it increases the likelihood of it being caught by between 9% and 50%.

That said I think the more important thing to consider here is whether it's worth trying to catch it early at all because it would be simpler to simply sleep and then check the timeout afterwards.

nfx:
@asnare i'm not sure we need it - the idea of `@retried` is to set the expected maximum time to run. If it fails for the first time, it'll get retried again in isolation by the anti-flake part of the infra https://github.com/databrickslabs/sandbox/blob/main/acceptance/ecosystem/runner.go#L55-L67

JCZuurmond (commented):
Need to discuss what to do with the retried

asnare:
The handling of timeouts with the existing/internal retry handler is not going to be changed.

PR #1944, opened 2024-06-27
Group renaming: wait for consistency before completing task

asnare:
## Changes

Renaming groups is eventually consistent. This PR updates the task that renames workspace groups so that it does not complete until the renaming is visible via direct get on the group and via group enumeration.

The integration tests already dealt with the eventual consistency problem, but downstream tasks in the workflow might also run into problems and this wasn't being handled. Ensuring that renaming has completed before the task prevents this from being an issue.

### Functionality 

- [X] modified existing workflow: `group-migration`

### Tests

- [X] added unit tests

### DIFF
--- a/src/databricks/labs/ucx/workspace_access/groups.py
+++ b/src/databricks/labs/ucx/workspace_access/groups.py
@@ -436,7 +436,7 @@ class GroupManager(CrawlerBase[MigratedGroup]):
             logger.info(f"Renaming: {migrated_group.name_in_workspace} -> {migrated_group.temporary_name}")
             tasks.append(
                 functools.partial(
-                    self._rename_group_and_wait,
+                    self._rename_group_and_wait_for_rename,
                     migrated_group.id_in_workspace,
                     migrated_group.name_in_workspace,
                     migrated_group.temporary_name,
--- a/src/databricks/labs/ucx/workspace_access/groups.py
+++ b/src/databricks/labs/ucx/workspace_access/groups.py
@@ -449,7 +449,7 @@ class GroupManager(CrawlerBase[MigratedGroup]):
         # present.
         self._wait_for_renamed_groups(renamed_groups)
 
-    def _rename_group_and_wait(self, group_id: str, old_group_name, new_group_name: str) -> tuple[str, str]:
+    def _rename_group_and_wait_for_rename(self, group_id: str, old_group_name, new_group_name: str) -> tuple[str, str]:
         logger.debug(f"Renaming group {group_id}: {old_group_name} -> {new_group_name}")
         self._rename_group(group_id, new_group_name)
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -425,18 +435,62 @@ def rename_groups(self):
...
-        if len(errors) > 0:
-            raise ManyError(errors)
+        renamed_groups = Threads.strict("rename groups in the workspace", tasks)
+        # Renaming is eventually consistent, and the tasks above have each polled to verify their rename completed.
+        # Here we also check that enumeration yields the updated names; this is necessary because otherwise downstream
+        # tasks (like reflect_account_groups_on_workspace()) may skip a renamed group because it doesn't appear to be
+        # present.
+        self._wait_for_renamed_groups(renamed_groups)
+
+    def _rename_group_and_wait(self, group_id: str, old_group_name, new_group_name: str) -> tuple[str, str]:
```
JCZuurmond, line 452:
```suggestion
    def _rename_group_and_wait_for_rename(self, group_id: str, old_group_name, new_group_name: str) -> tuple[str, str]:
```

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -425,18 +435,62 @@ def rename_groups(self):
...
+        logger.debug(f"Renaming group {group_id}: {old_group_name} -> {new_group_name}")
```
JCZuurmond, line 453: Maybe move this into the `_rename_group`

JCZuurmond (approved):
Some minor changes

src/databricks/labs/ucx/workspace_access/groups.py
asnare: I considered that, but it would move it within the scope of the retry so that upon retry you'd end up with repeated log entries stating the same thing. At this log level there's already logging from the retry handler (assuming that it's at the same log level) that shows what's going on.

src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -425,18 +435,62 @@ def rename_groups(self):
...
+            # Rename still pending.
+            raise GroupRenameIncompleteError(group_id, old_group_name, new_group_name)
+        if group.display_name != new_group_name:
+            # Group has an entirely unexpected name; something else is interfering.
+            msg = f"While waiting for group {group_id} rename ({old_group_name} -> {new_group_name} an unexpected name was observed: {group.display_name}"
+            raise RuntimeError(msg)
+        # Normal exit; group has been renamed.
+
+    @retried(on=[ManyError], timeout=timedelta(minutes=2))
+    def _wait_for_renamed_groups(self, expected_groups: Collection[tuple[str, str]]) -> None:
```
nfx, line 478: i like it. that's neat

nfx (approved):
lgtm

PR #1945, opened 2024-06-27
Revise integration test timeoutsIntegration test timeouts

asnare:
## Changes

Some of our integration tests have been set up to retry when specific failures occur make them more resilient in the face of timeouts and race conditions. One of the retry settings is a timeout, which does not interrupt a running test but rather specifies the deadline after which a new retry attempt will not be made.

For some of these tests, a single attempt can take longer than the configured timeout, especially when running over VPN. This defeats the purpose of the retry setup for that test: it's there because failures are known to occur, but because the attempt takes longer than the timeout deadline a second attempt will never be made.

This PR updates the timeouts for tests where the single attempt was noted as being within 20% (or longer) of the deadline. Increasing the timeouts means that the retry behaviour will be more likely to have its intended effect.


tests/integration/test_runtime.py
```diff
@@ -15,7 +15,7 @@
 logger = logging.getLogger(__name__)
 
 
-@retried(on=[NotFound, InvalidParameterValue], timeout=timedelta(minutes=5))
+@retried(on=[NotFound, InvalidParameterValue], timeout=timedelta(minutes=20))
```
JCZuurmond, line 18: This is getting long. I understand it is a "do not blame the messenger", the changes in this PR do not make the test run longer, but make it visible that it runs longer. Still something to keep an eye on

JCZuurmond (changes requested):
@asnare : It looks like we are missing the change you made to the `retried` fixture. Or am I missing something?

JCZuurmond:
Retried is done here: https://github.com/databrickslabs/ucx/pull/1936

JCZuurmond (approved):

tests/integration/test_runtime.py
asnare: Agreed.

And often this is an indication of a problem with something under test rather than the  test itself. In particular there's normally something within the end-to-end flow that needs to be configured for retry with a (much smaller!) timeout. However because that's missing the whole test fails and therefore the whole thing needs to be restarted from scratch.

nfx, line 18:
```suggestion
@retried(on=[NotFound, InvalidParameterValue], timeout=timedelta(minutes=8))
```

tests/integration/test_runtime.py
```diff
@@ -35,7 +35,7 @@ def test_running_real_assessment_job(ws, installation_ctx, make_cluster_policy,
     assert after[ws_group_a.display_name] == PermissionLevel.CAN_USE
 
 
-@retried(on=[NotFound, InvalidParameterValue], timeout=timedelta(minutes=5))
+@retried(on=[NotFound, InvalidParameterValue], timeout=timedelta(minutes=20))
```
nfx, line 38:
```suggestion
@retried(on=[NotFound, InvalidParameterValue], timeout=timedelta(minutes=8))
```

tests/integration/test_runtime.py
```diff
@@ -122,7 +122,7 @@ def test_running_real_validate_groups_permissions_job(
     installation_ctx.deployed_workflows.run_workflow("validate-groups-permissions")
 
 
-@retried(on=[NotFound], timeout=timedelta(minutes=5))
+@retried(on=[NotFound], timeout=timedelta(minutes=15))
```
nfx, line 125:
```suggestion
@retried(on=[NotFound], timeout=timedelta(minutes=8))
```

nfx (approved):
we can't raise timeouts indefinitely as tests would run for too long

PR #1961, opened 2024-07-01
Infer values across cells

ericvergnaud:
## Changes
When linting python code, infer values using not only code from current cell but also code from previous cells

### Linked issues
Progresses #1912 

### Functionality 
None

### Tests
- [x] manually tested
- [x] added unit tests

Resolved 60 our of 891 "cannot be computed" advices when running `make solacc`


ericvergnaud:
superseded by #1968

PR #1964, opened 2024-07-02
Adjust type hints in credentials handling for Azure API client

asnare:
## Changes

This PR is stacked on #1943; it is stacked to verify the type hinting changes via the CI environment.

The situation is:

 - The Databricks SDK has partially implemented type hints for the credentials handlers.
 - The `CredentialsProvider` is the fundamental type for a callable factory that can be used to obtain credentials.
 - The `CredentialsStrategy` is the type that can be used for a function that can be used to obtain a provider, given the `Config`.
 - The `@credentials_strategy` decorator is used to fuse the two:

   1. It decorates a function that, given a `Config`, returns a `CredentialsProvider`.
   2. The decorated function that it returns is a `CredentialsStrategy`.

 - The type upstream type hinting for the `@credentials_strategy` decorator is incomplete: the return type is omitted. This confuses IntelliJ/PyCharm a little but not Mypy.

### DIFF
--- a/src/databricks/labs/ucx/azure/resources.py
+++ b/src/databricks/labs/ucx/azure/resources.py
@@ -8,6 +8,7 @@ from databricks.sdk.core import (
     ApiClient,
     AzureCliTokenSource,
     Config,
+    CredentialsProvider,
     CredentialsStrategy,
     credentials_strategy,
 )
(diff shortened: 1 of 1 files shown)


JCZuurmond (approved):
LGTM, thanks

PR #1966, opened 2024-07-02
Added tracking for UCX workflows and as-library usage

nfx:
This PR adds `User-Agent` OtherInfo elements:
* `ucx/<version>` for all requests where UCX is imported as a library
* `cmd/install` for all installations/upgrades
* `cmd/<workflow>` for all managed job invocations

Relevant PRs:
- https://github.com/databrickslabs/lsql/pull/206
- https://github.com/databrickslabs/blueprint/pull/114

### DIFF
--- a/src/databricks/labs/ucx/__init__.py
+++ b/src/databricks/labs/ucx/__init__.py
@@ -1,7 +1,11 @@
-from databricks.sdk.core import with_user_agent_extra
+from databricks.sdk.core import with_user_agent_extra, with_product
 from databricks.labs.blueprint.logger import install_logger
 from databricks.labs.ucx.__about__ import __version__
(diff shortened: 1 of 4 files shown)


src/databricks/labs/ucx/__init__.py
```diff
@@ -1,3 +1,7 @@
+from databricks.sdk.core import with_user_agent_extra
 from databricks.labs.blueprint.logger import install_logger
+from databricks.labs.ucx.__about__ import __version__
 
 install_logger()
+
+with_user_agent_extra("ucx", __version__)
```
nfx, line 7: this may affect the SDK usage attribution once UCX rolls out as LSP

PR #1971, opened 2024-07-02
manually whitelist core.dbclient and core.logging_utils (from DB SAT)

ericvergnaud:
## Changes
manually whitelist core.dbclient and core.logging_utils (from DB SAT)

### Linked issues
Progresses #1901 

### Functionality 
None

### Tests
- [x] manually tested


nfx:
DB SAT is a prototype, we don't expect it to be used by customers

PR #2009, opened 2024-07-02
whitelist dbl-sat-sdk

ericvergnaud:
## Changes
whitelist dbl-sat-sdk

### Linked issues
Progresses #1901 

### Functionality 
None

### Tests
- [x] manually tested


nfx:
this is a joke library, we don't expect anyone to use it

PR #2020, opened 2024-07-02
Yield `DependencyProblem` if job on runtime DBR14+ and using .egg dependency

### LINKED ISSUE #1793: [FEATURE] Create DependencyProblem if job on runtime DBR14+ and using .egg dependency
I am not able to use `easy_install` as the [setuptools documentation mentions](https://setuptools.pypa.io/en/latest/deprecated/python_eggs.html). `easy_install` can not be found anymore, which is a bit odd as the setuptools change log mention it is deprecated in [v58.3.0](https://setuptools.pypa.io/en/stable/history.html#v58-3-0). I tried a version before that and no luck, maybe python 3.10+ version is causing the issue.

I tried in Databricks as well, but no `easy_install` there either, while up to runtime 14.0 [Python eggs are supported](https://learn.microsoft.com/en-us/azure/databricks/archive/legacy/workspace-libraries#--upload-a-jar-python-egg-or-python-wheel-file).

Screenshot 2024-05-29 at 11 13 43

Another route would be to use the `setup.py` from the (unzipped) egg. However, this is not always present, for example, the one I am using for testing.

The unzipping does let the integration test pass, though, installing would be a preferred route

_Originally posted by @JCZuurmond in https://github.com/databrickslabs/ucx/pull/1789#discussion_r1618530889_

pritishpai:
## Changes
Create DependencyProblem if job on runtime DBR14+ and using .egg dependency

### Linked issues

Fix #1793

### DIFF
--- a/src/databricks/labs/ucx/source_code/jobs.py
+++ b/src/databricks/labs/ucx/source_code/jobs.py
@@ -124,16 +124,7 @@ class WorkflowTaskContainer(SourceContainer):
             if problems:
                 yield from problems
         if library.egg:
-            if self.runtime_version > (14, 0):
-                yield DependencyProblem(
-                    'not-supported', 'Installing eggs is no longer supported on Databricks 14.0 or higher'
-                )
-            logger.info(f"Registering library from {library.egg}")
-            with self._ws.workspace.download(library.egg, format=ExportFormat.AUTO) as remote_file:
-                with tempfile.TemporaryDirectory() as directory:
-                    local_file = Path(directory) / Path(library.egg).name
-                    local_file.write_bytes(remote_file.read())
-                    yield from graph.register_library(local_file.as_posix())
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/source_code/jobs.py
```diff
@@ -124,6 +124,10 @@ def _register_library(self, graph: DependencyGraph, library: compute.Library) ->
             if problems:
                 yield from problems
         if library.egg:
+            if self.runtime_version > (14, 0):
+                yield DependencyProblem(
+                    'not-yet-implemented', 'Installing eggs is no longer supported ' 'on Databricks 14.0 or higher'
```
nfx, line 129:
```suggestion
                    'not-supported', 'Installing eggs is no longer supported on Databricks 14.0 or higher'
```

nfx (approved):

PR #2025, opened 2024-07-03
Exclude VIEW from "Non-DELTA format: UNKNOWN" findings in assessment summary chart

qziyuan:
## Changes
The count of views is included in `Non-DELTA format: UNKNOWN` finding in the `Assessment Summary` chart in the assessment main dashboard. The customers were confused as they cannot find any unknown format tables but eventually find they are all views.
This PR will exclude views from `Non-DELTA format: UNKNOWN` finding.
image


nfx (approved):
lgtm

PR #2071, opened 2024-07-03
whitelist openai

ericvergnaud:
## Changes
whitelist openai

### Linked issues
Progresses #1901 

### Functionality 
None

### Tests
- [x] manually tested


src/databricks/labs/ucx/source_code/known.json
```diff
@@ -7323,6 +7323,278 @@
     "pydevd_plugins.extensions": [],
     "pydevd_plugins.extensions.pydevd_plugin_omegaconf": []
   },
+  "openai": {
```
JCZuurmond, line 7326: @ericvergnaud : OpenAI probably does not work with Spark. Does it make sense to include it in that case?

src/databricks/labs/ucx/source_code/known.json
```diff
@@ -7323,6 +7323,278 @@
...
+    "openai": [],
+    "openai._base_client": [],
```
JCZuurmond, line 7328: Do we keep the hidden modules?

src/databricks/labs/ucx/source_code/known.json
JCZuurmond: Or maybe only the top level to avoid linting?

JCZuurmond (changes requested):
See comments

src/databricks/labs/ucx/source_code/known.json
ericvergnaud: It's here because it's used in solacc.
There's no concept of 'top-level'. If you `import openai.cli`, ucx looks for openai.cli, not for openai. The top-level key is the name of the distribution, not of a module.

src/databricks/labs/ucx/source_code/known.json
ericvergnaud: @nfx what do you think ?

PR #2075, opened 2024-07-03
Infer values from child notebook in run cell

### LINKED ISSUE #1927: [FEATURE]: Track variables across `%run` during linting

### Problem statement

During linting we currently follow %run for tracking notebook dependencies, but do not track variable references which means the value inference misses things.

When %run happens:

 - The local globals are exposed to the target notebook.
 - The updated globals are passed back to the calling notebook when the target finishes.

This is often used for common settings across notebooks, such as datasets paths and the like.

### Proposed Solution

During linting the state of globals should be tracked through %run magic.

### Additional Context

_No response_

ericvergnaud:
## Changes
Infer values from child notebook in run cell

### Linked issues
Progresses #1901 
Progresses #1205 
Fixes #1927 

### Functionality 
None

### Tests
- [x] added unit tests

Solves 87 'not computed' advices when running `make solacc`


src/databricks/labs/ucx/source_code/notebooks/cells.py
```diff
@@ -195,6 +187,19 @@ def build_dependency_graph(self, parent: DependencyGraph) -> list[DependencyProb
         )
         return [problem]
 
+    def read_notebook_path(self):
```
nfx, line 190:
```suggestion
    def maybe_notebook_path() -> Path | None:
      ... = self._read_notebook_path()
      
    def _read_notebook_path(self):
```

return type interface exposes too much unnecessary private information, make this method private and wrap it with public method with only what we need.

```
def _load_source_from_run_cell(self, run_cell: RunCell):
        path, _, _ = run_cell.read_notebook_path()
```

src/databricks/labs/ucx/source_code/notebooks/sources.py
```diff
@@ -111,7 +125,29 @@ def lint(self) -> Iterable[Advice]:
...
+        return self._context.linter(language)
+
+    def _load_source_from_run_cell(self, run_cell: RunCell):
+        path, _, _ = run_cell.read_notebook_path()
+        if path is None:
+            return  # malformed run cell already reported
+        resolved = self._path_lookup.resolve(path)
+        if resolved is None:
+            return  # already reported during dependency building
+        # TODO deal with workspace notebooks
```
nfx, line 137: `resolved` already has workspace notebook because of WorkspacePath.

nfx (changes requested):

src/databricks/labs/ucx/source_code/notebooks/sources.py
ericvergnaud: noted.

src/databricks/labs/ucx/source_code/notebooks/cells.py
ericvergnaud: done

nfx (approved):
lgtm

PR #2083, opened 2024-07-04
Infer values from child notebook loaded via dbutils.notebook.run() call

ericvergnaud:
## Changes
When linting notebooks, use AST from child notebooks loaded via `dbutils.notebook.run()` to infer values

### Linked issues
Progresses #1205 
Progresses #1901 

### Functionality 
None

### Tests
- [x] added unit tests

marginally improves _not computed_ (-9/744) when running `make solacc`

PR #2091, opened 2024-07-05
Infer values from child notebook in magic line

### LINKED ISSUE #1201: [FEATURE]: Detect notebook include graph by analysing `%run` magic cells

ericvergnaud:
## Changes
When linting notebooks, use values from child notebooks loaded via %run magic line to improve value inference

### Linked issues
Resolves #1201 
Progresses #1901 

### Functionality
None

### Tests
- [x] added unit tests

Depends on #2083


src/databricks/labs/ucx/source_code/notebooks/cells.py
```diff
@@ -228,7 +229,7 @@ def is_runnable(self) -> bool:
         return True  # TODO
 
     def build_dependency_graph(self, graph: DependencyGraph) -> list[DependencyProblem]:
-        return PipMagic(self.original_code).build_dependency_graph(graph)
+        return PipCommand(self.original_code).build_dependency_graph(graph)
```
ericvergnaud, line 232: Renamed because the magic comes from the prefix, not the command itself

ericvergnaud:
@nfx the failed no-cheat comes from a refactoring, not a new cheat

src/databricks/labs/ucx/source_code/notebooks/cells.py
```diff
@@ -496,27 +494,57 @@ def _make_commands_for_magic_command_call_nodes(cls, nodes: list[Call]):
...
         self._code = code
 
+    @abstractmethod
+    def build_dependency_graph(self, graph: DependencyGraph) -> list[DependencyProblem]: ...
+
+
+class RunCommand(MagicCommand):
+
+    def build_dependency_graph(self, graph: DependencyGraph) -> list[DependencyProblem]:
+        # TODO
```
nfx, line 532: implement this

src/databricks/labs/ucx/source_code/notebooks/sources.py
```diff
@@ -93,35 +112,157 @@ class NotebookLinter:
...
+        self._python_linter.append_nodes(nodes)
+
+    @staticmethod
+    def _list_run_magic_lines(tree: Tree) -> list[MagicLine]:
+
+        def _ignore_problem(_code: str, _message: str, _node: NodeNG) -> None:
+            return None
+
+        commands, _ = MagicLine.extract_from_tree(tree, _ignore_problem)
+        return list(filter(lambda cmd: isinstance(cmd.as_magic(), RunCommand), commands))
```
nfx, line 201: rewrite as for loop, as it'll be difficult to debug once we start getting bug reports about this portion of code

tests/unit/source_code/test_notebook_linter.py
```diff
@@ -585,3 +589,51 @@ def test_computes_values_across_notebooks_using_run(extended_test_index, mock_pa
         )
     ]
     assert advices == expected
+
+
+def test_ignores_values_across_notebooks_using_dbutils_notebook_run(extended_test_index, mock_path_lookup):
```
nfx, line 594: can you rewrite this as a functional test? they're easier to read. and maintain. and write as well.

tests/unit/source_code/test_notebook_linter.py
```diff
@@ -585,3 +589,51 @@ def test_computes_values_across_notebooks_using_run(extended_test_index, mock_pa
...
+            start_line=3,
+            start_col=0,
+            end_line=3,
+            end_col=19,
+        ),
+    ]
+    assert advices == expected
+
+
+def test_computes_values_across_notebooks_using_magic_line(extended_test_index, mock_path_lookup):
```
nfx, line 622: can you rewrite this as a functional test? they're easier to read. and maintain. and write as well.

tests/unit/source_code/samples/values_across_notebooks_magic_line.py
```diff
@@ -0,0 +1,4 @@
+# Databricks notebook source
+
+%run "./values_across_notebooks_child.py"
```
nfx, line 3: is this a valid notebook code? please attach a screenshot.

nfx (changes requested):

src/databricks/labs/ucx/source_code/notebooks/cells.py
ericvergnaud: I was planning to do this in a separate PR since it's unrelated, see https://github.com/databrickslabs/ucx/issues/2155

tests/unit/source_code/samples/values_across_notebooks_magic_line.py
asnare: This is interesting…

It's a valid notebook:
image

However there seems to be an issue[^1] at the moment where this line-magic isn't interpreted correctly in a cell that contains other code:
image
(This is limited to `%run`: things like `%pip` seem to be fine.

It works correctly if split out to a cell on its own, as the other test sample in this PR demonstrates.

[^1]: Also mentioned [here](https://community.databricks.com/t5/data-engineering/cannot-run-another-notebook-from-same-directory/td-p/68627). In Jupyter it's perfectly fine:
image

asnare: I should add that my opinion on whether we should support this is or not comes down to whether the implementation is simpler if we allow it. Given that line-magic like this is allowed in general, I can imagine that _not_ allowing the specific `%run` case might be more work than just leaving it in.

ericvergnaud: Screenshot 2024-07-11 at 11 16 38

ericvergnaud: It works, but there are some limitations:
 - if `path` is a notebook, the line should be the only one in the cell (and should not have an extension)
 - if `path` is a file, it must have an extension, and the line should not be the first one
@asnare FYI

src/databricks/labs/ucx/source_code/notebooks/sources.py
ericvergnaud: Done

src/databricks/labs/ucx/source_code/notebooks/cells.py
```diff
@@ -496,27 +494,57 @@ def _make_commands_for_magic_command_call_nodes(cls, nodes: list[Call]):
...
+
+    @property
+    def notebook_path(self) -> Path | None:
+        for quote in ('"', "'"):
+            start = self._code.find(quote)
+            if start >= 0:
+                end = self._code.find(quote, start + 1)
+                if end >= 0:
+                    return Path(self._code[start + 1 : end])
+        return None
```
asnare, line 543: The quotes for the path areoptional; they're only required if there is whitespace in the path.

(Most of the solacc examples omit the quotes; people seem to only use them when required.)

I'd really like to see a unit test dedicated to just this property, although updating the samples to exercise the variants would also be fine.

It also looks like `RunCell` has an independent implementation of the same thing in `_read_notebook_path()`.

asnare (commented):
Looking at the bigger picture here, I have an observation and some questions.

First the observation: at the moment we seem to treat `%-`magic differently if it's a complete cell versus a command within a cell. I understand that this is mainly historic but think it's starting to cause issues. The situation is really:

   - `%-`magic is (by default) a command within a cell…
   - …except for on the first line of a cell, where instead of being a command it _might_ be (and normally is) a marker for the cell language to override the notebook default.

The bigger questions I have are more specific to this PR:

1. We have tests covering values flowing out of a child notebook, but what about values flowing in? Values defined before the `%run` are visible within the child file.
2. The focus of this PR is inference, but what happens if there are linting findings within a notebook that we `%run`? Are these reported? (I think so?) Should they be reported by the workflow when the child notebook is probably also scanned independently?</br>My feeling on this is that reporting the findings is fine: they may be different due to values flowing in. However when we're scanning lots of notebooks we might need to deduplicate because now the same file is covered multiple times due to each point of inclusion. (Is this deduplication already present?)

tests/unit/source_code/test_notebook_linter.py
ericvergnaud: done

tests/unit/source_code/test_notebook_linter.py
ericvergnaud: done

tests/unit/source_code/samples/values_across_notebooks_magic_line.py
ericvergnaud: See #2155 and #2156

src/databricks/labs/ucx/source_code/notebooks/cells.py
ericvergnaud: Thanks for that. Fixed the optionality. The reason we have a different implementation is because run cells can span multiple lines whereas magic lines can only be one-liners

ericvergnaud:
> Looking at the bigger picture here, I have an observation and some questions.
> 
> First the observation: at the moment we seem to treat `%-`magic differently if it's a complete cell versus a command within a cell. I understand that this is mainly historic but think it's starting to cause issues. The situation is really:
> 
> * `%-`magic is (by default) a command within a cell…
> * …except for on the first line of a cell, where instead of being a command it _might_ be (and normally is) a marker for the cell language to override the notebook default.
> 
> The bigger questions I have are more specific to this PR:
> 
> 1. We have tests covering values flowing out of a child notebook, but what about values flowing in? Values defined before the `%run` are visible within the child file.
> 2. The focus of this PR is inference, but what happens if there are linting findings within a notebook that we `%run`? Are these reported? (I think so?) Should they be reported by the workflow when the child notebook is probably also scanned independently?My feeling on this is that reporting the findings is fine: they may be different due to values flowing in. However when we're scanning lots of notebooks we might need to deduplicate because now the same file is covered multiple times due to each point of inclusion. (Is this deduplication already present?)

Re magic locations I'm not sure we are far from where we should be. If a cell starts with a magic line then we treat it as a cell language.
Re bigger questions, I think they deserve dedicated PRs, see #2155 and #2156. We might be missing a ticket for values flowing in via parameters. We're missing tests for values flowing in via cells, but it should already work.

tests/unit/source_code/samples/functional/values_across_notebooks_dbutils_notebook_run.py
```diff
@@ -0,0 +1,7 @@
+# Databricks notebook source
+
+dbutils.notebook.run("./values_across_notebooks_child.py")
+# ucx[table-migrate:+3:0:+3:19] The default format changed in Databricks Runtime 8.0, from Parquet to Delta
+# dbutils.notebook.run runs in a separate process and thus does NOT import values
```
asnare, line 5: Thanks; I was looking for this. 👍

src/databricks/labs/ucx/source_code/notebooks/cells.py
nfx: it hides a bug of mishandling `%run` magic and command. it's almost `return graph.register_notebook(self.notebook_path)` + error checking. and one more functional test. Implement it in this PR, otherwise it'll be forgotten. you can make a PR to this PR.

tests/unit/source_code/test_functional.py
```diff
@@ -166,4 +166,15 @@ def _comments(f) -> Generator[Comment, None, None]:
...
 def test_functional(sample: Functional, mock_path_lookup) -> None:
-    sample.verify(mock_path_lookup)
+    path_lookup = mock_path_lookup.change_directory(sample.path.parent)
+    sample.verify(path_lookup)
+
+
+@pytest.mark.skip(reason="Used for troubleshooting failing tests")
+def test_one_functional(mock_path_lookup):
+    path = Path(
+        "/Users/eric.vergnaud/development/ucx/tests/unit/source_code/samples/functional/values_across_notebooks_magic_line.py"
```
nfx, line 176: no hardcoded absolute paths.

tests/unit/source_code/samples/functional/values_across_notebooks_dbutils_notebook_run.py
nfx: this is way more maintainable

nfx (changes requested):

asnare:
> […] Re bigger questions, I think they deserve dedicated PRs, see #2155 and #2156. We might be missing a ticket for values flowing in via parameters. We're missing tests for values flowing in via cells, but it should already work.

I also think that values flowing into the child notebook should work, but would like to see the tests covering it as part of this PR if it works. Just so we know. :) (If it doesn't work, I'm fine with a new PR to deal with it.)

Those PRs are good for covering the linting of child code, although there's still the higher-level issue about how to deal with the noise of duplicate findings that it will introduce at the workflow level.

tests/unit/source_code/test_functional.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/notebooks/cells.py
ericvergnaud: I agree. I'll prioritize #2155

ericvergnaud: It's done in PR #2164

nfx (approved):
lgtm

PR #2092, opened 2024-07-05
Add more checks for spark-connect linter

vsevolodstep-db:
## Changes
Add more checks to detect code incompatibilities with UC Shared Clusters:
- use of Python UDF unsupported eval types
- spark.catalog.X APIs on DBR < 14.3
- use of commandContext

### Tests
- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/source_code/linters/context.py
```diff
@@ -21,7 +21,13 @@
 
 
 class LinterContext:
-    def __init__(self, index: MigrationIndex | None = None, session_state: CurrentSessionState | None = None):
+    def __init__(
+        self,
+        index: MigrationIndex | None = None,
+        session_state: CurrentSessionState | None = None,
+        dbr_version: tuple[int, int] | None = None,
+        is_serverless: bool = False,
```
nfx, line 29: can we move `is_serverless` to `CurrentSessionState`? also, please modify the jobs.py to lint serverless jobs properly

src/databricks/labs/ucx/source_code/linters/context.py
```diff
@@ -21,7 +21,13 @@
 
 
 class LinterContext:
-    def __init__(self, index: MigrationIndex | None = None, session_state: CurrentSessionState | None = None):
+    def __init__(
+        self,
+        index: MigrationIndex | None = None,
+        session_state: CurrentSessionState | None = None,
+        dbr_version: tuple[int, int] | None = None,
```
nfx, line 28: make dbr version a property of `CurrentSessionState` as well. we'll have few more and we should have just one container to pass along, not many parameters we may forget about.

tests/unit/source_code/samples/functional/spark-connect/catalog-api_13_3.py
```diff
@@ -0,0 +1,11 @@
+# ucx[catalog-api-in-shared-clusters:+1:0:+1:13] spark.catalog functions require DBR 14.3 LTS or above on UC Shared Clusters
+spark.catalog.tableExists("table")
+# ucx[catalog-api-in-shared-clusters:+1:0:+1:13] spark.catalog functions require DBR 14.3 LTS or above on UC Shared Clusters
+spark.catalog.listDatabases()
+
+
+def catalog():
```
nfx, line 7: add more false positives

```
class Fatalog:
   def tableExists(self, x): ...
class Foo:
   def catalog(self): Fatalog()

x = Foo()
x.catalog.tableExists("...")
```

tests/unit/source_code/test_functional.py
```diff
@@ -62,6 +62,13 @@ class Functional:
     )
     _location = Path(__file__).parent / 'samples/functional'
 
+    TEST_DBR_VERSION = {
+        'python-udfs_13_3.py': (13, 3),
```
nfx, line 66: don't you want to introduce `# ucx session state: {"dbr_version": [13, 3]}` / `# ucx session state: {"is_serverless": true}` meta-comments, where you parse JSON and set override the current session state?.. make sure to validate this instruction appears only once per file, because of the way we initialise linters. 

technically, we can make it into processing instrustion and be in the parser flow, but it's more complicated.

src/databricks/labs/ucx/source_code/linters/spark_connect.py
```diff
@@ -172,14 +172,84 @@ def _match_jvm_log(self, node: NodeNG) -> Iterator[Advice]:
...
+
+
+class CommandContextMatcher(SharedClusterMatcher):
+    def lint(self, node: NodeNG) -> Iterator[Advice]:
+        if not isinstance(node, Call):
+            return
+        function_name = Tree.get_full_function_name(node)
+        if function_name and function_name.endswith('getContext.toJson'):
+            yield Failure.from_node(
+                code='toJson-in-shared-clusters',
```
nfx, line 234: use lowercase for codes, as i suspect that's how LSP would normalise everything.

nfx (changes requested):

tests/unit/source_code/test_functional.py
```diff
@@ -102,39 +106,54 @@ def _lint(self) -> Iterable[Advice]:
                 MigrationStatus('other', 'matters', dst_catalog='some', dst_schema='certain', dst_table='issues'),
             ]
         )
-        session_state = CurrentSessionState()
+        session_state = self._test_session_state()
+        print(str(session_state))
```
nfx, line 110: use logging, not print statement. i wonder why linter didn't catch it.

nfx (approved):
lgtm

PR #2101, opened 2024-07-08
Allow-listing language update

asnare:
## Changes

This PR updates the language we use for allow-listing functionality. Some public APIs are affected, but I think the compatibility issues here are acceptable.

### DIFF
--- a/src/databricks/labs/ucx/source_code/known.py
+++ b/src/databricks/labs/ucx/source_code/known.py
@@ -63,7 +63,7 @@ UNKNOWN = Compatibility(False, [])
 _DEFAULT_ENCODING = sys.getdefaultencoding()
 
 
-class Whitelist:
+class AllowList:
     def __init__(self):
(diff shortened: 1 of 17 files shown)


src/databricks/labs/ucx/source_code/known.py
```diff
@@ -63,7 +63,7 @@ def as_dict(self):
 _DEFAULT_ENCODING = sys.getdefaultencoding()
 
 
-class Whitelist:
+class AllowList:
```
nfx, line 66: can you rename this class to `Known`? allow_list is fine for varnames.

nfx (changes requested):

src/databricks/labs/ucx/source_code/known.py
asnare: Yeah that's a better name, thanks. Done.

nfx (approved):
lgtm

PR #2127, opened 2024-07-10
improve no-cheat CI check

ericvergnaud:
## Changes
improve no-cheat CI check

Currently we get blocked when we change a line of code that already contains a cheat. 
For example, changing from:
`some_function("abc") # pylint: disable=some-code`
to:
`some_function("xyz") # pylint: disable=some-code`
triggers a no-cheat erroneously.
We want to avoid that, and also get better feedback on the new cheats.

The current script diffs the incoming branch with main, not the target branch. 
This PR changes that by using `$GITHUB_BASE_REF`.

The logic is too complex for bash, so delegating the analysis to python.

### Linked issues
None

### Functionality
None

### Tests
- [x] manually tested

### DIFF
--- a/.github/workflows/no-cheat.yml
+++ b/.github/workflows/no-cheat.yml
@@ -23,8 +23,10 @@ jobs:
       - name: Verify no additional disabled lint in the new code
         run: |
           DIFFS=$(git diff origin/main..$(git branch --show-current))
+          echo $DIFFS
           RESULTS=$(python tests/unit/no_cheat.py ${DIFFS})
-          if [ "${#RESULTS}" -gt 0 ]; then
+          echo ${RESULTS}
+          if [ ${#RESULTS} -gt 0 ]; then
             echo ${RESULTS}
             exit 1
           fi
--- a/.github/workflows/no-cheat.yml
+++ b/.github/workflows/no-cheat.yml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Verify no additional disabled lint in the new code
         run: |
-          git diff origin/main..$(git branch --show-current) >> diff_data.txt
+          git diff origin/main...$(git branch --show-current) >> diff_data.txt
           python tests/unit/no_cheat.py diff_data.txt >> cheats.txt
           COUNT=$(cat cheats.txt | wc -c)
           if [ ${COUNT} -gt 0 ]; then
--- a/.github/workflows/no-cheat.yml
+++ b/.github/workflows/no-cheat.yml
@@ -24,6 +24,7 @@ jobs:
         run: |
           git diff origin/main..$(git branch --show-current) >> diff_data.txt
           python tests/unit/no_cheat.py diff_data.txt >> cheats.txt
+          cat cheats.txt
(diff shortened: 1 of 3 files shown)


ericvergnaud:
@nfx the 2 new cheats come from the no-cheat.yml file, which shows that the script works

.github/workflows/no-cheat.yml
```diff
@@ -20,11 +20,14 @@ jobs:
...
         run: |
-          NEW_CODE=$(git diff origin/main..$(git branch --show-current) | grep -e '^+')
-          CHEAT=$(echo "${NEW_CODE}" | grep '# pylint: disable' | grep -v "CHEAT" | wc -c)
-          if [ "${CHEAT}" -ne 0 ]; then
-            echo "Do not cheat the linter: ${CHEAT}"
+          MAIN_CODE=$(git diff origin/main..$(git branch --show-current) | grep -e '^-')
+          MAIN_CHEATS=$(echo "${MAIN_CODE}" | grep -c '# pylint: disable')
+          BRANCH_CODE=$(git diff origin/main..$(git branch --show-current) | grep -e '^+')
+          BRANCH_CHEATS=$(echo "${BRANCH_CODE}" | grep -c '# pylint: disable')
+          NEW_CHEATS=$((BRANCH_CHEATS-MAIN_CHEATS))
```
nfx, line 29: this would be incorrect when 

```
- ... # pylint: disable=foo
+ ... # pylint: disable=bar
```

🤷‍♂️ probably comparing on $1 from `^.* # pylint: disable=([\w-]+)$` is sufficient then. please create a PR into this PR to demonstrate that this case works.

nfx (changes requested):

.github/workflows/no-cheat.yml
ericvergnaud: I've moved the analysis to python

ericvergnaud: PR #2154 uses this script successfully, see https://github.com/databrickslabs/ucx/actions/runs/9878593859/job/27282880063?pr=2154

Change in tests/integration/conftest.py triggers:
Do not cheat the linter: found 1 additional # pylint: disable=some-code

Changes in tests/unit/workspace_access/test_listing.py do not trigger a cheat

Other cheats are triggered by code in this PR

JCZuurmond (commented):
Could you elaborate the intent of the PR? What is "improve"?

I think you only want to check on the created lines, would a grep for lines in the diff starting with `+` not suffice instead of introducing python code?

ericvergnaud:
> Could you elaborate the intent of the PR? What is "improve"?
> 
> I think you only want to check on the created lines, would a grep for lines in the diff starting with `+` not suffice instead of introducing python code?
> 
> Quick seach yields: `git diff -U0 <target> | grep '^+'`

See updated PR description

ericvergnaud:
Created 2 test PRs #2154 and #2158 that produce the expected outcome
Screenshot 2024-07-11 at 17 30 30
Screenshot 2024-07-11 at 17 30 55

PR #2137, opened 2024-07-10
whitelist h11

ericvergnaud:
## Changes
whitelist h11

### Linked issues
Progresses #1901

### Functionality
None

### Tests
- [x] manually tested

### DIFF
--- a/src/databricks/labs/ucx/source_code/known.json
+++ b/src/databricks/labs/ucx/source_code/known.json
@@ -4434,6 +4434,19 @@
   "gunicorn": {
     "gunicorn": []
   },
+  "h11": {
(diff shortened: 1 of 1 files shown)


src/databricks/labs/ucx/source_code/known.json
```diff
@@ -4434,6 +4434,19 @@
   "gunicorn": {
     "gunicorn": []
   },
+  "h11": {
+    "h11": [],
+    "h11._abnf": [],
```
JCZuurmond, line 4439: Same question about the hidden modules, only hidden modules btw

JCZuurmond (commented):
See comment

PR #2140, opened 2024-07-10
whitelist jsonschema and jsonschema-specifications

ericvergnaud:
## Changes
whitelist jsonschema and jsonschema-specifications

### Linked issues
Progresses #1901

### Functionality
None

### Tests
- [x] manually tested


src/databricks/labs/ucx/source_code/known.json
```diff
@@ -5149,6 +5149,24 @@
...
+    "jsonschema._format": [],
+    "jsonschema._keywords": [],
+    "jsonschema._legacy_keywords": [],
+    "jsonschema._types": [],
+    "jsonschema._typing": [],
+    "jsonschema._utils": [],
+    "jsonschema.cli": [],
+    "jsonschema.exceptions": [],
+    "jsonschema.protocols": [],
+    "jsonschema.tests": [],
```
JCZuurmond, line 5163: Remove

src/databricks/labs/ucx/source_code/known.json
```diff
@@ -5149,6 +5149,24 @@
...
   },
+  "jsonschema": {
+    "jsonschema": [],
+    "jsonschema._format": [],
+    "jsonschema._keywords": [],
+    "jsonschema._legacy_keywords": [],
+    "jsonschema._types": [],
+    "jsonschema._typing": [],
+    "jsonschema._utils": [],
+    "jsonschema.cli": [],
```
JCZuurmond, line 5160: Also looks unnecessary, but maybe just keep it

src/databricks/labs/ucx/source_code/known.json
```diff
@@ -5149,6 +5149,24 @@
...
+    "jsonschema.exceptions": [],
+    "jsonschema.protocols": [],
+    "jsonschema.tests": [],
+    "jsonschema.validators": []
+  },
+  "jsonschema-specifications": {
```
JCZuurmond, line 5166: Looks like an internally used package.

JCZuurmond (commented):
See comments

src/databricks/labs/ucx/source_code/known.json
ericvergnaud: What would be the rationale for removing these modules? We're adding them to improve linting speed.

src/databricks/labs/ucx/source_code/known.json
ericvergnaud: No it's not. Check https://pypi.org/project/jsonschema-specifications/

src/databricks/labs/ucx/source_code/known.json
JCZuurmond: I think we should aim to keep the `know.json` relevant to the code that users use so that it does not blow up. The same for hidden modules, a user can import it, but (s)he probably does not.

ericvergnaud: Indeed it doesn't hur,  `known.json` is only loaded once. `jsonschema` is a very popular package, see statistics here: https://pypi.org/project/jsonschema/

JCZuurmond (approved):

PR #2146, opened 2024-07-10
whitelist toolz

ericvergnaud:
## Changes
whitelist toolz

### Linked issues
Progresses #1901

### Functionality
None

### Tests
- [x] manually tested


src/databricks/labs/ucx/source_code/known.json
```diff
@@ -26274,6 +26274,25 @@
   "tomlkit": {
     "tomlkit": []
   },
+  "toolz": {
+    "tlz": [],
```
JCZuurmond, line 26278: Same question about the `tlz` not matching `toolz` as in other Pr

JCZuurmond (commented):
See comment

src/databricks/labs/ucx/source_code/known.json
ericvergnaud: We don't control this. The pypi package is `toolz`, it contains a `tlz` module, see https://github.com/pytoolz/toolz/

JCZuurmond: yes, I understand that, but want to confirm that our logic will still work when this is the case

ericvergnaud: this was tested by running `make solacc` so I confirm

JCZuurmond (approved):

PR #2159, opened 2024-07-11
Fix infinite recursion when checking module of expression

ericvergnaud:
## Changes
Don't drop existing statements for globals when appending trees

### Linked issues
None

### Functionality
None

### Tests
- [x] added unit tests

Was discovered when running make solacc. Fixes 1 of the 2 issue. Other issue is fixed by #2157


nfx (approved):
lgtm

PR #2200, opened 2024-07-17
Minor grammar corrections in logged warnings

asnare:
## Changes

This PR corrects some grammar in logged warnings.

### DIFF
--- a/src/databricks/labs/ucx/source_code/jobs.py
+++ b/src/databricks/labs/ucx/source_code/jobs.py
@@ -138,10 +138,10 @@ class WorkflowTaskContainer(SourceContainer):
                 for requirement in contents.splitlines():
                     clean_requirement = requirement.replace(" ", "")  # requirements.txt may contain spaces
                     if clean_requirement.startswith("-r"):
-                        logger.warning(f"References to other requirements file is not supported: {requirement}")
+                        logger.warning(f"References to other requirements files are not supported: {requirement}")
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/source_code/jobs.py
```diff
@@ -138,10 +138,10 @@ def _register_library(self, graph: DependencyGraph, library: compute.Library) ->
                 for requirement in contents.splitlines():
                     clean_requirement = requirement.replace(" ", "")  # requirements.txt may contain spaces
                     if clean_requirement.startswith("-r"):
-                        logger.warning(f"References to other requirements file is not supported: {requirement}")
+                        logger.warning(f"References to other requirements files are not supported: {requirement}")
```
JCZuurmond, line 141: It is singular, right?

asnare: The subject is _References_, plural. Either of the following is correct:

 - _Reference_ to other requirements file is not supported: {requirement}
 - References to other requirements files _are_ not supported: {requirement}

The meaning is subtly different (general vs specific) but either would be appropriate.

JCZuurmond: > Reference to other requirements file is not supported: {requirement}

Keep that one!

nfx (approved):
lgtm

PR #2210, opened 2024-07-18
Improve error messages in case of connection errors

### LINKED ISSUE #1323: [FEATURE]: Support for HTTP Proxy Configuration in Air-gapped Databricks Workspaces

### Problem statement

After UCX installation, runing the assessment workflow in an air-gapped Databricks workspace is not possible.

Some tasks in the assessment Workflow require downloading libraries from the internet, example "setup_tacl" Task.
 Our workspace cannot access the internet without going through the HTTP proxy, so these tasks are failing.

### Proposed Solution

- There needs to be a way to pass the proxy configuration as a parameter to the tasks in the workflow that require installing libraries from the internet. 

- On possible solution, is to use init scripts. Example of the init script, applied to main and tacl clusters:

```
export HTTPS_PROXY=http://myproxy:8080
export HTTP_PROXY=http://myproxy:8080
export https_proxy=http://myproxy:8080
export http_proxy=http://myproxy:8080

pip install databricks-labs-blueprint
pip install databricks-sdk==0.24.0
pip install pyyaml
pip install databricks-labs-lsql==0.3.0
```

 Could you please propose a new version that supports execution of workflows in an air-gapped Workspaces?

### Additional Context

Related to:
- https://github.com/databrickslabs/ucx/issues/573

JCZuurmond:
## Changes
Improve error messages in case of connection errors

### Linked issues
Partially resolves #1323

### Functionality

- [x] modified existing command: `databricks labs ucx (un)install`

### Tests

- [x] manually tested
- [x] added integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -1350,7 +1350,7 @@ from a VPC, or from a specific IP range.
 
 **Solution:** Please check that your laptop has network connectivity to
 the Databricks account and workspace. If not, you might need to be
-connected to a VPN to access your workspace.
+connected to a VPN or configure an HTTP proxy to access your workspace.
 
 **From local machine to GitHub:** UCX needs internet access to connect
 to [<u>github.com</u>](https://github.com) (to download the tool) from
--- a/src/databricks/labs/ucx/install.py
+++ b/src/databricks/labs/ucx/install.py
@@ -293,10 +293,9 @@ class WorkspaceInstaller(WorkspaceContext):
         try:
             config = self.installation.load(WorkspaceConfig)
             self._compare_remote_local_versions()
-            if self._confirm_force_install():
-                return self._configure_new_installation(default_config)
-            self._apply_upgrades()
-            return config
+            if not self._confirm_force_install():
+                self._apply_upgrades()
+                return config
         except NotFound as err:
             logger.debug(f"Cannot find previous installation: {err}")
         except (PermissionDenied, SerdeError, ValueError, AttributeError):
--- a/src/databricks/labs/ucx/install.py
+++ b/src/databricks/labs/ucx/install.py
@@ -298,7 +298,14 @@ class WorkspaceInstaller(WorkspaceContext):
             # eventually raising a TimeoutError if tries take too long
             if isinstance(err.__cause__, ConnectionError):  # raise TimeoutError(...) from ConnectionError(...)
                 logger.warning(f"Cannot connect with {self.workspace_client.config.host}: {err}")
-        return self._configure_new_installation(default_config)
+                if default_config is None:
+                    raise err
+        try:
+            return self._configure_new_installation(default_config)
+        except TimeoutError as err:
+            if isinstance(err.__cause__, ConnectionError):
+                logger.warning(f"Cannot connect with {self.workspace_client.config.host}: {err}")
+            raise err
 
     def replace_config(self, **changes: Any) -> WorkspaceConfig | None:
         """
--- a/src/databricks/labs/ucx/install.py
+++ b/src/databricks/labs/ucx/install.py
@@ -11,6 +11,8 @@ from functools import cached_property
 from pathlib import Path
 from typing import Any
 
+from requests.exceptions import ConnectionError as RequestsConnectionError
+
 import databricks.sdk.errors
 from databricks.labs.blueprint.entrypoint import get_logger, is_in_debug
 from databricks.labs.blueprint.installation import Installation, SerdeError
@@ -46,7 +48,6 @@ from databricks.sdk.service.sql import (
     EndpointInfoWarehouseType,
     SpotInstancePolicy,
 )
-from requests.exceptions import ConnectionError as RequestsConnectionError
 
 from databricks.labs.ucx.__about__ import __version__
 from databricks.labs.ucx.assessment.azure import AzureServicePrincipalInfo
--- a/src/databricks/labs/ucx/install.py
+++ b/src/databricks/labs/ucx/install.py
@@ -291,6 +291,11 @@ class WorkspaceInstaller(WorkspaceContext):
         return False
 
     def configure(self, default_config: WorkspaceConfig | None = None) -> WorkspaceConfig:
+        """Configure the workspaces
+
+        Notes:
+        1. Connection errors are not handled within this configure method.
+        """
         try:
             config = self.installation.load(WorkspaceConfig)
             self._compare_remote_local_versions()
--- a/src/databricks/labs/ucx/install.py
+++ b/src/databricks/labs/ucx/install.py
@@ -283,6 +283,10 @@ class WorkspaceInstaller(WorkspaceContext):
         return False
 
     def configure(self, default_config: WorkspaceConfig | None = None) -> WorkspaceConfig:
+        connection_error_message = (
+            f"Cannot connect with {self.workspace_client.config.host} see "
+            "https://github.com/databrickslabs/ucx#network-connectivity-issues for help: %s"
+        )
         try:
             config = self.installation.load(WorkspaceConfig)
             self._compare_remote_local_versions()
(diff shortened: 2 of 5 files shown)


src/databricks/labs/ucx/install.py
```diff
@@ -285,14 +286,18 @@ def configure(self, default_config: WorkspaceConfig | None = None) -> WorkspaceC
...
+                return config
         except NotFound as err:
             logger.debug(f"Cannot find previous installation: {err}")
         except (PermissionDenied, SerdeError, ValueError, AttributeError):
             logger.warning(f"Existing installation at {self.installation.install_folder()} is corrupted. Skipping...")
+        except TimeoutError as err:
+            # API calls are wrapped with a retry mechanism for ConnectionError
+            # eventually raising a TimeoutError if tries take too long
+            if isinstance(err.__cause__, ConnectionError):  # raise TimeoutError(...) from ConnectionError(...)
+                logger.warning(f"Cannot connect with {self.workspace_client.config.host}: {err}")
```
JCZuurmond, line 300: @nfx : What message would you like to have here? We can include something about a proxy, but that does not always have to be the reason the connection error is raised

nfx: @JCZuurmond the best thing would a link to https://github.com/databrickslabs/ucx#network-connectivity-issues and expand documentation over there a bit.

nfx (approved):

README.md
```diff
@@ -1348,7 +1348,7 @@ from a VPC, or from a specific IP range.
 
 **Solution:** Please check that your laptop has network connectivity to
 the Databricks account and workspace. If not, you might need to be
-connected to a VPN to access your workspace.
+connected to a VPN or configure an HTTP proxy to access your workspace.
```
asnare, line 1351:
```suggestion
connected to a VPN or configure a HTTP proxy to access your workspace.
```

src/databricks/labs/ucx/install.py
```diff
@@ -282,18 +284,31 @@ def _confirm_force_install(self) -> bool:
...
+            # API calls are wrapped with a retry mechanism for requests.exceptions.ConnectionError
+            # eventually raising a TimeoutError if tries take too long
+            if isinstance(err.__cause__, RequestsConnectionError):  # raise TimeoutError(...) from ConnectionError(...)
+                logger.warning(connection_error_message, err)
+        try:
+            return self._configure_new_installation(default_config)
+        except TimeoutError as err:
+            if isinstance(err.__cause__, RequestsConnectionError):
+                logger.warning(connection_error_message, err)
+            raise err
```
asnare, line 311: I think the `TimeoutError` needs to be at a higher level than the existing catch blocks: if we hit the `TimeoutError` on the first call, I don't think it makes sense to continue the second call?

If we do want to continue I'd prefer (slightly) different messages:
 - The first message should indicate we're skipping.
 - Even though we normally have backtraces in the logs, it helps locate the right part of the code quickly when the messages are different. (In this case locating the installation versus configuring the new installation.

src/databricks/labs/ucx/install.py
```diff
@@ -282,18 +284,31 @@ def _confirm_force_install(self) -> bool:
...
         except NotFound as err:
             logger.debug(f"Cannot find previous installation: {err}")
         except (PermissionDenied, SerdeError, ValueError, AttributeError):
             logger.warning(f"Existing installation at {self.installation.install_folder()} is corrupted. Skipping...")
-        return self._configure_new_installation(default_config)
+        except TimeoutError as err:
+            # API calls are wrapped with a retry mechanism for requests.exceptions.ConnectionError
+            # eventually raising a TimeoutError if tries take too long
+            if isinstance(err.__cause__, RequestsConnectionError):  # raise TimeoutError(...) from ConnectionError(...)
+                logger.warning(connection_error_message, err)
```
asnare, line 305: What is supposed to happen if a `TimeoutError` reaches here that is _not_ caused by `RequestsConnectionError`? At the moment we simply swallow it without any indication whereas previously it was propagated.

tests/integration/install/test_installation.py
```diff
@@ -409,3 +412,48 @@ def test_installation_with_dependency_upload(ws, installation_ctx, mocker):
...
+
+
+@pytest.fixture
+def config_without_credentials() -> Config:
+    """Mock no internet access by raising a ConnectionError"""
+
+    @credentials_strategy("raise_connection_error", [])
+    def raise_connection_error(_: Any):
+        def inner():
+            raise RequestsConnectionError("no internet")
```
asnare, line 424:
```suggestion
            raise RequestsConnectionError("simulated no internet")
```

src/databricks/labs/ucx/install.py
```diff
@@ -282,18 +284,31 @@ def _confirm_force_install(self) -> bool:
         return False
 
     def configure(self, default_config: WorkspaceConfig | None = None) -> WorkspaceConfig:
+        connection_error_message = (
+            f"Cannot connect with {self.workspace_client.config.host} see "
```
nfx, line 288:
```suggestion
            f"Cannot connect with {self.workspace_client.config.host}, see "
```

src/databricks/labs/ucx/install.py
```diff
@@ -282,18 +284,31 @@ def _confirm_force_install(self) -> bool:
...
         try:
             config = self.installation.load(WorkspaceConfig)
             self._compare_remote_local_versions()
-            if self._confirm_force_install():
-                return self._configure_new_installation(default_config)
-            self._apply_upgrades()
-            return config
+            if not self._confirm_force_install():
+                self._apply_upgrades()
+                return config
```
nfx, line 296:
```suggestion
            if not self._confirm_force_install():
                self._apply_upgrades()
            return config
```
otherwise a bug

tests/integration/conftest.py
```diff
@@ -755,7 +755,10 @@ def installation_ctx(  # pylint: disable=too-many-arguments
         ws,
     )
     yield ctx.replace(workspace_client=ws, sql_backend=sql_backend)
-    ctx.workspace_installation.uninstall()
+    try:
+        ctx.workspace_installation.uninstall()
+    except TimeoutError:
+        pass  # Best effort uninstall, we do not want the teardown to raise an exception
```
nfx, line 761: well... log a warning...

nfx (changes requested):

README.md
JCZuurmond: https://www.pristineword.com/grammar-an-http/

src/databricks/labs/ucx/install.py
JCZuurmond: > if we hit the TimeoutError on the first call, I don't think it makes sense to continue the second call?

If there is a `default_config` then we want to try

> I think the TimeoutError needs to be at a higher level than the existing catch blocks:

I thought so too at first, but it resulted in introducing it in a context where it not always made sense. 

It is not perfect, but my reasoning was that this is (hopefully) an exceptional flow. 99%+ of the cases a connection with Databricks can be made. Hence, my tradeoff of keeping code complexity lower vs that the backtraces are not right were the first try fails

src/databricks/labs/ucx/install.py
JCZuurmond: We retry with default config, see other comment about code complexity. If it fails there again, we raise it

JCZuurmond: Also, this was only the case if `_confirm_force_install` is `False` otherwise the flow remains the same. That those two code flows are different is also confusing

src/databricks/labs/ucx/install.py
JCZuurmond: Thanks, it was not caught by my local test runs for some reason

tests/integration/install/test_installation.py
```diff
@@ -409,3 +415,50 @@ def test_installation_with_dependency_upload(ws, installation_ctx, mocker):
...
+        """Mock no internet access by raising a ConnectionError"""
+
+        def inner():
+            raise RequestsConnectionError("no internet")
+
+        return inner
+
+    config = Config(credentials_strategy=raise_connection_error, retry_timeout_seconds=1)
+    ws = WorkspaceClient(host=env_or_skip("DATABRICKS_HOST"), config=config)
+    ctx = MockInstallationContext(
```
nfx, line 438: this is a unit test (uses mocks), not an integration test (uses fixtures). move.

nfx (changes requested):

tests/integration/install/test_installation.py
JCZuurmond: I agree it make sense now to move this to the unit tests as it (now) works without internet connection. Note that all integration tests using the `installation_ctx` fixture are using the `MockInstallationContext`

JCZuurmond:
Decided to move the connection error handler into the `WorkspaceInstaller.run` instead of `WorkspaceInstaller.configure`. It does mean that connection error should still be handled when calling the configure, however, currently this method is only called during testing. Moreover, it simplifies the error handling, as only one `try-except` is required now

nfx (approved):
lgtm

PR #2214, opened 2024-07-19
Disable the builtin pip version check when running pip commands

asnare:
## Changes

This PR disables Pip's built-in version check when using it to install dependencies. Pip itself is outside the control of UCX, and the version check often produces a lot of noise in the pip-related errors and finding messages.

### Functionality

- modified existing workflow: `...`

### Tests

- manually tested
- unit tests


JCZuurmond (approved):
lgtm

PR #1920, opened 2024-06-19
Use lsql Lakeview dashboard-as-code implementation to create dashboards

JCZuurmond:
## Changes
Use lsql Lakeview dashboard-as-code implementation to create dashboards

### Linked issues
Resolves #1478 
Resolves #1535
Resolves #1536
Resolves #1740

### Functionality 

- [x] modified existing workflow: that have dashboards

### Tests

- [x] manually tested
- [x] used existing unit tests: `test_dashboard`
- [x] using exiting integration tests: `test_running_real_migrate_groups_job`
- [x] verified on staging environment (screenshot attached)

PR #2224, opened 2024-07-22
Fix create database in workspace installation

JCZuurmond:
## Changes
Database creation should be done asynchronously. Mypy probably missed this as the method did not have any type hints. Introduced in #1920

### DIFF
--- a/src/databricks/labs/ucx/install.py
+++ b/src/databricks/labs/ucx/install.py
@@ -485,7 +485,7 @@ class WorkspaceInstallation(InstallationMixin):
         """
         logger.info(f"Installing UCX v{self._product_info.version()}")
         install_tasks = [self._create_database]  # Need the database before creating the dashboards
-        install_tasks.extend(self._create_dashboards)
+        install_tasks.extend(self._get_create_dashboard_tasks())
         Threads.strict("installing components", install_tasks)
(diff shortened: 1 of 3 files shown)


JCZuurmond:
`databricks.sdk.errors.platform.InternalError: A database error occurred during import-dashboard-new` pops up again; waiting on how to resolve this. 

Options:
1. Do not create the database in parallel to the dashboard creation
2. Increase the `retried` on `InternalError` for the dashboard creation and hope that it is sufficient

src/databricks/labs/ucx/install.py
```diff
@@ -476,9 +476,15 @@ def config(self):
...
+        """Run workflow installation.
+
+        Returns
+            bool :
+                True, installation finished. False installation did not finish.
+        """
         logger.info(f"Installing UCX v{self._product_info.version()}")
-        install_tasks = [self._create_database()]  # Need the database before creating the dashboards
+        install_tasks = [self._create_database]  # Need the database before creating the dashboards
         install_tasks.extend(self._create_dashboards())
```
nfx, line 488:
```suggestion
        install_tasks.extend(self._create_dashboards)
```
why not the same for dashboards then?

nfx (approved):
lgtm

PR #2225, opened 2024-07-22
Fix mypy flags `import distutils` in Python 3.12

JCZuurmond:
## Changes
Fix mypy flags `import distutils` in python 3.12. Local `make lint` fails with Python 3.12.


nfx (approved):
lgtm

Issue #2156, opened 2024-07-11
[BUG]: Child files imported via %run magic line are not assessed


ericvergnaud:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Child files invoked from %run magic line are not assessed

### Expected Behavior

Child files imported from %run magic line should be assessed by being added to the dependency graph

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

ericvergnaud:
Databricks only support importing child files if they have an extension and the `%run` command is not the first line in the cell.

ericvergnaud:
This needs to be implemented in relation to #2155 i.e.:
 - the same magic line has a different behavior depending on the context, if the magic line is not for a notebook, then a file dependency must be registered
 - the assessment also needs to be done in context

Issue #2155, opened 2024-07-10
[BUG]: Child notebooks imported from %run magic line are not assessed


ericvergnaud:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Child notebooks imported from %run magic line are not assessed

### Expected Behavior

Child notebooks imported from %run magic line should be assessed by being added to the dependency graph

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

ericvergnaud:
Databricks supports importing notebooks only if they do not have an extension and the %run magic line is the only line. If this is not the first cell, we already support this behavior via `RunCell`

ericvergnaud:
The notebooks are now assessed, but that assessment is not performed in context i.e. the notebook is not aware of global variables set by the parent notebook, which may lead to avoidable Advices

PR #2226, opened 2024-07-22
Lint child dependencies recursively

ericvergnaud:
## Changes
The current code lints local files without considering parent/child relationships (where a given 'parent' file imports or runs a 'child' file).
This makes it impossible to lint files in context, where a child file uses variables inherited from a parent file.
This PR is progress towards that, by linting child dependencies recursively.

### Linked issues
Progresses #2155 
Progresses #2156

### Functionality
- [x] modified existing command: `databricks labs ucx lint-local-code`

### Tests
- [x] checked integration tests

### DIFF
--- a/src/databricks/labs/ucx/source_code/graph.py
+++ b/src/databricks/labs/ucx/source_code/graph.py
@@ -111,6 +111,10 @@ class DependencyGraph:
         return self if self._parent is None else self._parent.root
 
     @property
+    def parent(self):
+        return self._parent
+
+    @property
     def all_dependencies(self) -> set[Dependency]:
         dependencies: set[Dependency] = set()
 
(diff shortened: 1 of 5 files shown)


src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -123,6 +127,28 @@ def add_to_dependencies(graph: DependencyGraph) -> bool:
         self.visit(add_to_dependencies, set())
         return dependencies
 
+    @property
+    def root_dependencies(self) -> set[Dependency]:
```
nfx, line 131:
```suggestion
    def roots(self) -> set[Dependency]:
```
nit

nfx (approved):
lgtm

PR #2228, opened 2024-07-23
Parallelize the install deployment differently

JCZuurmond:
Stacked on top of #2224

## Changes
Parallelize the install deployment as follows:

image

### Functionality

- [x] modified existing command: `databricks labs install ucx`

### Tests

- [ ] manually tested


JCZuurmond:
Something is off with the git history, will recreate

PR #2229, opened 2024-07-23
Update installation parallelization

JCZuurmond:
Stacked on top of #2224

## Changes
Parallelize the install deployment as follows:

image

### Functionality

- [x] modified existing command: `databricks labs install ucx`

### Tests

- [x] manually tested

### DIFF
--- a/tests/unit/test_install.py
+++ b/tests/unit/test_install.py
@@ -228,7 +228,7 @@ def test_create_database(ws, caplog, mock_installation, any_prompt):
             raise e.errs[0]
 
     assert "Kindly uninstall and reinstall UCX" in str(failure.value)
-    wheels.upload_to_wsfs.assert_not_called()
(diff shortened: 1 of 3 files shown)


tests/unit/test_install.py
```diff
@@ -228,7 +228,7 @@ def test_create_database(ws, caplog, mock_installation, any_prompt):
             raise e.errs[0]
 
     assert "Kindly uninstall and reinstall UCX" in str(failure.value)
-    wheels.upload_to_wsfs.assert_not_called()
+    wheels.upload_to_wsfs.assert_called_once()
```
JCZuurmond, line 231: `wheels.upload_to_wsfs` is now called as job creation calls this method and is reached when deploying jobs in parallel to the database creation

nfx (approved):
lgtm

PR #2240, opened 2024-07-24
Group manager optimisation: during group enumeration only request the attributes that are needed.

asnare:
## Changes

Group enumeration is expensive: we already have a special code-path when members are requested while enumerating groups due to API issues. This PR improves that code-path such that during enumeration only the bare minimum set of attributes is requested.

### Functionality

- modified existing workflow: `group-migration`


asnare:
Failed acceptance test is exhibiting the same problem as described in #2227.

nfx (approved):
lgtm

PR #2236, opened 2024-07-24
Lint dependencies in context

ericvergnaud:
## Changes
Introduces `InheritedContext` class which gathers code fragments from file/notebook parents, and uses it when linting child file/notebook

### Linked issues
Resolves #2155 
Resolves #2156 
Resolves #2221 

### Functionality
None

### Tests
- [x] added unit tests
- [x] added functional tests
- [x] added integration tests

PR #2254, opened 2024-07-26
fix computation of root dependencies and dependency routes

ericvergnaud:
## Changes
fix computation of root dependencies by only returning files/notebooks

### Linked issues
None

### Functionality
None

### Tests
- [x] added unit tests


src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -134,19 +134,31 @@ def root_dependencies(self) -> set[Dependency]:
 
         def add_to_dependencies(graph: DependencyGraph) -> bool:
             dependency = graph.dependency
+            # if already encountered we're done
             if dependency in children:
-                return False  # Nothing to do
+                return False
+            # if it's not a real file, then it's not a root
+            if not dependency.path.is_file():
```
nfx, line 141: if `WorkspacePath` is used, then this will skip notebooks in the workspace, that are valid roots. can be do `not dependency.path.is_dir()` or something like it?

```
    def is_file(self) -> bool:
        """Return True if the path points to a file in Databricks Workspace."""
        try:
            return self._object_info.object_type == ObjectType.FILE
        except DatabricksError:
            return False
```

nfx: which cyclic dependencies?

ericvergnaud: see https://github.com/databrickslabs/ucx/pull/2236/files/0950ab1f809795fc7a086b1d4b04936ec83b9e26#r1697132861
https://github.com/databrickslabs/ucx/pull/2236/files#r1697135314
and
https://github.com/databrickslabs/ucx/pull/2236/files#r1697136495

ericvergnaud:
Already merged via #2236, closing

PR #2247, opened 2024-07-25
Group migration: improve robustness while deleting workspace groups

asnare:
## Changes

This PR updates the group manager so that deleting workspace groups is more reliable. Changes include:

 - We no longer skip deletion of groups that don't appear to be present. Due to eventual consistency issues this was occurring with groups that had recently been renamed (to their temporary name).
 - Deletion now waits for the effects of deletion to be visible by double-checking that a group can no longer be directly retrieved from the API, and that it no longer appears in the list of groups during enumeration. (Due to API limitations this is not a guarantee that the groups are no longer visible, but it does decrease the likelihood of anything noticed afterwards.)
 - Improved logging.

A subsequent PR will update group renaming to use a similar approach to that here.

### Linked issues

Resolves #2227.

### Functionality

- modified existing workflow: `remove-workspace-local-backup-groups`

### Tests

- updated unit tests
- updated integration tests

PR #2263, opened 2024-07-29
Group migration: improve robustness when renaming groups

asnare:
## Changes

This PR updates the way we check for the completion of group renames; similar to deletion we now double-check before assuming the rename has taken effect.

### Linked issues

Follows on from #2247.

### Functionality

  - `migrate-groups`
  - `migrate-groups-experimental`

### Tests

- updated unit tests
- existing integration tests


nfx (approved):
lgtm

PR #2265, opened 2024-07-29
Update purge suffix so that resources are preserved for _at least_ as long as the configured timeout.

asnare:
## Changes

This PR updates the purge suffix used by integration tests so that resources are marked for preservation _at least_ as long as the configured timeout. The situation is that:

 - Many resources can be marked with a tag (or a special suffix in the name) to ensure they are preserved until a specific time.
 - The timestamp refers to a UTC-based hour, such as: `2024072915`. This means that resources can be deleted at any moment after 3 p.m. (UTC).
 - Prior to this change, resources created as late as 2:44:59 p.m. would be tagged for deletion with `2024072915` meaning deletion as of 3 p.m. This is 0:15 later instead of the intended 1:15.

The default timeout for tests was 1:15 in an attempt to deal with resources being deleted when the test started near the hour boundary. This PR also updates the default timeout to 1:00 because the extra 0:15 isn't needed to ensure this.

### DIFF
--- a/src/databricks/labs/ucx/installer/workflows.py
+++ b/src/databricks/labs/ucx/installer/workflows.py
@@ -51,10 +51,11 @@ from databricks.labs.ucx.framework.tasks import Task
 from databricks.labs.ucx.installer.logs import PartialLogRecord, parse_logs
 from databricks.labs.ucx.installer.mixins import InstallationMixin
 
+from ..mixins.fixtures import get_test_purge_time
+
 logger = logging.getLogger(__name__)
 
-TEST_JOBS_PURGE_TIMEOUT = timedelta(hours=1, minutes=15)
-TEST_NIGHTLY_CI_JOBS_PURGE_TIMEOUT = timedelta(hours=3)  # Buffer for debugging nightly integration test runs
+TEST_NIGHTLY_CI_RESOURCES_PURGE_TIMEOUT = timedelta(hours=3)  # Buffer for debugging nightly integration test runs
 EXTRA_TASK_PARAMS = {
     "job_id": "{{job_id}}",
     "run_id": "{{run_id}}",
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -51,10 +51,11 @@
 from databricks.labs.ucx.installer.logs import PartialLogRecord, parse_logs
 from databricks.labs.ucx.installer.mixins import InstallationMixin
 
+from ..mixins.fixtures import get_test_purge_time
```
nfx, line 54: this includes `pytest` into the runtime path of UCX installation, which is going to break customers. copy the code. otherwise adjust the `new_installation` fixture, which is a more proper way of doing this.

nfx (changes requested):

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -463,10 +462,15 @@ def _is_nightly():
...
-            return get_test_purge_time(TEST_NIGHTLY_CI_RESOURCES_PURGE_TIMEOUT)
-        return get_test_purge_time()
+    @classmethod
+    def _get_test_purge_time(cls) -> str:
+        # Duplicate of mixins.fixtures.get_test_purge_time(); we don't want to import pytest as a transitive dependency.
+        timeout = TEST_NIGHTLY_CI_RESOURCES_PURGE_TIMEOUT if cls._is_nightly() else TEST_RESOURCE_PURGE_TIMEOUT
+        now = datetime.now(timezone.utc)
+        purge_deadline = now + timeout
+        # Round UP to the next hour boundary: that is when resources will be deleted.
+        purge_hour = purge_deadline + (datetime.min.replace(tzinfo=timezone.utc) - purge_deadline) % timedelta(hours=1)
```
JCZuurmond, line 472: I find the following logic complicated to round up to an hour: `(datetime.min.replace(tzinfo=timezone.utc) - purge_deadline) % timedelta(hours=1)`

JCZuurmond (approved):
LGTM

PR #2274, opened 2024-07-30
Integration tests: purge suffix for more group fixtures

asnare:
## Changes

This PR updates the `make_ucx_group()` fixture that creates a pair of workspace & account groups to include a purge suffix in the group name(s) if none is supplied. This ensures that the groups are not purged while the integration test is running.


nfx (approved):
lgtm

PR #2280, opened 2024-07-30
Make ucx pylsp plugin configurable

vsevolodstep-db:
## Changes
Make LSP linter plugin configurable with cluster information. This config can be provided either in a file or by a client and its provisioning is handled by pylsp infrastructure.
Spark Connect linter is now applied only to UC Shared clusters, as Single-User clusters are running in Spark Classic mode.

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/migration_status.py
+++ b/src/databricks/labs/ucx/hive_metastore/migration_status.py
@@ -26,6 +26,17 @@ class MigrationStatus:
     def destination(self):
         return f"{self.dst_catalog}.{self.dst_schema}.{self.dst_table}".lower()
 
+    @classmethod
+    def from_json(cls, json: dict):
+        return cls(
+            src_schema=json['src_schema'],
+            src_table=json['src_table'],
+            dst_catalog=json.get('dst_catalog', None),
+            dst_schema=json.get('dst_schema', None),
+            dst_table=json.get('dst_table', None),
+            update_ts=json.get('update_ts', None),
+        )
+
 
 @dataclass(frozen=True)
 class TableView:
--- a/src/databricks/labs/ucx/hive_metastore/migration_status.py
+++ b/src/databricks/labs/ucx/hive_metastore/migration_status.py
@@ -27,14 +27,14 @@ class MigrationStatus:
         return f"{self.dst_catalog}.{self.dst_schema}.{self.dst_table}".lower()
 
     @classmethod
-    def from_json(cls, json: dict):
+    def from_json(cls, raw: dict[str, str]) -> "MigrationStatus":
         return cls(
-            src_schema=json['src_schema'],
-            src_table=json['src_table'],
-            dst_catalog=json.get('dst_catalog', None),
-            dst_schema=json.get('dst_schema', None),
-            dst_table=json.get('dst_table', None),
-            update_ts=json.get('update_ts', None),
+            src_schema=raw['src_schema'],
+            src_table=raw['src_table'],
+            dst_catalog=raw.get('dst_catalog', None),
+            dst_schema=raw.get('dst_schema', None),
+            dst_table=raw.get('dst_table', None),
+            update_ts=raw.get('update_ts', None),
         )
 
 
--- a/src/databricks/labs/ucx/source_code/base.py
+++ b/src/databricks/labs/ucx/source_code/base.py
@@ -183,13 +183,13 @@ class CurrentSessionState:
             catalog=json.get('catalog', DEFAULT_CATALOG),
             spark_conf=json.get('spark_conf', None),
             named_parameters=json.get('named_parameters', None),
-            data_security_mode=CurrentSessionState._parse_security_mode(json.get('data_security_mode', None)),
+            data_security_mode=cls.parse_security_mode(json.get('data_security_mode', None)),
             is_serverless=json.get('is_serverless', False),
             dbr_version=tuple(json['dbr_version']) if 'dbr_version' in json else None,
         )
 
     @staticmethod
-    def _parse_security_mode(mode_str: str | None) -> compute.DataSecurityMode | None:
+    def parse_security_mode(mode_str: str | None) -> compute.DataSecurityMode | None:
         try:
             return compute.DataSecurityMode(mode_str) if mode_str else None
         except ValueError:
--- a/src/databricks/labs/ucx/source_code/base.py
+++ b/src/databricks/labs/ucx/source_code/base.py
@@ -178,21 +178,23 @@ class CurrentSessionState:
 
     @classmethod
     def from_json(cls, json: dict) -> CurrentSessionState:
-        mode_str = json.get('data_security_mode', None)
-        try:
-            data_security_mode = compute.DataSecurityMode(mode_str) if mode_str else None
-        except ValueError:
-            logger.error(f'Unknown data_security_mode {mode_str}')
-            data_security_mode = None
         return cls(
             schema=json.get('schema', DEFAULT_SCHEMA),
             catalog=json.get('catalog', DEFAULT_CATALOG),
             spark_conf=json.get('spark_conf', None),
             named_parameters=json.get('named_parameters', None),
-            data_security_mode=data_security_mode,
+            data_security_mode=CurrentSessionState._parse_security_mode(json.get('data_security_mode', None)),
             is_serverless=json.get('is_serverless', False),
             dbr_version=tuple(json['dbr_version']) if 'dbr_version' in json else None,
         )
+
+    @staticmethod
+    def _parse_security_mode(mode_str: str | None) -> compute.DataSecurityMode | None:
+        try:
+            return compute.DataSecurityMode(mode_str) if mode_str else None
+        except ValueError:
+            logger.warning(f'Unknown data_security_mode {mode_str}')
+            return None
 
 
 class SequentialLinter(Linter):
--- a/src/databricks/labs/ucx/source_code/base.py
+++ b/src/databricks/labs/ucx/source_code/base.py
@@ -179,12 +179,17 @@ class CurrentSessionState:
     @classmethod
     def from_json(cls, json: dict) -> CurrentSessionState:
         mode_str = json.get('data_security_mode', None)
+        try:
+            data_security_mode = compute.DataSecurityMode(mode_str) if mode_str else None
+        except ValueError:
+            logger.error(f'Unknown data_security_mode {mode_str}')
+            data_security_mode = None
         return cls(
             schema=json.get('schema', DEFAULT_SCHEMA),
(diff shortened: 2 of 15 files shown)


src/databricks/labs/ucx/source_code/lsp_plugin.py
```diff
@@ -1,16 +1,50 @@
...
 from databricks.labs.ucx.source_code.lsp import Diagnostic
+from databricks.sdk.service.compute import DataSecurityMode
+
+
+logger = logging.getLogger(__name__)
 
 
 @hookimpl
-def pylsp_lint(workspace: Workspace, document: Document) -> list[dict]:  # pylint: disable=unused-argument
-    # TODO: initialize migration index and session state from config / env variables
```
JCZuurmond, line 11: What about the initialization of the migration index?

src/databricks/labs/ucx/source_code/lsp_plugin.py
```diff
@@ -1,16 +1,50 @@
...
+        is_serverless=bool(cfg.get('isServerless', False)),
+    )
+    languages = LinterContext(index=None, session_state=session_state)
     analyser = languages.linter(Language.PYTHON)
     code = document.source
     diagnostics = [Diagnostic.from_advice(_) for _ in analyser.lint(code)]
     return [d.as_dict() for d in diagnostics]
+
+
+def parse_dbr_version(version: str | None) -> tuple[int, int] | None:
```
JCZuurmond, line 33: Python has core utils for this: https://packaging.pypa.io/en/stable/version.html#usage

src/databricks/labs/ucx/source_code/lsp_plugin.py
```diff
@@ -1,16 +1,50 @@
...
+    if not version:
+        return None
+    version_parts = [int(i) for i in version.split('.') if i.isnumeric()]
+    if len(version_parts) < 2:
+        logger.error(f'Incorrect DBR version string: {version}')
+        return None
+    return version_parts[0], version_parts[1]
+
+
+def parse_data_security_mode(mode_str: str | None):
```
JCZuurmond, line 43:
```suggestion
def parse_data_security_mode(mode_str: str | None) -> DataSecurityMode | None:
```

src/databricks/labs/ucx/source_code/linters/spark_connect.py
```diff
@@ -238,6 +240,10 @@ def lint(self, node: NodeNG) -> Iterator[Advice]:
 
 class SparkConnectLinter(PythonLinter):
     def __init__(self, session_state: CurrentSessionState):
+        if session_state.data_security_mode != DataSecurityMode.USER_ISOLATION:
```
JCZuurmond, line 243: Why do we want to run this only for `USER_ISOLATION` security mode?

JCZuurmond (commented):
Have some questions

src/databricks/labs/ucx/source_code/base.py
```diff
@@ -178,12 +178,13 @@ class CurrentSessionState:
...
     @classmethod
     def from_json(cls, json: dict) -> CurrentSessionState:
+        mode_str = json.get('data_security_mode', None)
         return cls(
             schema=json.get('schema', DEFAULT_SCHEMA),
             catalog=json.get('catalog', DEFAULT_CATALOG),
             spark_conf=json.get('spark_conf', None),
             named_parameters=json.get('named_parameters', None),
-            data_security_mode=json.get('data_security_mode', None),
+            data_security_mode=compute.DataSecurityMode(mode_str) if mode_str else None,
```
nfx, line 187: this may crash when we introduce a new data security mode and old clients are not upgraded to the newer SDK. make it a bit more robust.

src/databricks/labs/ucx/source_code/lsp_plugin.py
```diff
@@ -1,16 +1,50 @@
...
+    if not mode_str:
+        return None
+    try:
+        return DataSecurityMode(mode_str)
+    except ValueError as _:
+        logger.error(f'Unknown data_security_mode: {mode_str}')
+        return None
```
nfx, line 50: shouldn't we return `DataSecurityMode.NONE` as a fallback?

nfx (changes requested):

src/databricks/labs/ucx/source_code/linters/spark_connect.py
vsevolodstep-db: UC Shared Clusters use Spark connect, this linter detects use of legacy APIs that are not supported  in Spark Connect mode

src/databricks/labs/ucx/source_code/lsp_plugin.py
vsevolodstep-db: I think that `None` suits better, as `DataSecurityMode.NONE` represents a no isolation cluster without UC, while `None` stands for a missing or a invalid access mode

src/databricks/labs/ucx/source_code/base.py
```diff
@@ -178,12 +178,18 @@ class CurrentSessionState:
 
     @classmethod
     def from_json(cls, json: dict) -> CurrentSessionState:
+        mode_str = json.get('data_security_mode', None)
+        try:
+            data_security_mode = compute.DataSecurityMode(mode_str) if mode_str else None
+        except ValueError:
+            logger.error(f'Unknown data_security_mode {mode_str}')
```
nfx, line 185:
```suggestion
            logger.warning(f'Unknown data_security_mode {mode_str}')
```
error logs fail the workflow

src/databricks/labs/ucx/source_code/base.py
```diff
@@ -178,12 +178,18 @@ class CurrentSessionState:
 
     @classmethod
     def from_json(cls, json: dict) -> CurrentSessionState:
+        mode_str = json.get('data_security_mode', None)
```
nfx, line 181: nit: it's better to extract this into a private method

src/databricks/labs/ucx/source_code/linters/spark_connect.py
nfx: `databricks labs ucx lint-local-code` command sets no data security mode, so these linters would be omitted. 
```suggestion
        if session_state.data_security_mode is not None and session_state.data_security_mode != DataSecurityMode.USER_ISOLATION:
```

nfx (changes requested):

src/databricks/labs/ucx/source_code/linters/spark_connect.py
vsevolodstep-db: What's the expectation for `databricks labs ucx lint-local-code`? These errors emitted by `SparkConnectLinter` are relevant for UC Shared Clusters, but they have no use for Single-User ones. While it's still better to migrate off using deprecated APIs, it's not possible to achieve for all workloads (e.g. using RDD APIs on a UC SIngle User cluster). I wonder if we need to pass `data_security_mode` to `lint-local-code` as an optional flag

src/databricks/labs/ucx/hive_metastore/migration_status.py
```diff
@@ -26,6 +26,17 @@ class MigrationStatus:
     def destination(self):
         return f"{self.dst_catalog}.{self.dst_schema}.{self.dst_table}".lower()
 
+    @classmethod
+    def from_json(cls, json: dict):
```
JCZuurmond, line 30:
```suggestion
    def from_json(cls, raw: dict[str, str]) -> "MigrationStatus":
```

src/databricks/labs/ucx/source_code/base.py
```diff
@@ -183,11 +183,19 @@ def from_json(cls, json: dict) -> CurrentSessionState:
             catalog=json.get('catalog', DEFAULT_CATALOG),
             spark_conf=json.get('spark_conf', None),
             named_parameters=json.get('named_parameters', None),
-            data_security_mode=json.get('data_security_mode', None),
+            data_security_mode=CurrentSessionState._parse_security_mode(json.get('data_security_mode', None)),
```
JCZuurmond, line 186: Now accessing the private method from outside the class, does the following work? What does mypy say

```suggestion
            data_security_mode=cls._parse_security_mode(json.get('data_security_mode', None)),
```

src/databricks/labs/ucx/source_code/lsp_plugin.py
```diff
@@ -1,16 +1,60 @@
...
-    languages = LinterContext(index=None, session_state=None)
+def pylsp_lint(config: Config, document: Document) -> list[dict]:
+    cfg = config.plugin_settings('pylsp_ucx', document_path=document.uri)
+
+    migration_index_list = cfg.get('migration_index', None)
+    migration_index = (
+        None
+        if not migration_index_list
+        else MigrationIndex([MigrationStatus.from_json(st) for st in migration_index_list])
+    )
```
JCZuurmond, line 28: Maybe initialise an empty index instead of `None`
```suggestion
    migration_index = MigrationIndex([MigrationStatus.from_json(st) for st in cfg.get('migration_index', [])])
```

tests/unit/source_code/test_lsp_plugin.py
```diff
@@ -48,3 +74,112 @@ def test_pylsp_lint(workspace):
...
+            'tags': [],
+        },
+    ]
+
+
+def test_pylsp_no_config(workspace, config):
+    code = 'sc.emptyRDD()'
+    _, doc = temp_document(code, workspace)
+    diagnostics = sorted(lsp_plugin.pylsp_lint(config, doc), key=lambda d: d['code'])
+    assert diagnostics == []
```
JCZuurmond, line 115: Indepedent of the code, should this always return no diagnostic if the config is not set?

JCZuurmond (commented):
To the best of my judgement, it looks good. Added some comments. @nfx : Could you give the final approval?

src/databricks/labs/ucx/source_code/lsp_plugin.py
```diff
@@ -1,16 +1,60 @@
...
+        return None
+    try:
+        release_version = version.parse(version_str).release
+        return release_version[0], release_version[1]
+    except version.InvalidVersion:
+        logger.warning(f'Incorrect DBR version string: {version_str}')
+        return None
+
+
+def parse_data_security_mode(mode_str: str | None) -> DataSecurityMode | None:
```
nfx, line 53: didn't you add `CurrentSessionState._parse_security_mode` method? just make it public and use here.

tests/unit/source_code/test_lsp_plugin.py
nfx: i think so

nfx (approved):
few last nits

PR #2281, opened 2024-07-30
Mock WheelsV2 in unit-tests to avoid lengthy wheel-building.

asnare:
## Changes

This PR updates some of the unit-testing fixtures to mock `WheelsV2` where this is referenced instead of the real thing; this avoids lengthy wheel-building during unit tests.

(Wheel-building can easily exceed the 30s unit-testing deadline, especially on developer laptops with the solacc dataset present.)

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -14,7 +14,7 @@ from databricks.labs.blueprint.entrypoint import is_in_debug
 from databricks.labs.blueprint.installation import Installation, MockInstallation
 from databricks.labs.blueprint.parallel import Threads
 from databricks.labs.blueprint.tui import MockPrompts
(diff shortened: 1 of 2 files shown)


JCZuurmond (changes requested):
I expect a `assert_` to be required on the mock

JCZuurmond (approved):
Mock is oke as it is not within the scope of the test

asnare:
> I expect a `assert_` to be required on the mock

I didn't realise yesterday that this feedback was intended to address an expecting linting failure: pylint has checks for unused mocks. However these checks are don't (seem to?) apply when the mock is being created as part of the fixture.

asnare:
AFAICT tell the failing acceptance tests are not related to changes in this PR.

PR #2332, opened 2024-07-31
Filter out missing import problems for imports within a try-except clause with ImportError

### LINKED ISSUE #1705: [BUG] Raising a library not found when linting with try-except ImportError
We raise a `library-not-found` when a library is imported within a try-except ImortError. We should extend the linter to cover the following scenario without raising a `library-not-found`

``` python
__all__ = ["__version__", "version_tuple"]

try:
    from ._version import version as __version__
    from ._version import version_tuple
except ImportError:  # pragma: no cover
    # broken installation, we don't even try
    # unknown only works because we do poor mans version compare
    __version__ = "unknown"
    version_tuple = (0, 0, "unknown")
```

_Originally posted by @JCZuurmond in https://github.com/databrickslabs/ucx/pull/1662#discussion_r1598510691_

ericvergnaud:
## Changes
Filter out missing import problems for imports within a try-except clause with ImportError 

### Linked issues
Resolves #1705 

### Functionality
None

### Tests
- [x] added unit tests

### DIFF
--- a/tests/unit/source_code/linters/test_python_imports.py
+++ b/tests/unit/source_code/linters/test_python_imports.py
@@ -1,12 +1,16 @@
 from __future__ import annotations
 
+from pathlib import Path
+
 import pytest
 
 from databricks.labs.ucx.source_code.base import CurrentSessionState
-from databricks.labs.ucx.source_code.graph import DependencyProblem
+from databricks.labs.ucx.source_code.graph import DependencyProblem, DependencyGraph, Dependency
+from databricks.labs.ucx.source_code.linters.files import FileLoader
 
 from databricks.labs.ucx.source_code.linters.imports import DbutilsLinter, ImportSource, SysPathChange
 from databricks.labs.ucx.source_code.linters.python_ast import Tree
+from databricks.labs.ucx.source_code.notebooks.cells import PythonCodeAnalyzer
 
 
 def test_linter_returns_empty_list_of_dbutils_notebook_run_calls():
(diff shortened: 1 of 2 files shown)


tests/unit/source_code/linters/test_python_imports.py
```diff
@@ -1,12 +1,16 @@
 from __future__ import annotations
 
+from pathlib import Path
+
 import pytest
 
 from databricks.labs.ucx.source_code.base import CurrentSessionState
-from databricks.labs.ucx.source_code.graph import DependencyProblem
+from databricks.labs.ucx.source_code.graph import DependencyProblem, DependencyGraph, Dependency
```
JCZuurmond, line 8:
```suggestion
from databricks.labs.ucx.source_code.graph import Dependency, DependencyGraph, DependencyProblem
```

tests/unit/source_code/linters/test_python_imports.py
```diff
@@ -27,27 +31,27 @@ def test_linter_returns_list_of_dbutils_notebook_run_calls():
 
 def test_linter_returns_empty_list_of_imports():
     tree = Tree.parse('')
-    assert not ImportSource.extract_from_tree(tree, DependencyProblem)[0]
+    assert not ImportSource.extract_from_tree(tree, DependencyProblem.from_node)[0]
```
JCZuurmond, line 34: Why is this required in this PR?

JCZuurmond (changes requested):
Some minor changes

tests/unit/source_code/linters/test_python_imports.py
ericvergnaud: It's not required, but as I was touching that file, I noticed the old code was incorrect (it didn't fail because the factory is not used)

PR #2341, opened 2024-08-01
Add `validate-table-locations` command for checking overlapping tables across workspaces

### LINKED ISSUE #673: Add `databricks labs ucx validate-table-locations` command
### Description
`databricks labs ucx validate-table-locations --workspace-profiles profile-one,profile-two,profile-three` to detect conflicts for table migration across the specified set of workspaces.

### Implementation
Create an exception list at the account level containing tables that show up on more than on workspace:
- Pointing to the same cloud storage location
- With different metadata
- With different ACLs

Addressing table conflicts/duplication require special processing we have the following options
- Define a "master" and create derivative objects as views
- Flag and skip the dupes
- Duplicate the data and create dupes
- Consider upgrading a workspace at a time. Highlight the conflict with prior upgrades.

Now for tables, there also needs to be a report on table/db inconsistency - like:

| Workspace | Table(s) |
|------------|----------|
| A                | db1.tbl1, db1.tbl3 |
| B                | db1.tbl2 |

And the team(s) that are driving UC Migration within account would make a decision after some time in review (of excel spreadsheet). By the way, we can split UCX installation across different Azure Subscriptions. And every installation would just focus on defining target catalog mapping per database. 

#### Questions
Situation
- Two workspaces, same dbs, all different tables and columns (all managed tables, effectively)
- Two workspaces, same dbs, 90% same tables, 10% are different tables
- Two workspaces, two different dbs

We can technically support both db_to_catalog and workspace_to_catalog, and even at the same time, but db_to_catalog will override workspace_to_catalog. We also need default_catalog_for_workspace, if workspace_to_catalog is set (default catalog for all workspaces is set per metastore)..

We can also do another override for tables, but we have unanswered questions:

- What if same db, same workspace, same table, but different columns/order/types? Do we ignore and keep in hive metastore? And then rerun the scan for tables and grants?
- What if during migration catalog/database/table were deleted either from hms and/or uc?

Speaking of metastores, in the beginning, there needs to be workspace_to_metastore mapping with default_metastore_for_workspace:
- Can we come up with a good default mapping here? 
   - Coarse or fine grained? 
   - Select between the two? 
   - Ask for inline input? 
- How many conflicts we expect to justify the need to create/support custom mapping?
   - In what what future-proof configuration format might we need for this mapping.

Thinking of an incremental approach here:
- Add workspace-id to the schema-name
   - eg: ucx-##########
   - This can be retrieved via the workspace URL; prob best added to the SDK
   - This ensures that each workspace inventory is isolated; regardless of if there is an external HMS or not
- No need to alter "destroy" scripts
- Add workspace-id as a column to every inventory table to treat tables as multi-tenant
   - eg: workspace_id
   - This ensures that data is always segmented; even if a user overrides the config for schema name
- Add metastore object to inventory
   - Not sure if there is a natural identifier or not
   - Capture if this metastore is internal, external, glue, config, etc.
- Add metastore-id column to any related inventory object
- Create Workflow to switch inventory from HMS to UC
- These tasks would ideally be included at the end of the workflow for Step 1 "Assign Metastore to Workspace"
 - Copy inventory data from HMS ucx-########## schema to ucx CATALOG in the same set of tables
- Since each table has workspace-id column and same schema, this is a simple `INSERT INTO SELECT * FROM`
- Run Destroy workflow to purge data from HMS
- Update config file to redirect ucx location from HMS to UC
- From this point forward, UCX in this workspace will reference the ucx CATALOG
- When running the "mapping" portion for any object, UCX logic will examine the data in the UC tables and highlight if there are any mappings which must be reviewed and approved manually
Each object will have a mapping table from old to new
Exporting / Importing to a google sheet would be a good interface for this
The same process can be run for each new workspace that gets added
Migration of objects is performed based on the mapping tables

### Related issues:
- https://github.com/databrickslabs/ucx/issues/335

**Upstream dependencies**:
- https://github.com/databrickslabs/ucx/issues/104
- https://github.com/databrickslabs/ucx/issues/669

**Downstream dependencies**:
- https://github.com/databrickslabs/ucx/issues/670

**There are 4 possible actions in case of failure**:
- https://github.com/databrickslabs/ucx/issues/671
- https://github.com/databrickslabs/ucx/issues/672
- https://github.com/databrickslabs/ucx/issues/675
- https://github.com/databrickslabs/ucx/issues/676

JCZuurmond:
## Changes
Add `validate` command for checking overlapping tables across workspaces. This command is intended to run before migrating tables to validate tables can be migrated across workspaces.

### Linked issues

Resolves #673

### Functionality

- [x] added relevant user documentation
- [x] added new CLI command: `validate-table-locations`

### Tests

- [x] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/labs.yml
+++ b/labs.yml
@@ -51,11 +51,9 @@ commands:
       - name: workspace-ids
         description: List of workspace IDs to create account groups from.
 
-  - name: validate
+  - name: validate-table-locations
     is_account_level: true
-    description: |
-      Validate migration readiness:
-      - Check if tables can be migrated across workspaces
+    description: Validate if the table locations are overlapping across workspaces
     flags:
       - name: workspace-ids
         description: |
--- a/labs.yml
+++ b/labs.yml
@@ -54,8 +54,8 @@ commands:
   - name: validate
     is_account_level: true
     description: |
-      Validate migration readiness across workspaces:
-      - Check if tables can be migrated
+      Validate migration readiness:
+      - Check if tables can be migrated across workspaces
     flags:
       - name: workspace-ids
         description: |
--- a/labs.yml
+++ b/labs.yml
@@ -53,7 +53,7 @@ commands:
 
   - name: validate-table-locations
     is_account_level: true
-    description: Validate if the table locations are overlapping across workspaces
+    description: Validate if the table locations are overlapping in a workspace and across workspaces.
     flags:
       - name: workspace-ids
         description: |
--- a/labs.yml
+++ b/labs.yml
@@ -51,6 +51,17 @@ commands:
       - name: workspace-ids
         description: List of workspace IDs to create account groups from.
 
+  - name: validate
+    is_account_level: true
+    description: |
+      Validate migration readiness across workspaces:
+      - Check if tables can be migrated
+    flags:
+      - name: workspace-ids
+        description: |
+          List of workspace IDs to include.
+          If --workspace-ids is not provided, it will use all workspaces present in the account.
+
   - name: manual-workspace-info
     description: only supposed to be run if cannot get admins to run `databricks labs ucx sync-workspace-info`
 
--- a/src/databricks/labs/ucx/account/aggregate.py
+++ b/src/databricks/labs/ucx/account/aggregate.py
@@ -128,9 +128,13 @@ class AccountAggregate:
         )
         if len(tables_with_location) <= 1:  # One table can not overlap
             return
-        for previous_table, table in tables_with_location[:-1], tables_with_location[1:]:
-            if table is not None and table.location.startswith(previous_table.location):
+        previous_table = tables_with_location[0]
+        for table in tables_with_location[1:]:
+            if table.location.startswith(previous_table.location):
                 logger.warning(f"Tables {previous_table} and {table} have overlapping locations")
+            else:
+                # Only set previous table if there is no overlap to capture overlaps between more than two tables
+                previous_table = table
 
     def validate(self) -> None:
         """Validate migration readiness across workspaces:
--- a/src/databricks/labs/ucx/account/aggregate.py
+++ b/src/databricks/labs/ucx/account/aggregate.py
@@ -34,8 +34,9 @@ class AssessmentObject:
 class TableFromWorkspace(Table):
     workspace_id: int = 0
 
-    def __str__(self):
-        return f"{self.workspace_id}:{super().__str__()}"
+    @property
+    def key(self) -> str:
+        return f"{self.workspace_id}:{super().key}"
 
 
 class AccountAggregate:
@@ -145,7 +146,7 @@ class AccountAggregate:
                 for conflicting_table in sub_node.tables:
                     conflicts.append(conflicting_table)
                     seen_tables.add(table.key)
-            conflict_message = " and ".join(str(conflict) for conflict in conflicts)
+            conflict_message = " and ".join(conflict.key for conflict in conflicts)
             logger.warning(f"Overlapping table locations: {conflict_message}")
             all_conflicts.append(conflicts)
         return all_conflicts
--- a/src/databricks/labs/ucx/account/aggregate.py
+++ b/src/databricks/labs/ucx/account/aggregate.py
@@ -136,12 +136,11 @@ class AccountAggregate:
                 continue
             seen_tables.add(str(table))
             node = trie.find(table)
-            if node is None:
+            if node is None or not node.has_children() or len(node.tables) == 1:
                 continue
-            if node.has_children() or len(node.tables) > 1:
-                conflicts = []
-                for sub_node in node:
-                    for t in sub_node.tables:
-                        conflicts.append(str(t))
-                        seen_tables.add(str(t))
-                logger.warning(f"Overlapping table locations: {' and '.join(conflicts)}")
+            conflicts = []
+            for sub_node in node:
+                for conflicting_table in sub_node.tables:
+                    conflicts.append(str(conflicting_table))
+                    seen_tables.add(str(conflicting_table))
+            logger.warning(f"Overlapping table locations: {' and '.join(conflicts)}")
--- a/src/databricks/labs/ucx/account/aggregate.py
+++ b/src/databricks/labs/ucx/account/aggregate.py
@@ -123,13 +123,16 @@ class AccountAggregate:
     def _validate_table_locations_not_overlapping(self) -> None:
         """The table locations should not be overlapping."""
         tables_with_location = sorted(
+            # Mypy does not catch the filter below
             filter(lambda table: table.location is not None, self._fetch_tables()),
-            key=lambda table: table.location,
+            key=lambda table: table.location or "",
         )
-        if len(tables_with_location) <= 1:  # One table can not overlap
+        if len(tables_with_location) <= 1:
             return
         previous_table = tables_with_location[0]
+        assert previous_table.location is not None
         for table in tables_with_location[1:]:
+            assert table.location is not None
             if table.location.startswith(previous_table.location):
                 logger.warning(f"Tables {previous_table} and {table} have overlapping locations")
             else:
--- a/src/databricks/labs/ucx/account/aggregate.py
+++ b/src/databricks/labs/ucx/account/aggregate.py
@@ -121,8 +121,9 @@ class AccountAggregate:
         for failure, objects in failures.items():
             logger.info(f"{failure}: {len(objects)} objects")
 
-    def _validate_table_locations_not_overlapping(self) -> None:
+    def validate_table_locations(self) -> None:
         """The table locations should not be overlapping."""
+        logger.info("Validating migration readiness")
         tables_with_location = sorted(
             # Mypy does not catch the filter below
             filter(lambda table: table.location is not None, self._fetch_tables()),
@@ -139,12 +140,3 @@ class AccountAggregate:
             else:
                 # Only set previous table if there is no overlap to capture overlaps between more than two tables
                 previous_table = table
-
-    def validate(self) -> None:
-        """Validate migration readiness across workspaces:
-
-        Readiness includes:
-        - Tables can be migrated
-        """
-        logger.info("Validating migration readiness")
-        self._validate_table_locations_not_overlapping()
--- a/src/databricks/labs/ucx/account/aggregate.py
+++ b/src/databricks/labs/ucx/account/aggregate.py
@@ -135,7 +135,7 @@ class AccountAggregate:
         for table in tables_with_location[1:]:
             assert table.location is not None
             if table.location.startswith(previous_table.location):
-                logger.warning(f"Tables {previous_table} and {table} have overlapping locations")
+                logger.warning(f"Overlapping table locations: {previous_table} and {table}")
             else:
                 # Only set previous table if there is no overlap to capture overlaps between more than two tables
                 previous_table = table
--- a/src/databricks/labs/ucx/account/aggregate.py
+++ b/src/databricks/labs/ucx/account/aggregate.py
@@ -135,7 +135,9 @@ class AccountAggregate:
         for table in tables:
             if table.key in seen_tables:
                 continue
-            seen_tables.add(str(table))
+            if table.location is None:
+                continue
+            seen_tables.add(table.key)
             node = trie.find(table)
             if node is None:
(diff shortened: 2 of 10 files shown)


labs.yml
```diff
@@ -51,6 +51,17 @@ commands:
       - name: workspace-ids
         description: List of workspace IDs to create account groups from.
 
+  - name: validate
+    is_account_level: true
+    description: |
+      Validate migration readiness across workspaces:
```
ericvergnaud, line 57: I'd call it 'feasability' rather than 'readiness'

src/databricks/labs/ucx/account/aggregate.py
```diff
@@ -28,6 +29,14 @@ class AssessmentObject:
     failures: list[str]
 
 
+@dataclass
+class Table(HiveMetastoreTable):
```
ericvergnaud, line 33: maybe rename to AccountTable ?

src/databricks/labs/ucx/account/aggregate.py
```diff
@@ -106,3 +119,24 @@ def readiness_report(self):
...
+            key=lambda table: table.location is not None,
+        )
+        if len(tables_with_location) <= 1:  # One table can not overlap
+            return
+        for previous_table, table in tables_with_location[:-1], tables_with_location[1:]:
+            if table is not None and table.location.startswith(previous_table.location):
+                logger.warning(f"Tables {previous_table} and {table} have overlapping locations")
+
+    def validate(self) -> None:
+        """Validate migration readiness across workspaces:
```
ericvergnaud, line 136: same here

src/databricks/labs/ucx/cli.py
```diff
@@ -101,6 +101,14 @@ def report_account_compatibility(a: AccountClient, ctx: AccountContext | None =
     ctx.account_aggregate.readiness_report()
 
 
+@ucx.command(is_account=True)
+def validate(a: AccountClient, ctx: AccountContext | None = None, **named_parameters):
+    """Validate migration readiness across workspaces."""
```
ericvergnaud, line 106: same here

ericvergnaud (changes requested):

labs.yml
JCZuurmond: We use 'readiness' more often, e.g. [here](https://github.com/databrickslabs/ucx/blob/cdb7709c3f430a1534321846f1454dc0a91cf74a/src/databricks/labs/ucx/account/aggregate.py#L90)

src/databricks/labs/ucx/account/aggregate.py
JCZuurmond: hmm, yes, there is no such thing as an "account table", maybe `TabeWithWorkspaceId` or `WorkspaceTable`

JCZuurmond:
@nfx : Please have a look at this PR. It implements flagging tables which have overlapping locations. Note that the feature is broader, it also discusses how to resolve tables with overlapping locations. I've added relevant resolution information to the README, but let resolving the issue be decided by the user

src/databricks/labs/ucx/account/aggregate.py
```diff
@@ -106,3 +120,31 @@ def readiness_report(self):
...
+        """The table locations should not be overlapping."""
+        tables_with_location = sorted(
+            # Mypy does not catch the filter below
+            filter(lambda table: table.location is not None, self._fetch_tables()),
+            key=lambda table: table.location or "",
+        )
+        if len(tables_with_location) <= 1:
+            return
+        previous_table = tables_with_location[0]
+        assert previous_table.location is not None
```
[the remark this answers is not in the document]
nfx: can we use a Trie data structure here? it'll allow to track overlapping paths and would improve our location discovery code in https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/hive_metastore/locations.py

```
@dataclass
class LocationTrie:
    part: str = ''
    parent: 'LocationTrie' = None
    children: dict[str, 'LocationTrie'] = dataclasses.field(default_factory=dict)
    tables: list[Table] =  dataclasses.field(default_factory=list)

    def find(self, path: str):
        parts = path.split('/')
        current = self
        for part in parts:
            if part not in current.children:
                parent = current
                current = LocationTrie(part, parent)
                parent.children[part] = current
                continue
            current = current.children[part]
        return current

    @cached_property
    def parts(self):
        parts = []
        current = self
        while current:
            parts.append(current.part)
            current = current.parent
        return list(reversed(parts))[1:]

    @property
    def full(self):
        return '/'.join(self.parts)

    def is_valid(self):
        """A valid path has at least 3 slash-parts: scheme, empty string, and bucket name: s3://bucket1"""
        return len(self.parts) >= 3

    def has_children(self):
        return len(self.children) > 0

    def __iter__(self):
        if self.is_valid():
            yield self
        for child in self.children.values():
            yield from child

def test_trie():
    locations = [
        's3://bucket1/a/b/c',
        's3://bucket1/a/b/d',
        's3://bucket1/a/b/e',
        's3://bucket1/a/b/f',
        's3://bucket1/a/b/f/g',
        's3://bucket1/a/b/f/h',
        's3://bucket1/b/c/d',
        's3://bucket1/b/e/f',
        's3://bucket1/c/d/e',
        's3://bucket1/c/e/f',
    ]
    root = LocationTrie()
    for location in locations:
        root.find(location)
    for node in root:
        print(node.full)

    f_node = root.find('s3://bucket1/a/b/f')
    assert f_node.has_children()

    d_node = root.find('s3://bucket1/a/b/d')
    assert not d_node.has_children()
```

nfx (changes requested):

src/databricks/labs/ucx/cli.py
```diff
@@ -101,6 +101,15 @@ def report_account_compatibility(a: AccountClient, ctx: AccountContext | None =
     ctx.account_aggregate.readiness_report()
 
 
+@ucx.command(is_account=True)
+def validate(a: AccountClient, ctx: AccountContext | None = None, **named_parameters):
+    """Validate migration readiness"""
+    logger.warning("The validate command is work in progress and does not cover a complete migration readiness (yet).")
```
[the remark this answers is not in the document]
JCZuurmond: I decided to make a specific `validate-location-tables` command, similar to the other `validate-` commands, instead of a generic `validate` command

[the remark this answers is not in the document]
nfx: having too many commands is also not good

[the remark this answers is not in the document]
JCZuurmond: We could merge the `validate-` commands, would like to separate that from this PR

src/databricks/labs/ucx/account/aggregate.py
```diff
@@ -106,3 +121,17 @@ def readiness_report(self):
...
+        trie = LocationTrie()
+        for table in sorted(self._fetch_tables(), key=lambda table: table.location or "", reverse=True):
+            if table.location is None:
+                continue
+            node = trie.find(table)
+            if node is not None:
+                for sub_node in node:
+                    for t in sub_node.tables:
+                        logger.warning(f"Overlapping table locations: {table} and {t}")
+            trie.insert(table)
```
nfx, line 137:
```suggestion
        all_tables = list(self._fetch_tables())
        # Step 1: create a trie from all tables
        for table in all_tables:
            if table.location is None:
                continue
            trie.insert(table)
        # Step 2: check for 
        seen: dict[str, str] = {}
        for table in all_tables:
            if not table.location: continue
            node = trie.find(table.location)
            if len(node.tables) > 1 or node.has_children():
               conflicts = []
               for sub in node:
                 for t in sub.tables:
                   conflicts.append(t)
              logger.warning(f"Overlapping table locations: {conflicts}") # better formatting
```

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -31,6 +37,71 @@ class Mount:
...
+
+    def insert(self, table: Table) -> None:
+        current = self
+        for part in self._parse_location(table.location):
+            if part not in current.children:
+                parent = current
+                current = LocationTrie(part, parent)
+                parent.children[part] = current
+            else:
+                current = current.children[part]
```
nfx, line 77:
```suggestion
                continue
            current = current.children[part]
```

tests/integration/account/test_aggregate.py
```diff
@@ -0,0 +1,32 @@
...
+    ]
+    sql_backend.save_table(f"{inventory_schema}.tables", tables, Table)
+
+    account_workspaces = AccountWorkspaces(acc)
+    w_ctx = WorkspaceContext(ws).replace(sql_backend=sql_backend, config=WorkspaceConfig(inventory_schema))
+    account_aggregate = AccountAggregate(
+        account_workspaces,
+        lambda w: w_ctx if w.get_workspace_id() == ws.get_workspace_id() else WorkspaceContext(w),
+    )
+    with caplog.at_level(logging.WARNING, logger="databricks.labs.ucx.account.aggregate"):
```
nfx, line 30: it's better to test on concrete return values than a log messages.

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -31,6 +37,71 @@ class Mount:
...
+        current.tables.append(table)
+
+    def find(self, table: Table) -> Optional["LocationTrie"]:
```
nfx, line 80:
```suggestion
    def find(self, table: Table) -> "LocationTrie" | None:
```

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -31,6 +37,71 @@ class Mount:
...
+        current = self
+        for part in self._parse_location(table.location):
+            if part not in current.children:
+                return None
+            current = current.children[part]
+        return current
+
+    def is_valid(self):
```
nfx, line 88:
```suggestion
    def is_valid(self) -> bool:
```

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -31,6 +37,71 @@ class Mount:
...
+            current = current.parent
+        return list(reversed(parts))[1:]
+
+    @property
+    def location(self):
+        scheme, netloc, *path = self._path
+        return f"{scheme}://{netloc}/{'/'.join(path)}"
+
+    @staticmethod
+    def _parse_location(location: str | None) -> list[str]:
```
nfx, line 63:
```suggestion
    def _parse_location(location: str | None) -> list[str]:
        if not location: return []
```

src/databricks/labs/ucx/account/aggregate.py
```diff
@@ -106,3 +121,31 @@ def readiness_report(self):
...
+        """The table locations should not be overlapping."""
+        logger.info("Validating migration readiness")
+        tables = list(self._fetch_tables())
+        trie = LocationTrie()
+        for table in tables:
+            if table.location is not None:
+                trie.insert(table)
+        seen_tables, all_conflicts = set(), []
+        for table in tables:
+            if str(table) in seen_tables:
```
nfx, line 135: use `table.key`

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -31,6 +37,87 @@ class Mount:
...
+
+@dataclass
+class LocationTrie:
+    """A trie datastructure to search locations.
+
+    Used to find overlapping locations.
+
+    TODO
+    ----
+    - [ ] : Include external locations pointing to SQL sources
```
nfx, line 52: it's probably `wontfix`

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -102,6 +102,9 @@ def __hash__(self):
     def __eq__(self, other):
         return isinstance(other, Table) and self.key == other.key
 
+    def __str__(self):
+        return f"{self.catalog}.{self.database}.{self.name}"
```
nfx, line 106: there's already `key` property

nfx (approved):
lgtm

src/databricks/labs/ucx/hive_metastore/locations.py
JCZuurmond: Removed, kept issue to decide what to do with it

PR #2343, opened 2024-08-01
Add support for skipping views when migrating tables and views

### LINKED ISSUE #1937: [BUG]: Allow ucx skip command to skip views

### Problem statement

Currently able to skip schema/tables but views are unsupported.

### Proposed Solution

Allow functionality to determine object type or additional parameter to specify object is a view.

### Additional Context

_No response_

ericvergnaud:
## Changes
Add support for skipping views when migrating tables and views

### Linked issues
Resolves #1937 

### Functionality
- [ ] modified existing command: `databricks labs ucx skip`

### Tests
- [x] added unit tests
- [x] evolved integration tests

### DIFF
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -81,8 +81,7 @@ def skip(w: WorkspaceClient, schema: str | None = None, table: str | None = None
         return None
     ctx = WorkspaceContext(w)
     if table:
-        is_view = ctx.tables_crawler.is_view(schema, table)
-        return ctx.table_mapping.skip_table_or_view(schema, table, is_view)
+        return ctx.table_mapping.skip_table_or_view(schema, table, ctx.tables_crawler)
     return ctx.table_mapping.skip_schema(schema)
 
 
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -81,7 +81,8 @@ def skip(w: WorkspaceClient, schema: str | None = None, table: str | None = None
         return None
     ctx = WorkspaceContext(w)
     if table:
-        return ctx.table_mapping.skip_table(schema, table)
+        is_view = ctx.tables_crawler.is_view(schema, table)
+        return ctx.table_mapping.skip_table_or_view(schema, table, is_view)
     return ctx.table_mapping.skip_schema(schema)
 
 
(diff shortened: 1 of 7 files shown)


src/databricks/labs/ucx/cli.py
```diff
@@ -81,7 +81,8 @@ def skip(w: WorkspaceClient, schema: str | None = None, table: str | None = None
         return None
     ctx = WorkspaceContext(w)
     if table:
-        return ctx.table_mapping.skip_table(schema, table)
+        is_view = ctx.tables_crawler.is_view(schema, table)
+        return ctx.table_mapping.skip_table_or_view(schema, table, is_view)
```
nfx, line 85:
```suggestion
        return ctx.table_mapping.skip_table_or_view(schema, table)
```
is_view should be a private method, otherwise it's a detrimental change for the future modifications of this code and would make it more confusing down the line. please refactor

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -362,6 +366,16 @@ def snapshot(self) -> list[Table]:
         """
         return self._snapshot(partial(self._try_load), partial(self._crawl))
 
+    def is_view(self, schema_name: str, table_name: str) -> bool:
+        table = self._try_load_one(schema_name, table_name)
+        return False if table is None else table.what == What.VIEW
```
nfx, line 371: `skip_table_or_view` would `ALTER TABLE` resulting in an error when table is not a view and is not found. just expose `try_load_one`, remove this method, and incorporate this logic in `skip_table_or_view`

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/tables.py
ericvergnaud: done

src/databricks/labs/ucx/cli.py
ericvergnaud: done

src/databricks/labs/ucx/hive_metastore/tables.py
ericvergnaud: done

tests/integration/hive_metastore/test_migrate.py
```diff
@@ -370,30 +370,61 @@ def test_revert_migrated_table(sql_backend, runtime_ctx, make_catalog):
 
 @retried(on=[NotFound], timeout=timedelta(minutes=5))
 def test_mapping_skips_tables_databases(ws, sql_backend, runtime_ctx, make_catalog):
```
nfx, line 372: what's the added runtime  for this test? the more test does, the more flaky it may become, so i'm wondering if it makes sense to grow this test case or add another one. it's always a tradeoff in the integration tests - it's fine to add more "faster ones", but better to extend "slower ones".

nfx (changes requested):

PR #2349, opened 2024-08-02
Fix linting for Spark Python workflow tasks

### LINKED ISSUE #2213: [BUG]: Job linting for Spark Python tasks only works with notebooks

### Current Behavior

The `WorkflowLinter` currently lints tasks (and dependencies) for Spark Python tasks (`Task.spark_python_task='/path/to/python_script.py'`). However it incorrectly assumes that these files must be notebooks and fails if they are a simple python script instead of a notebook, raising `ValueError: Not a Databricks notebook source!`

The backtrace leading to this is:
```
../src/databricks/labs/ucx/source_code/jobs.py:356: in lint_job
    problems = self._lint_job(job)
../src/databricks/labs/ucx/source_code/jobs.py:371: in _lint_job
    for path, advice in self._lint_task(task, job):
../src/databricks/labs/ucx/source_code/jobs.py:400: in _lint_task
    problems = container.build_dependency_graph(graph)
../src/databricks/labs/ucx/source_code/jobs.py:103: in build_dependency_graph
    return list(self._register_task_dependencies(parent))
../src/databricks/labs/ucx/source_code/jobs.py:110: in _register_task_dependencies
    yield from self._register_spark_python_task(graph)
../src/databricks/labs/ucx/source_code/jobs.py:200: in _register_spark_python_task
    return graph.register_notebook(path)
../src/databricks/labs/ucx/source_code/graph.py:54: in register_notebook
    maybe_graph = self.register_dependency(maybe.dependency)
../src/databricks/labs/ucx/source_code/graph.py:78: in register_dependency
    container = dependency.load(self.path_lookup)
../src/databricks/labs/ucx/source_code/graph.py:194: in load
    return self._loader.load_dependency(path_lookup, self)
../src/databricks/labs/ucx/source_code/notebooks/loaders.py:68: in load_dependency
    return Notebook.parse(absolute_path, content, language)
../src/databricks/labs/ucx/source_code/notebooks/sources.py:52: in parse
    cells = default_cell_language.extract_cells(source)
```

The start of the problem is here:
https://github.com/databrickslabs/ucx/blob/4ab771675e33e190d3b65b74caf67620be8e284d/src/databricks/labs/ucx/source_code/jobs.py#L200

### Expected Behavior

Jobs with Spark Python tasks that refer to ordinary Python files (and not notebooks) should also work, both on Workspace paths and DBFS.

### Steps To Reproduce

The following integration tests demonstrate the failure and will pass when the problem is fixed:

```python
def test_job_spark_python_task_workspace_linter_happy_path(
    simple_ctx,
    make_job,
    make_random,
    make_cluster,
    make_workspace_directory,
):
    pyspark_job_path = make_workspace_directory() / "spark_job.py"
    pyspark_job_path.write_text("import greenlet\n")

    new_cluster = make_cluster(single_node=True)
    task = jobs.Task(
        task_key=make_random(4),
        spark_python_task=jobs.SparkPythonTask(python_file=pyspark_job_path.as_posix()),
        existing_cluster_id=new_cluster.cluster_id,
        libraries=[compute.Library(pypi=compute.PythonPyPiLibrary(package="greenlet"))],
    )
    j = make_job(tasks=[task])

    problems = simple_ctx.workflow_linter.lint_job(j.job_id)
    assert not [problem for problem in problems if problem.message == "Could not locate import: greenlet"]


def test_job_spark_python_task_dbfs_linter_happy_path(
    simple_ctx,
    make_job,
    make_random,
    make_cluster,
    make_dbfs_directory,
):
    pyspark_job_path = make_dbfs_directory() / "spark_job.py"
    pyspark_job_path.write_text("import greenlet\n")

    new_cluster = make_cluster(single_node=True)
    task = jobs.Task(
        task_key=make_random(4),
        spark_python_task=jobs.SparkPythonTask(python_file=f"dbfs:{pyspark_job_path.as_posix()}"),
        existing_cluster_id=new_cluster.cluster_id,
        libraries=[compute.Library(pypi=compute.PythonPyPiLibrary(package="greenlet"))],
    )
    j = make_job(tasks=[task])

    problems = simple_ctx.workflow_linter.lint_job(j.job_id)
    assert not [problem for problem in problems if problem.message == "Could not locate import: greenlet"]
```

Note that the existing integration tests for Spark Python tasks (eg.  [`test_job_spark_python_task_linter_happy_path`](https://github.com/databrickslabs/ucx/blob/4ab771675e33e190d3b65b74caf67620be8e284d/tests/integration/source_code/test_jobs.py#L382)) create notebooks and not basic python files.

### Cloud

Azure

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

asnare:
## Changes

This PR updates the linter to support PySpark tasks on workflows; currently the linter assumes the python file is a notebook, but this is not the case: the file is a (top-level) python file that is loaded and run as `__main__`.

### Linked issues

Fixes #2213.

### Functionality

- modified existing workflow: `experimental-workflow-linter`

### Tests

- [X] updated unit tests
- [X] added integration tests


src/databricks/labs/ucx/source_code/jobs.py
```diff
@@ -194,10 +194,10 @@ def _register_spark_python_task(self, graph: DependencyGraph):
...
             return []
         self._parameters = self._task.spark_python_task.parameters
-        notebook_path = self._task.spark_python_task.python_file
-        logger.info(f'Discovering {self._task.task_key} entrypoint: {notebook_path}')
-        path = self._as_path(notebook_path)
-        return graph.register_notebook(path, False)
+        python_file = self._task.spark_python_task.python_file
+        logger.info(f'Discovering {self._task.task_key} entrypoint: {python_file}')
+        path = self._as_path(python_file)
+        return graph.register_file(path)
```
ericvergnaud, line 200: I believe we should check whether the file is a file or a notebook and register it accordingly. Notebooks need to be loaded by the NotebookLoader.

src/databricks/labs/ucx/source_code/graph.py
```diff
@@ -41,10 +41,6 @@ def path_lookup(self):
     def dependency(self):
         return self._dependency
 
-    @property
```
ericvergnaud, line 44: How does this relate to this PR ? Every node in the graph points to a path, so not sure this is a simplification ?

ericvergnaud (changes requested):

tests/integration/source_code/test_jobs.py
```diff
@@ -509,6 +509,52 @@ def test_workflow_linter_lints_python_wheel_task(simple_ctx, ws, make_job, make_
...
+        spark_python_task=jobs.SparkPythonTask(python_file=f"dbfs:{pyspark_job_path.as_posix()}"),
+        existing_cluster_id=new_cluster.cluster_id,
+        libraries=[compute.Library(pypi=compute.PythonPyPiLibrary(package="greenlet"))],
+    )
+    j = make_job(tasks=[task])
+
+    problems = simple_ctx.workflow_linter.lint_job(j.job_id)
+    assert not [problem for problem in problems if problem.message == "Could not locate import: greenlet"]
+
+
```
ericvergnaud, line 557: we're missing an integration test with a 'real' notebook (i.e. one with multiple cells, not just a 1-liner).

ericvergnaud (changes requested):

src/databricks/labs/ucx/source_code/graph.py
asnare: I removed this in 5993eea3df109249fb162f5781079a7dd74b3992, because our linting rules only allow for 20 public methods and: a) we had 20; b) another arrives in the next commit.

It's either remove an existing method, or add a linting suppression. I chose to remove this because it wasn't used externally and was trivial in nature.

src/databricks/labs/ucx/source_code/jobs.py
asnare: Some context here is that I don't think notebooks are fully supported by this task interface, but on Monday I will perform some tests. (And maybe add some additional automated tests to confirm/explain the expected behaviour.)

My expectation here is that this is effectively a `spark-submit` interface and that it expects just a python file, effectively as if being run by the python interpreter without any of the notebook-specific handling. It's possible that a notebook might work, but only if the notebook is both python and has no magic because such a notebook can be run simply as a python script.

tests/integration/source_code/test_jobs.py
asnare: As mentioned above, I'm not sure that this interface supports notebooks. I'll test and add a test (either way) to confirm the behaviour.

src/databricks/labs/ucx/source_code/jobs.py
asnare: My expectation was correct: this is a `spark-submit` interface that runs the associated file as a python script. It is not notebook-aware, so our linting needs to treat it the same way.

tests/integration/source_code/test_jobs.py
asnare: I've now added a test for this handling: it verifies that notebooks are _not_ processed as such (but rather just as a python file) because that's how these tasks work.

src/databricks/labs/ucx/contexts/application.py
```diff
@@ -393,7 +393,9 @@ def file_resolver(self):
 
     @cached_property
     def dependency_resolver(self):
-        return DependencyResolver(self.pip_resolver, self.notebook_resolver, self.file_resolver, self.path_lookup)
+        return DependencyResolver(
+            self.pip_resolver, self.notebook_resolver, self.file_resolver, self.file_resolver, self.path_lookup
```
nfx, line 397: looked a bit strange but okay

src/databricks/labs/ucx/source_code/graph.py
nfx: eh... i'd rather have it done in a separate PR

nfx (approved):
lgtm

PR #2364, opened 2024-08-05
Improve logging when waiting for workflows to complete

asnare:
## Changes

This PR updates the logging and error handling when we (from code) run workflows. Changes include:

 - Improved logging before the workflow starts, and status information when it completes.
 - The timeout can now be specified instead of always being 20 minutes.
 - If a workflow times out the logs from the job are now replicated into the local logfile.
 - Handling of `skip_job_wait` has been fixed: previously we would always raise `NotFound` as an error if this was `True`.

### Linked issues

Relates #2353.

### Functionality

- modified existing commands:

  - `databricks labs ucx ensure-assessment-run`
  - `databricks labs ucx migrate-tables`

### Tests

- existing integration tests

### DIFF
--- a/src/databricks/labs/ucx/installer/workflows.py
+++ b/src/databricks/labs/ucx/installer/workflows.py
@@ -42,7 +42,7 @@ from databricks.sdk.errors import (
 )
 from databricks.sdk.retries import retried
 from databricks.sdk.service import compute, jobs
-from databricks.sdk.service.jobs import RunLifeCycleState, RunResultState
+from databricks.sdk.service.jobs import Run, RunLifeCycleState, RunResultState
 from databricks.sdk.service.workspace import ObjectType
 
 import databricks
@@ -119,24 +119,54 @@ class DeployedWorkflows:
         self._install_state = install_state
         self._verify_timeout = verify_timeout
 
-    def run_workflow(self, step: str, skip_job_wait: bool = False):
+    def run_workflow(self, step: str, skip_job_wait: bool = False, max_wait: timedelta = timedelta(minutes=20)) -> int:
         # this dunder variable is hiding this method from tracebacks, making it cleaner
         # for the user to see the actual error without too much noise.
         __tracebackhide__ = True  # pylint: disable=unused-variable
         job_id = int(self._install_state.jobs[step])
         logger.debug(f"starting {step} job: {self._ws.config.host}#job/{job_id}")
         job_initial_run = self._ws.jobs.run_now(job_id)
-        if job_initial_run.run_id and not skip_job_wait:
-            try:
-                self._ws.jobs.wait_get_run_job_terminated_or_skipped(run_id=job_initial_run.run_id)
-            except OperationFailed as err:
-                logger.info('---------- REMOTE LOGS --------------')
-                self._relay_logs(step, job_initial_run.run_id)
-                logger.info('---------- END REMOTE LOGS ----------')
-                job_run = self._ws.jobs.get_run(job_initial_run.run_id)
-                raise self._infer_error_from_job_run(job_run) from err
+        run_id = job_initial_run.run_id
+        if not run_id:
+            raise NotFound(f"job run not found for {step}")
+        run_url = f"{self._ws.config.host}#job/{job_id}/runs/{run_id}"
+        logger.info(f"Started {step} job: {run_url}")
+        if skip_job_wait:
+            return run_id
+        try:
+            logger.debug(f"Waiting for completion of {step} job: {run_url}")
+            job_run = self._ws.jobs.wait_get_run_job_terminated_or_skipped(run_id=run_id, timeout=max_wait)
+            self._log_completed_job(step, job_run)
+            return run_id
+        except TimeoutError:
+            logger.warning(f"Timeout while waiting for {step} job to complete: {run_url}")
+            logger.info('---------- REMOTE LOGS --------------')
+            self._relay_logs(step, run_id)
+            logger.info('------ END REMOTE LOGS (SO FAR) -----')
+            raise
+        except OperationFailed as err:
+            logger.info('---------- REMOTE LOGS --------------')
+            self._relay_logs(step, run_id)
+            logger.info('---------- END REMOTE LOGS ----------')
+            job_run = self._ws.jobs.get_run(run_id)
+            raise self._infer_error_from_job_run(job_run) from err
+
+    def _log_completed_job(self, step: str, job_run: Run) -> None:
+        if not logger.isEnabledFor(logging.INFO):
             return
-        raise NotFound(f"job run not found for {step}")
+        start_time = datetime.utcfromtimestamp(job_run.start_time / 1000) if job_run.start_time else None
+        end_time = datetime.utcfromtimestamp(job_run.end_time / 1000) if job_run.end_time else None
+        if job_run.run_duration:
+            duration = timedelta(milliseconds=job_run.run_duration)
+        elif start_time and end_time:
+            duration = end_time - start_time
+        else:
+            duration = None
+        result_state = job_run.state.result_state if job_run.state else None
+        state_message = job_run.state.state_message if job_run.state else None
+        logger.info(
+            f"Completed {step} job: {result_state or 'N/A'} ({state_message or 'N/A'}) {start_time or 'N/A'}-{end_time or 'N/A'} ({duration or 'N/A'})"
+        )
 
     def repair_run(self, workflow):
         try:


src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -119,24 +119,61 @@ def __init__(self, ws: WorkspaceClient, install_state: InstallState, verify_time
...
+        except OperationFailed as err:
+            logger.info('---------- REMOTE LOGS --------------')
+            self._relay_logs(step, run_id)
+            logger.info('---------- END REMOTE LOGS ----------')
+            job_run = self._ws.jobs.get_run(run_id)
+            raise self._infer_error_from_job_run(job_run) from err
+
+    @staticmethod
+    def _log_completed_job(step: str, run_id: int, job_run: Run) -> None:
+        if not logger.isEnabledFor(logging.INFO):
```
JCZuurmond, line 156: Why this?

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -119,24 +119,61 @@ def __init__(self, ws: WorkspaceClient, install_state: InstallState, verify_time
...
+        run_id = job_initial_run.run_id
+        if not run_id:
+            raise NotFound(f"job run not found for {step}")
+        run_url = f"{self._ws.config.host}#job/{job_id}/runs/{run_id}"
+        logger.info(f"Started {step} job: {run_url}")
+        if skip_job_wait:
+            return run_id
+        try:
+            logger.debug(f"Waiting for completion of {step} job: {run_url}")
+            job_run = self._ws.jobs.wait_get_run_job_terminated_or_skipped(run_id=run_id, timeout=max_wait)
```
JCZuurmond, line 138: Maybe reuse `self._verify_timeout`. Or add a `self._job_run_timeout` or something alike

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -119,24 +119,61 @@ def __init__(self, ws: WorkspaceClient, install_state: InstallState, verify_time
         self._install_state = install_state
         self._verify_timeout = verify_timeout
 
-    def run_workflow(self, step: str, skip_job_wait: bool = False):
+    def run_workflow(self, step: str, skip_job_wait: bool = False, max_wait: timedelta = timedelta(minutes=20)) -> int:
```
JCZuurmond, line 122: max_wait should only be given for test runs

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -119,24 +119,61 @@ def __init__(self, ws: WorkspaceClient, install_state: InstallState, verify_time
...
+        if not logger.isEnabledFor(logging.INFO):
             return
-        raise NotFound(f"job run not found for {step}")
+        if job_run.state:
+            result_state = job_run.state.result_state or "N/A"
+            state_message = job_run.state.state_message
+            state_description = f"{result_state} ({state_message})" if state_message else f"{result_state}"
+            logger.info(f"Completed {step} job run {run_id} with state: {state_description}")
+        else:
+            logger.info(f"Completed {step} job run {run_id}, but end state unknown.")
```
JCZuurmond, line 164:
```suggestion
            logger.info(f"Completed {step} job run {run_id} with unknown state.")
```

Or rewrite the above to be `state_description = "UNKNOWN"` when `not job_run.state`

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -119,24 +119,61 @@ def __init__(self, ws: WorkspaceClient, install_state: InstallState, verify_time
...
+        if job_run.start_time or job_run.end_time:
+            start_time = datetime.utcfromtimestamp(job_run.start_time / 1000) if job_run.start_time else None
```
JCZuurmond, line 166: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp

> Deprecated since version 3.12: Use [datetime.fromtimestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp) with [UTC](https://docs.python.org/3/library/datetime.html#datetime.UTC) instead.

```suggestion
            start_time = datetime.utcfromtimestamp(job_run.start_time / 1000) if job_run.start_time else None
```

src/databricks/labs/ucx/installer/workflows.py
JCZuurmond: Add a comment explaining that this early return should be here as long as we are logging info statements below. IMO it can also be removed, the code below does not hurt too much

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -119,24 +119,61 @@ def __init__(self, ws: WorkspaceClient, install_state: InstallState, verify_time
...
+            logger.debug(f"Waiting for completion of {step} job: {run_url}")
+            job_run = self._ws.jobs.wait_get_run_job_terminated_or_skipped(run_id=run_id, timeout=max_wait)
+            self._log_completed_job(step, run_id, job_run)
+            return run_id
+        except TimeoutError:
+            logger.warning(f"Timeout while waiting for {step} job to complete: {run_url}")
+            logger.info('---------- REMOTE LOGS --------------')
+            self._relay_logs(step, run_id)
+            logger.info('------ END REMOTE LOGS (SO FAR) -----')
+            raise
```
JCZuurmond, line 146: Does this raise the `TimeoutError` again?

JCZuurmond (changes requested):
See comments

src/databricks/labs/ucx/installer/workflows.py
asnare: Yes, it re-raises the existing exception.

Reference: https://docs.python.org/3/reference/simple_stmts.html#raise

src/databricks/labs/ucx/installer/workflows.py
asnare: That's interesting… I understand the argument, but am also wondering about the other main call location: CLI commands.

As things stand they can either not-wait, or wait. If waiting then right now they will error-out if the workflow takes longer than 20min. This wasn't a deliberate choice, but a default behaviour and I wonder if instead the CLI should wait indefinitely (or much longer). Changing the behaviour of the CLI in this respect is beyond the scope of this PR, but I imagine we have clients where the workflows can take hours and it's a completely normal situation instead of something erroneous.

src/databricks/labs/ucx/installer/workflows.py
asnare: I've dropped it; this isn't really performance-sensitive code, and there was a situation I wanted a warning for anyway so that's a win-win.

src/databricks/labs/ucx/installer/workflows.py
asnare: I've adjusted the logging here: it's now a warning because this is an abnormal situation. The wording has also been tweaked. Does it now look better?

I distinctly want different wording for this compared to the normal situation: in general each logging statement should be distinct so that a log-entry can unambiguously traced to the line on which it was generated; this makes troubleshooting much easier.

src/databricks/labs/ucx/installer/workflows.py
asnare: For this class, I happen to think that the `verify_timeout` property has been badly designed: in general it shouldn't be necessary to configure an instance with many properties when each only applies to narrow subset of its behaviour (in this case the call to `.repair_run()`). This applies especially when it can reasonably vary per call, as well as something that the direct caller cares about most.

I think it would be cleaner to refactor this so that the timeout for `.repair_run()` is _also_ an argument on the method instead of something indirectly specified via the constructor.

What do you think?

JCZuurmond: Agreed, but not in this PR. Could you log it as an `[TECH DEBT]` issue?

src/databricks/labs/ucx/installer/workflows.py
JCZuurmond: Could you log the issue to resolve this for the cli? For this PR, the behavior is preserved, therefore fine

JCZuurmond (approved):
LGTM, given the two issues for future tech debt

src/databricks/labs/ucx/installer/workflows.py
asnare: Done; I've created #2384 to cover updating the CLI behaviour.

nfx, line 122: small nit: let's use `timeout` for the consistency with SDK idioms. also - why do we need `skip_job_wait` if we have waiting timeout? `timedelta(seconds=0)` would mean no wait, right?

nfx (commented):
small nit

Issue #2361, opened 2024-08-05
 [FEATURE]: Add `delete-uber-principal` commands


JCZuurmond:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

-

### Proposed Solution

-

PR #2370, opened 2024-08-06
Clean up left over uber principal resources for Azure

### LINKED ISSUE #2360: [BUG]: `create-uber-principal` command have to cleanup/resumability
### Current Behavior

`create-uber-principal` commands creates residue roles/policies/etc.

### Expected Behavior

When the commands fails, we have to cleanup any residue created by those commands. 

### Cloud

AWS, Azure

### LINKED ISSUE #2363: [TECH DEBT]: Let `AzureResourcePermissions._get_storage_accounts` return `databricks.labs.ucx.azure.resources.StorageAccount` instead of `str`

### Problem statement

Let `AzureResourcePermissions._get_storage_accounts` return `databricks.labs.ucx.azure.resources.StorageAccount` instead of `str`. Now, we have code repetition every where `_get_storage_accounts` is used, the following loop is applied to get the `StorageAccount` object:

``` python
storage_accounts = []
for storage in self._azurerm.storage_accounts():
    if storage.name in used_storage_accounts:
        storage_accounts.append(storage)
``` 

### Proposed Solution

To reduce code repetition, move the above loop into `_get_storage_accounts` and adjust the code to use the `StorageAccount` object instead of the `str`.

### Additional Context

_No response_

JCZuurmond:
## Changes
Clean up the resources created for the Azure uber principal if the creation fails (half way).

### Linked issues
Partially resolves #2360 (the azure part)
Resolves #2363

### Functionality

- [x] modified existing command: `databricks labs ucx create-uber-principal`

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/azure/access.py
+++ b/src/databricks/labs/ucx/azure/access.py
@@ -419,11 +419,14 @@ class AzureResourcePermissions:
         for storage in self._azurerm.storage_accounts():
             if storage.name in used_storage_accounts:
                 storage_accounts.append(storage)
-        self._azurerm.delete_storage_permission(config.uber_spn_id, *storage_accounts, safe=True)
         try:
+            self._azurerm.delete_storage_permission(config.uber_spn_id, *storage_accounts, safe=True)
             self._azurerm.delete_service_principal(config.uber_spn_id, safe=True)
         except PermissionDenied:
-            logger.error(f"Missing permissions to delete service principal: {config.uber_spn_id}", exc_info=True)
+            logger.error(
+                f"Missing permissions to delete service principal and its permissions: {config.uber_spn_id}",
+                exc_info=True,
+            )
         secret_identifier = f"secrets/{config.inventory_database}/{self._UBER_PRINCIPAL_SECRET_KEY}"
         if config.policy_id is not None:
             try:
--- a/src/databricks/labs/ucx/azure/access.py
+++ b/src/databricks/labs/ucx/azure/access.py
@@ -410,7 +410,7 @@ class AzureResourcePermissions:
             self._delete_uber_principal()  # Clean up dangling resources
             raise
 
-    def _delete_uber_principal(self):
+    def _delete_uber_principal(self) -> None:
         config = self._installation.load(WorkspaceConfig)
         if config.uber_spn_id is None:
             return
--- a/src/databricks/labs/ucx/azure/access.py
+++ b/src/databricks/labs/ucx/azure/access.py
@@ -420,14 +420,7 @@ class AzureResourcePermissions:
             if storage.name in used_storage_accounts:
                 storage_accounts.append(storage)
         for storage_account in storage_accounts:
-            try:
-                self._azurerm.delete_storage_permission(config.uber_spn_id, storage_account)
-            except NotFound:
-                pass  # Already deleted
-            except PermissionDenied:
-                logger.error(
-                    f"Missing permissions to delete storage permission for {storage_account.id}", exc_info=True
-                )
+            self._azurerm.delete_storage_permission(config.uber_spn_id, storage_account, safe=True)
         try:
             self._azurerm.delete_service_principal(config.uber_spn_id)
         except NotFound:
--- a/src/databricks/labs/ucx/azure/access.py
+++ b/src/databricks/labs/ucx/azure/access.py
@@ -210,8 +210,8 @@ class AzureResourcePermissions:
         principal_secret_identifier: str,
         storage_accounts: list[StorageAccount],
     ):
+        policy_definition = ""
         try:
-            policy_definition = ""
             cluster_policy = self._ws.cluster_policies.get(policy_id)
             self._installation.save(cluster_policy, filename="policy-backup.json")
             if cluster_policy.definition is not None:
--- a/src/databricks/labs/ucx/azure/access.py
+++ b/src/databricks/labs/ucx/azure/access.py
@@ -2,10 +2,11 @@ import json
 import logging
 import re
 import uuid
-from collections.abc import ValuesView
+from collections.abc import Callable, ValuesView
 from dataclasses import dataclass
-from functools import partial
+from functools import partial, wraps
 from datetime import timedelta
+from typing import Any, ParamSpec, TypeVar
 
 from databricks.labs.blueprint.installation import Installation
 from databricks.labs.blueprint.parallel import ManyError, Threads
@@ -29,6 +30,8 @@ from databricks.labs.ucx.config import WorkspaceConfig
 from databricks.labs.ucx.hive_metastore.locations import ExternalLocations
 
 logger = logging.getLogger(__name__)
+P = ParamSpec('P')
+R = TypeVar('R')
 
 
 @dataclass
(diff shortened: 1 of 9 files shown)


src/databricks/labs/ucx/azure/access.py
```diff
@@ -183,63 +187,98 @@ def _policy_config(value: str):
...
+        )
+
+    def _revert_sql_dac_with_spn(
+        self,
+        storage_accounts: list[StorageAccount],
+        principal: PrincipalSecret,
+        principal_secret_identifier: str,
+    ):
+        warehouse_config = self._installation.load(
+            GetWorkspaceWarehouseConfigResponse, filename="warehouse-config-backup.json"
```
JCZuurmond, line 271: Need to test what happens if this does not exists

src/databricks/labs/ucx/azure/access.py
```diff
@@ -267,27 +307,57 @@ def create_uber_principal(self, prompts: Prompts):
...
+        if config.uber_spn_id is not None:
+            used_storage_accounts = self._get_storage_accounts()
+            storage_accounts = []
+            for storage in self._azurerm.storage_accounts():
+                if storage.name in used_storage_accounts:
+                    storage_accounts.append(storage)
+            for storage_account in storage_accounts:
+                try:
+                    self._azurerm.delete_storage_permission(str(storage_account.id), principal_id=config.uber_spn_id)
+                except NotFound:
```
JCZuurmond, line 344: These are databricks exceptions while azure resources are deleted

nfx: because we use Databricks SDK API client. that's fine

src/databricks/labs/ucx/azure/access.py
```diff
@@ -151,16 +156,16 @@ def save_spn_permissions(self) -> str | None:
     def _update_cluster_policy_definition(
         self,
         policy_definition: str,
+        principal_client_id: str,
+        principal_secret_identifier: str,
         storage_accounts: list[StorageAccount],
-        uber_principal: PrincipalSecret,
```
nfx, line 155: another way could be extending the `PrincipalSecret` class and add secret scope props to it, but this way is fine

src/databricks/labs/ucx/azure/access.py
```diff
@@ -183,63 +188,98 @@ def _policy_config(value: str):
...
+                    storage_accounts,
                 )
             if cluster_policy.name is not None:
                 self._ws.cluster_policies.edit(policy_id, cluster_policy.name, definition=policy_definition)
         except NotFound:
             msg = f"cluster policy {policy_id} not found, please run UCX installation to create UCX cluster policy"
             raise NotFound(msg) from None
 
+    def _revert_cluster_policy(self, policy_id: str):
+        policy = self._installation.load(Policy, filename="policy-backup.json")
```
nfx, line 213: which policy? ucx policy? be a bit specific with naming

src/databricks/labs/ucx/azure/access.py
```diff
@@ -183,63 +188,98 @@ def _policy_config(value: str):
...
+
+    def _revert_sql_dac_with_spn(
+        self,
+        principal_client_id: str,
+        principal_secret_identifier: str,
+        storage_accounts: list[StorageAccount],
+    ):
+        try:
+            warehouse_config = self._installation.load(
+                GetWorkspaceWarehouseConfigResponse, filename="warehouse-config-backup.json"
```
nfx, line 270:
```suggestion
                GetWorkspaceWarehouseConfigResponse, filename="workspace-warehouse-config-backup.json"
```
might be a bit better name for consistency

src/databricks/labs/ucx/azure/access.py
```diff
@@ -267,27 +307,62 @@ def create_uber_principal(self, prompts: Prompts):
...
-        logger.info(f"Update UCX cluster policy {policy_id} with spn connection details for storage accounts")
+            self._update_sql_dac_with_spn(uber_principal.client.client_id, secret_identifier, storage_accounts)
+        except (PermissionError, NotFound):
+            logger.error("Failed to create service principal", exc_info=True)
+            self._delete_uber_principal()  # Clean up dangling resources
+            return
+        config.uber_spn_id = uber_principal.client.client_id
+        self._installation.save(config)
+        logger.info(f"Created service principal ({config.uber_spn_id}) with access to used storage accounts.")
+        logger.info(f"Updated UCX cluster policy {policy_id} with spn connection details for storage accounts")
```
nfx, line 336: don't log two consecutive events if they can be merged to one - it's always a red flag. rather move these logger calls to proper places.

src/databricks/labs/ucx/azure/access.py
```diff
@@ -267,27 +307,62 @@ def create_uber_principal(self, prompts: Prompts):
...
+
+    def _delete_uber_principal(self):
+        config = self._installation.load(WorkspaceConfig)
+        if config.uber_spn_id is not None:
```
nfx, line 340: reduce method nesting level by `if not config.uber_spn_id: return`

src/databricks/labs/ucx/azure/access.py
```diff
@@ -267,27 +307,62 @@ def create_uber_principal(self, prompts: Prompts):
...
+        except (PermissionError, NotFound):
+            logger.error("Failed to create service principal", exc_info=True)
+            self._delete_uber_principal()  # Clean up dangling resources
+            return
+        config.uber_spn_id = uber_principal.client.client_id
+        self._installation.save(config)
+        logger.info(f"Created service principal ({config.uber_spn_id}) with access to used storage accounts.")
+        logger.info(f"Updated UCX cluster policy {policy_id} with spn connection details for storage accounts")
+
+    def _delete_uber_principal(self):
```
nfx, line 338:
```suggestion
    def _delete_uber_principal(self) -> None:
```
apparently mypy likes that

src/databricks/labs/ucx/azure/access.py
```diff
@@ -267,27 +307,62 @@ def create_uber_principal(self, prompts: Prompts):
...
+                self._azurerm.delete_service_principal(config.uber_spn_id)
+            except NotFound:
+                pass  # Already deleted
+            except PermissionDenied:
+                logger.error(f"Missing permissions to delete service principal: {config.uber_spn_id}", exc_info=True)
+            secret_identifier = f"secrets/{config.inventory_database}/{self._UBER_PRINCIPAL_SECRET_KEY}"
+            self._revert_sql_dac_with_spn(config.uber_spn_id, secret_identifier, storage_accounts)
+        if config.policy_id is not None:
+            self._revert_cluster_policy(config.policy_id)
+        self._safe_delete_scope(config.inventory_database)
```
nfx, line 365: add a bit more failsafes around deletions - so that DAC deletion timeout (or any other IO error) would still get us to deletion of secret scope or policy. I assume we should also want to expose some sort of `cleanup` command.

src/databricks/labs/ucx/azure/access.py
JCZuurmond: yes, could be confusing as the PrincipalSecret is an Azure resource, where the secret scope and identifier are Databricks resources, but they belong together

src/databricks/labs/ucx/azure/access.py
JCZuurmond: Created an issue for exposing the a `cleanup` command: https://github.com/databrickslabs/ucx/issues/2361

Could explain your failsafe a bit more? Now, it is indeed a "all or nothing": if anything fails during the creation of the uber SP, we delete everything that has been created. You want different behavior? Like the cluster policy and warehouse config should not trigger the revert?

src/databricks/labs/ucx/azure/access.py
```diff
@@ -130,116 +145,210 @@ def _map_storage(self, storage: StorageAccount) -> ValuesView[StoragePermissionM
...
-                    ),
-                    EndpointConfPair(
-                        f"spark_conf.fs.azure.account.auth.type.{storage.name}.dfs.core.windows.net", "OAuth"
-                    ),
-                    EndpointConfPair(
-                        f"spark_conf.fs.azure.account.oauth2.client.secret.{storage.name}.dfs.core.windows.net",
-                        "{{secrets/" + inventory_database + "/uber_principal_secret}}",
-                    ),
-                ]
+        return [
```
nfx, line 288: can you use `_create_service_principal_cluster_policy_configuration_pairs` method output? it's duplicated code, per se :)

src/databricks/labs/ucx/azure/access.py
```diff
@@ -130,116 +145,210 @@ def _map_storage(self, storage: StorageAccount) -> ValuesView[StoragePermissionM
...
+            warehouse_config = self._ws.warehouses.get_workspace_warehouse_config()
+        sql_dac = warehouse_config.data_access_config or []
+
+        for storage_account in storage_accounts:
+            configuration_pairs = self._create_storage_account_data_access_configuration_pairs(
+                principal_client_id, principal_secret_identifier, storage_account
             )
+            for configuration_pair in configuration_pairs:
+                if configuration_pair in sql_dac:
+                    sql_dac.remove(configuration_pair)
```
nfx, line 350: does it have a unit test? i'm afraid that `EndpointConfigInfo` may not have `__eq__` and `__hash__` generated for this to work.

src/databricks/labs/ucx/azure/access.py
```diff
@@ -260,34 +369,76 @@ def create_uber_principal(self, prompts: Prompts):
...
+            )
+        message = "Missing permissions to revert SQL warehouse config"
+        log_permission_denied(
+            self._remove_service_principal_configuration_from_workspace_warehouse_config, message=message
+        )(config.uber_spn_id, secret_identifier, storage_accounts)
+        message = "Missing permissions to delete secret scope"
+        log_permission_denied(self._safe_delete_scope, message=message)(config.inventory_database)
+        message = "Missing permissions to save the configuration"
+        config.uber_spn_id = None
+        log_permission_denied(self._installation.save, message=message)(config)
```
nfx, line 441: i see `self._installation.load(WorkspaceConfig)`, but i don't see `self._installation.save(config)`

src/databricks/labs/ucx/azure/access.py
```diff
@@ -355,25 +504,50 @@ def _apply_storage_permission(
...
-            self._ws.secrets.create_scope(inventory_database)
+            self._ws.secrets.create_scope(scope)
         except ResourceAlreadyExists:
-            logger.warning(f"Secret scope {inventory_database} already exists, using the same")
-        self._ws.secrets.put_secret(inventory_database, "uber_principal_secret", string_value=uber_principal.secret)
+            logger.warning(f"Secret scope {scope} already exists, using the same")
+        self._ws.secrets.put_secret(scope, self._UBER_PRINCIPAL_SECRET_KEY, string_value=principal_secret.secret)
+        return self._get_secret(scope, self._UBER_PRINCIPAL_SECRET_KEY)
+
+    @retried(on=[ResourceDoesNotExist], timeout=timedelta(minutes=2))
```
nfx, line 524:
```suggestion
    @retried(on=[ResourceDoesNotExist], timeout=timedelta(minutes=1))
```
one minute is fine.

src/databricks/labs/ucx/azure/resources.py
```diff
@@ -322,22 +324,129 @@ def create_service_principal(self, display_name: str) -> PrincipalSecret:
...
         assert directory_id is not None
         assert secret is not None
-        return PrincipalSecret(Principal(client_id, display_name, object_id, principal_type, directory_id), secret)
-
-    def delete_service_principal(self, principal_id: str):
+        principal_secret = PrincipalSecret(
+            Principal(client_id, display_name, object_id, principal_type, directory_id), secret
+        )
+        logger.info(
+            f"Created service principal ({principal_secret.client.client_id}) with access to used storage accounts."
```
nfx, line 331: nit: also log display name

nfx (changes requested):
i see self._installation.load(WorkspaceConfig), but i don't see self._installation.save(config)

src/databricks/labs/ucx/azure/access.py
```diff
@@ -355,25 +492,50 @@ def _apply_storage_permission(
...
             if location.location.startswith("abfss://"):
                 start = location.location.index("@")
                 end = location.location.index(".dfs.core.windows.net")
                 storage_acct = location.location[start + 1 : end]
-                if storage_acct not in storage_accounts:
-                    storage_accounts.append(storage_acct)
+                if storage_acct not in used_storage_accounts:
+                    used_storage_accounts.append(storage_acct)
+        storage_accounts = []
+        for storage_account in self._azurerm.storage_accounts():
```
JCZuurmond, line 538: Added this (small) refactor because the complexity in `_delete_uber_principal` was too high with this (duplicated) logic

src/databricks/labs/ucx/azure/access.py
JCZuurmond: Good catch! It was not tested, now it is. It works: https://github.com/databrickslabs/ucx/pull/2370/commits/6b6b4d95b30067627ae60fc34391750a792bee47

src/databricks/labs/ucx/azure/access.py
```diff
@@ -25,6 +37,8 @@
 from databricks.labs.ucx.hive_metastore.locations import ExternalLocations
 
 logger = logging.getLogger(__name__)
+P = ParamSpec('P')
+R = TypeVar('R')
```
nfx, line 41: are they even used?... 

UPDATE: found usage. define type params close to their usage, as it confuses while reading code.

nfx (approved):
lgtm.

src/databricks/labs/ucx/azure/access.py
JCZuurmond: hmm, it's convention to have them as globals, so to make this makes sense. Do you want to make it part of the class then? It is above the class which uses it, with a small dataclass in between.

PR #2380, opened 2024-08-07
Spike: Use a wheelhouse to install ucx at runtime

JCZuurmond:
## Changes
The changes are a spike to use a wheelhouse to install ucx at runtime. It is part of the bigger plan to support offline installs, which require changes on the [cli](https://github.com/databricks/cli/issues/1646) as well

### Functionality

- [ ] modified existing command: `databricks labs ucx install`

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,10 @@ share/python-wheels/
 *.egg
 !tests/integration/source_code/samples/library-egg/*.egg
 MANIFEST
+# TODO: Do we want to keep (generated) requirements.txt in git?
+# Pro: it would allow us to update single requirements instead of all
+requirements.txt
+wheelhouse.zip
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
--- a/.gitignore
+++ b/.gitignore
@@ -31,8 +31,6 @@ share/python-wheels/
 *.egg
 !tests/integration/source_code/samples/library-egg/*.egg
 MANIFEST
-# TODO: Do we want to keep (generated) requirements.txt in git?
-# Pro: it would allow us to update single requirements instead of all
 requirements.txt
 ucx.wheelhouse.zip
 
--- a/.gitignore
+++ b/.gitignore
@@ -34,7 +34,7 @@ MANIFEST
 # TODO: Do we want to keep (generated) requirements.txt in git?
 # Pro: it would allow us to update single requirements instead of all
 requirements.txt
-wheelhouse.zip
+ucx.wheelhouse.zip
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,12 +106,10 @@ verify      = ["black --check . --extend-exclude 'tests/unit/source_code/samples
                "pylint --output-format=colorized -j 0 src tests"]
 lint         = ["pylint --output-format=colorized -j 0 src tests"]
 cmds         = "python tests/integration/cmds.py"
-wheelhouse   = [
-    "rm -rf wheels/ ucx.wheelhouse.zip",
-    "pip-compile --all-build-deps --all-extras --output-file=requirements.txt --strip-extras pyproject.toml",
-    "pip download --platform any --no-deps -r requirements.txt --dest wheels/",
-    "zip -j -r ucx.wheelhouse.zip wheels/*.whl requirements.txt",
-]
+wheelhouse   = ["rm -rf wheels/ ucx.wheelhouse.zip",
+                "pip-compile --all-build-deps --all-extras --output-file=requirements.txt --strip-extras pyproject.toml",
+                "pip download --platform any --no-deps -r requirements.txt --dest wheels/",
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -31,6 +31,10 @@ share/python-wheels/
 *.egg
 !tests/integration/source_code/samples/library-egg/*.egg
 MANIFEST
+# TODO: Do we want to keep (generated) requirements.txt in git?
```
JCZuurmond, line 34: @nfx : What do you think? Should we keep this in git to be able to upgrade dependencies one by one? Or should we always generate the latest wheelhouse

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -567,11 +569,30 @@ def _library_dep_order(library: str):
...
 
-    def _upload_wheel(self):
+    def _upload_wheelhouse(self, path: Path) -> list[str]:
+        """Upload a wheelhouse.
+
+        TODO
+        ----
+        Move this method into the WheelsV2 class.
+        """
+        # We can not upload a .zip file to the workspace as the platform fails when extracting all files inside the zip
```
JCZuurmond, line 579: There are some mentions of a wheelhouse on the Databricks documentation pages (see below). BUT:
- if I upload a zip file to the workspace via the Databricks UI, the zip is extracted
- Uploading a zip via the API does not work
- The docs mention the wheelhouse should have the following extension: `.wheelhouse.zip`

The following also works, but it is not using a wheelhouse within Databricks

Doc pages:
- https://docs.databricks.com/en/jobs/jobs-2.0-api.html#library
- https://docs.databricks.com/en/libraries/cluster-libraries.html#install-a-library-on-a-cluster

JCZuurmond (commented):
@nfx : Please have a look at the changes in the PR. It shows how to make a wheelhouse and how to use it within ucx. 

To let users use it, we could extend the question about the workspace blocking internet access to let users provide a path to the ucx wheelhouse. Then we create the wheelhouse during a release and upload it to Github and tell users they can download it there for workspaces that block internet access.

Alternatively, we could make the wheelhouse part of ucx and always use it to install the dependencies at runtime.

In parallel, we can [extend the cli](https://github.com/databricks/cli/issues/1646) to support offline installs. 

Finally, should the wheelhouse include ucx itself or only its dependencies

JCZuurmond:
To properly test this: run a job on a cluster with a policy that blocks access to PyPi

.gitignore
nfx: we'll generate wheelhouse on release. we have a `pyproject.toml` to manage dependencies.

pyproject.toml
```diff
@@ -105,6 +106,10 @@ verify      = ["black --check . --extend-exclude 'tests/unit/source_code/samples
                "pylint --output-format=colorized -j 0 src tests"]
 lint         = ["pylint --output-format=colorized -j 0 src tests"]
 cmds         = "python tests/integration/cmds.py"
+wheelhouse   = ["rm -rf wheels/ ucx.wheelhouse.zip",
+                "pip-compile --all-extras --output-file=requirements.txt --strip-extras pyproject.toml",
+                "pip download --platform any --no-deps -r requirements.txt --dest wheels/",
+                "zip -j -r ucx.wheelhouse.zip wheels/*.whl requirements.txt"]
```
nfx, line 112: what about labs.yaml?

src/databricks/labs/ucx/install.py
```diff
@@ -235,8 +235,14 @@ def _prompt_for_new_installation(self) -> WorkspaceConfig:
         configure_groups.run()
         include_databases = self._select_databases()
         upload_dependencies = self.prompts.confirm(
-            f"Does given workspace {self.workspace_client.get_workspace_id()} " f"block Internet access?"
+            f"Does given workspace {self.workspace_client.get_workspace_id()} block Internet access?"
         )
+        wheelhouse: str | None = None
```
nfx, line 240: hm... should not be here, per se..

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -566,11 +569,30 @@ def _library_dep_order(library: str):
...
+        # archive, therefore, we upload the wheels separately.
+        remote_paths = []
+        with tempfile.TemporaryDirectory() as directory, zipfile.ZipFile(path, "r") as zip_ref:
```
nfx, line 582: what about `requirements.txt`?

src/databricks/labs/ucx/install.py
JCZuurmond: yes, this is something to be decided. As mentioned in [this comment](https://github.com/databrickslabs/ucx/pull/2380#pullrequestreview-2225047462), we can also ship the wheelhouse with the ucx install, then fetch it from there

src/databricks/labs/ucx/installer/workflows.py
JCZuurmond: Not required in the workflow, but we can add it

JCZuurmond:
Not merging the spike

PR #2381, opened 2024-08-07
Avoid false positives when linting for pyspark patterns

ericvergnaud:
## Changes
Check originating module when linting for pyspark calls

### Linked issues
None

### Functionality
None

### Tests
- [x] ran unit tests

### DIFF
--- a/tests/unit/source_code/linters/test_pyspark.py
+++ b/tests/unit/source_code/linters/test_pyspark.py
@@ -266,7 +266,8 @@ for i in range(10):
         ),
         # Test for option function
         (
-            """(df.write
+            """df=spark.sql('SELECT * FROm dual')
+(df.write
   .format("parquet")
   .option("path", "s3a://your_bucket_name/your_directory/")
   .option("spark.hadoop.fs.s3a.access.key", "your_access_key")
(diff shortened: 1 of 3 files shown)


tests/unit/source_code/linters/test_pyspark.py
```diff
@@ -277,9 +278,9 @@ def test_spark_sql_fix(migration_index):
                     code='direct-filesystem-access',
                     message='The use of direct filesystem references is deprecated: '
                     "s3a://your_bucket_name/your_directory/",
-                    start_line=0,
+                    start_line=1,
                     start_col=1,
-                    end_line=2,
+                    end_line=3,
                     end_col=59,
```
asnare, line 284: Not a comment on this PR per se, but more a general observation: should't these warnings refer to the problematic argument?

In this case the positioning information should really be something like: 3:19 thru 3:59

Humans can understand the current positioning information, even if more accurate highlighting is possible. But will this start to cause problems when it comes to automatic rewriting/fixing?

ericvergnaud: I agree with you, and it's been done with some linters, but let's have field engineers opine on this?

asnare: Yup; this was intended as the start of a conversation. 👍

nfx (approved):
nice

PR #2382, opened 2024-08-07
Fix raise logs `ResourceDoesNotExists` when iterating the log paths

JCZuurmond:
## Changes
Fix raise logs `ResourceDoesNotExists` when iterating the log paths. The test did not mimic the actual behavior, these changes do

### Tests

- [ ] added unit tests

### DIFF
--- a/src/databricks/labs/ucx/installer/workflows.py
+++ b/src/databricks/labs/ucx/installer/workflows.py
@@ -255,7 +255,7 @@ class DeployedWorkflows:
         """
         log_path = f"{self._install_state.install_folder()}/logs/{workflow}"
         try:
-            log_path_objects = self._ws.workspace.list(log_path)
+            log_path_objects = list(self._ws.workspace.list(log_path))
         except ResourceDoesNotExist:
             logger.warning(f"Can not fetch logs as folder {log_path} does not exist")
             return []
--- a/tests/unit/installer/test_workflows.py
+++ b/tests/unit/installer/test_workflows.py
@@ -13,10 +13,15 @@ from databricks.labs.ucx.framework.tasks import Task
 from databricks.labs.ucx.installer.workflows import DeployedWorkflows, WorkflowsDeployment
 
 
+class ResourceDoesNotExistIter:
+    def __iter__(self):
+        raise ResourceDoesNotExist("logs")
+
+
 def test_deployed_workflows_handles_log_folder_does_not_exists(mock_installation):
     ws = create_autospec(WorkspaceClient)
     ws.jobs.list_runs.return_value = [BaseRun(run_id=456)]
-    ws.workspace.list.side_effect = ResourceDoesNotExist("logs")
+    ws.workspace.list.return_value = ResourceDoesNotExistIter()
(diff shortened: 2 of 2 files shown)


src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -255,7 +255,7 @@ def _get_log_run_folders(self, workflow: str, run_id: str) -> list[str]:
         """
         log_path = f"{self._install_state.install_folder()}/logs/{workflow}"
         try:
-            log_path_objects = self._ws.workspace.list(log_path)
+            log_path_objects = list(self._ws.workspace.list(log_path))
```
asnare, line 258:
```suggestion
            # Ensure any exception is triggered early.
            log_path_objects = list(self._ws.workspace.list(log_path))
```

tests/unit/installer/test_workflows.py
```diff
@@ -13,10 +13,15 @@
...
+class ResourceDoesNotExistIter:
+    def __iter__(self):
+        raise ResourceDoesNotExist("logs")
+
+
 def test_deployed_workflows_handles_log_folder_does_not_exists(mock_installation):
     ws = create_autospec(WorkspaceClient)
     ws.jobs.list_runs.return_value = [BaseRun(run_id=456)]
-    ws.workspace.list.side_effect = ResourceDoesNotExist("logs")
+    ws.workspace.list.return_value = ResourceDoesNotExistIter()
```
ericvergnaud, line 24: wouldn't a lambda do the job ?

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -255,7 +255,7 @@ def _get_log_run_folders(self, workflow: str, run_id: str) -> list[str]:
...
         except ResourceDoesNotExist:
             logger.warning(f"Can not fetch logs as folder {log_path} does not exist")
```
ericvergnaud, line 261: Fix typo ? "Can not" -> "Cannot"

ericvergnaud (changes requested):

tests/unit/installer/test_workflows.py
asnare, line 24: This change shouldn't be necessary?
> **`side_effect`**<br/>
> This can either be a function to be called when the mock is called, an iterable **or an exception (class or instance) to be raised**.

Ref: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect

JCZuurmond: I think it does

asnare: Ahh. This is quite subtle:

```suggestion
    # Raise the error when the result is iterated over, NOT when the method is called.
    ws.workspace.list.return_value = ResourceDoesNotExistIter()
```

asnare (approved):
Outstanding catch, that the unit test wasn't mocking the real code accurately.

ericvergnaud (approved):

nfx (approved):
lgtm

PR #2388, opened 2024-08-07
Integration tests: purge suffixes and `RemoveAfter` tags for fixtures that don't already have this

asnare:
## Changes

This PR updates all remaining integration test fixtures so that where possible any resources they create have a `RemoveAfter` tag or the equivalent suffix on their name. (This prevents the resource from being removed from the test environment while the test is underway.)

Doing this in a single PR is intended to reduce future effort tracking down spurious test failures when this occurs.

The main remaining resource that is often deleted while integration tests are underway is the folder for the UCX installation itself. Unfortunately the name of this folder is fixed and tightly coupled to the generated product-name being used for the test. Dealing with this is non-trivial and has been omitted from this PR.

### Linked issues

Supersedes #2369.

### Tests

- integration tests


src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -970,7 +976,7 @@ def inventory_schema(make_schema):
 @pytest.fixture
 def make_catalog(ws, sql_backend, make_random) -> Generator[Callable[..., CatalogInfo], None, None]:
     def create() -> CatalogInfo:
-        name = f"ucx_C{make_random(4)}".lower()
+        name = f"ucx_c{make_random(4).lower()}_{get_purge_suffix()}"
```
nfx, line 979: done via properties. purge suffix is irrelevant here

src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -1047,7 +1053,7 @@ def create(  # pylint: disable=too-many-locals,too-many-arguments,too-many-state
         elif non_delta:
             table_type = TableType.EXTERNAL  # pylint: disable=redefined-variable-type
             data_source_format = DataSourceFormat.JSON
-            storage_location = f"dbfs:/tmp/ucx_test_{make_random(4)}"
+            storage_location = f"dbfs:/tmp/ucx_test_{make_random(4)}_{get_purge_suffix()}"
```
nfx, line 1056: we don't purge dbfs, just in case...

src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -1156,7 +1163,7 @@ def create(
             schema_name = schema.name
 
         if name is None:
-            name = f"ucx_T{make_random(4)}".lower()
+            name = f"ucx_t{make_random(4).lower()}_{get_purge_suffix()}"
```
nfx, line 1166: done via properties. purge suffix is irrelevant here

src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -1328,7 +1336,7 @@ def make_mounted_location(make_random, make_dbfs_data_copy, env_or_skip):
         location; the mounted location is made with fixture setup already.
     """
     existing_mounted_location = f'dbfs:/mnt/{env_or_skip("TEST_MOUNT_NAME")}/a/b/c'
-    new_mounted_location = f'dbfs:/mnt/{env_or_skip("TEST_MOUNT_NAME")}/a/b/{make_random(4)}'
+    new_mounted_location = f'dbfs:/mnt/{env_or_skip("TEST_MOUNT_NAME")}/a/b/{make_random(4)}-{get_purge_suffix()}'
```
nfx, line 1339: and we don't purge dbfs either..

src/databricks/labs/ucx/mixins/fixtures.py
asnare: I've updated this to set a `RemoveAfter` tag on the catalog instead.

src/databricks/labs/ucx/mixins/fixtures.py
asnare: Thanks; noted.

src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -967,6 +973,7 @@ def make_catalog(ws, sql_backend, make_random) -> Generator[Callable[..., Catalo
     def create() -> CatalogInfo:
         name = f"ucx_C{make_random(4)}".lower()
         sql_backend.execute(f"CREATE CATALOG {name}")
+        sql_backend.execute(f"ALTER CATALOG {name} SET TAGS ('RemoveAfter' = '{get_test_purge_time()}')'")
```
nfx, line 976: afaik, we cannot retrieve tags via rest api - see https://github.com/databrickslabs/watchdog/blob/main/scan/objects/uc.go#L68-L76

nfx (changes requested):

src/databricks/labs/ucx/mixins/fixtures.py
asnare: Noted, thanks.

src/databricks/labs/ucx/mixins/fixtures.py
asnare: This was a bit of a head-scratcher, mostly because via the SQL interface doesn't allow these to be modified and I wasn't sure it was worth switching to the REST API.

That said, after poking around in the Watchdog a bit, I don't think functions are explicitly being scanned/nuked? Instead it looks like they're bycatch from the schema scanner?

src/databricks/labs/ucx/mixins/fixtures.py
asnare: As noted elsewhere, I've dropped the tags as well and added a note in the code that catalogs can't currently be protected from purging by the watchdog.

nfx (approved):
lgtm

PR #2390, opened 2024-08-08
Optimisation: when detecting if a file is a notebook only read the start instead of the whole file

asnare:
## Changes

This PR updates the detection during linting of whether a file is a notebook or not for non-Workspace paths so that when performing the header-check we only read the start of the file (enough for the header) instead of loading it all. This is an optimisation rather than an issue of correctness.

### Tests

- existing unit tests


nfx (approved):
lgtm

PR #2392, opened 2024-08-08
Crawlers: Use `TRUNCATE TABLE` instead of `DELETE FROM` when resetting crawler tables.

asnare:
## Changes

This PR updates the `.reset()` implementation for crawlers to use `TRUNCATE TABLE` instead of `DELETE FROM`. This is slightly more efficient and idiomatic (but not an issue of correctness).

### Functionality

- [X] modified existing workflow: `migrate-data-reconciliation`

### Tests

- [X] updated unit tests
- [X] covered by integration tests

### DIFF
--- a/src/databricks/labs/ucx/framework/crawlers.py
+++ b/src/databricks/labs/ucx/framework/crawlers.py
@@ -52,7 +52,7 @@ class CrawlerBase(Generic[Result]):
         Delete the content of the inventory table.
         The next call to `snapshot` will re-populate the table.
         """
-        self._exec(f"DELETE FROM {self.full_name}")
+        self._exec(f"TRUNCATE TABLE {self.full_name}")
 
     @staticmethod
     def _valid(name: str) -> str:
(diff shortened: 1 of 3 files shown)


ericvergnaud:
You'd expect the SQL engine to automatically convert `DELETE FROM` without `WHERE` clause to have the same behavior as `TRUNCATE`. Maybe worth signaling this to engineering ?

asnare:
> You'd expect the SQL engine to automatically convert `DELETE FROM` without `WHERE` clause to have the same behavior as `TRUNCATE`. Maybe worth signaling this to engineering ?

I agree, and before making this change I did check the query plans for both versions to check they weren't being rewritten to the same thing: they are not.

It's possible that the Delta internals optimise `DELETE FROM` beyond the visibility of the Spark engine: I didn't check this.

I vaguely recall in the early days of Delta that there might have been some subtle differences between the two that might prevent the optimisation, maybe related to the semantics of opportunistic locking. (SQL purists sometimes argue about trigger semantics, but that doesn't apply here.)

Either way, I think that `TRUNCATE` is more idiomatic and without needing to look into further into details we know it's not doing the inefficient thing.

HariGS-DB:
Truncate table is not applicable for external table, from our code its only used to reset migration_status and recon_results which are all managed table so it should be ok I feel

nfx (approved):
lgtm

Issue #2074, opened 2024-07-03
[EPIC]: Track migration process and display it in a dashboard


nfx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

There's no easy way to know what still needs to be migrated to UC within a given workspace

### Proposed Solution

- [x] #2578
- [x] #2576
   - [x] #2575
- [x] #2574
- [x] #2573
  - [x] #2571
  - [x] #2572
  - [x] #2600
  - [ ] #2603
- [x] #2577
- [x] #2816
- [x] #2594
   - [ ] Pre-requisite: [Change in watchdog](https://github.com/databrickslabs/watchdog/issues/54)
- [x] #2595
- [x] #2596
  - [ ] #3045
- [x] #3064
   - [x] #3057 
   - [x] #3058
   - [x] #3059
   - [x] #3060
   - [x] #3061
   - [x] #3062
   - [x] #3063
- Dashboards
  - [ ] #3366
  - [ ] #3367
  - [ ] #3368
  - [ ] #3369

## Potential challenges
- [x] #2597
- [ ] ~but for migration progress purposes, we need to overwrite the tables. or add another column with a timestamp and modify "fetch latest" queries to fetch the latest timestamp of the snapshot. fetching the latest timestamp from the snapshot allows to build a bar-chart widget to see how fast migration progresses, but we don't really care about it it. if we do fetch-latest-timestamp, all our views and dashboards would become a bit more complicated. but that's fine.~ --> Decision is made to keep the current ucx inventory and store history in a separate table (see proposed solution above)
- [ ] ~let's keep the status of migration progress in HMS (for now), but we can change this decision in a few weeks.~ --> Decision is made to store the migration process in a ucx catalog.

## Migration process crawlers

Assessment tasks that make sense to re-run on `migration-progress` workflow:
- `crawl_tables`
- `assess_jobs` - potentially harden the code there as well
- `assess_clusters` - potentially harden as well
- `assess_pipelines` - potentially harden as well
- `crawl_cluster_policies`
- `assess_global_init_scripts`

not to be re-run:
- `crawl_mounts` - we already pre-created external locations
- `setup_tacl` - we don't need to crawl grants
- `crawl_grants` - no need to, i think
-  `estimate_table_size_for_migration` - most likely not necessary
- `guess_external_locations` - we already migrated external locations by this point
- `assess_incompatible_submit_runs `- not going to be necessary in september
[...]

PR #2402, opened 2024-08-09
Refactor: `CrawlerBase` and `PermissionsManager` snapshotting pattern

asnare:
## Changes

This PR:

 - Refactors the existing permissions crawler so that it uses the same fetch/crawl/snapshot pattern as the rest of the crawlers.
 - Removes the permissions crawler's bespoke `.cleanup()` implementation in favour of the `.reset()` method already available on the base class.
 - Some dead code was removed on the permissions crawler: `.load_for_all()`
 - Specifies `.snapshot()` as an interface (on `CrawlerBase`) that all crawlers need to support.

### Linked issues

Progresses #2074, by laying the ground for crawlers to support refreshing.

Note that this PR and #2392 have a behaviour conflict that will trigger a failure in `test_runtime_crawl_permissions` if either is merged: resolving this is trivial.

### Functionality

- modified existing workflow: `assessment`

### Tests

- [X] updated unit tests
- [X] updated integration tests


src/databricks/labs/ucx/workspace_access/groups.py
```diff
@@ -431,11 +432,9 @@ def __init__(  # pylint: disable=too-many-arguments
         self._external_id_match = external_id_match
         self._verify_timeout = verify_timeout
 
-    def snapshot(self) -> list[MigratedGroup]:
-        return self._snapshot(self._fetcher, self._crawler)
-
     def has_groups(self) -> bool:
-        return len(self.snapshot()) > 0
+        first_element = islice(self.snapshot(), 1)
```
nfx, line 436: can you revert the change on this method? i find it risky. it's fine to load snapshot twice, as it's materialized

nfx (changes requested):
one tiny change

src/databricks/labs/ucx/workspace_access/groups.py
asnare: I've reverted it.

This was intended to be a poor-man's `limit 1` because `.snapshot()` is normally lazy, etc.

asnare: Pfft. And I just noticed the method isn't actually called from anywhere (including tests). So I'm just going to nuke it entirely.

asnare: … and removed.

nfx (approved):
lgtm

Issue #2405, opened 2024-08-09
[BUG]: `create-uber-principal` command failed when updating the SQL warehouse config


qziyuan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

1. `create-uber-principal` will fail with `ERROR [d.l.ucx.create-uber-principal] InvalidParameterValue: Endpoint security policy is required and must be one of NONE, DATA_ACCESS_CONTROL, PASSTHROUGH.` when updating the SQL warehouse config to add the uber principal to data_access_config.
This can be fixed by adding `security_policy` to the [set_workspace_warehouse_config api call](https://github.com/databrickslabs/ucx/blob/642ececafb8ec15cb91bc1466d8eb212521461ef/src/databricks/labs/ucx/azure/access.py#L243-L246)

2. Furthermore, the same api call is still failing, after above is fixed, with this error: `databricks.sdk.errors.platform.InvalidParameterValue: enable_serverless_compute is required.`. 
For a workspace with serverless enabled, the undocumented `enable_serverless_compute` field must be provided in the api request body, [but python SDK does not support this field](https://github.com/databricks/databricks-sdk-py/issues/305). 
UCX may have to call the [REST API](https://docs.databricks.com/api/azure/workspace/warehouses/setworkspacewarehouseconfig#enabled_warehouse_types) directly instead of using python SDK, if the SDK cannot support this field in the near future.

### Cloud

Azure

### Operating System

macOS

### Version

latest via Databricks CLI

qziyuan:
PR #2049 will fix 1st `Endpoint security policy is required and must be one of NONE, DATA_ACCESS_CONTROL, PASSTHROUGH.` error.
The 2nd `enable_serverless_compute is required.` error is pending upstream issue https://github.com/databricks/databricks-sdk-py/issues/305. PR #2049 will provide a workaround by printing the data access config in logs which can be then manually applied into the admin UI.

PR #2409, opened 2024-08-13
Fix missing `security_policy` when updating SQL warehouse config

qziyuan:
1. Fix `InvalidParameterValue: Endpoint security policy is required and must be one of NONE, DATA_ACCESS_CONTROL, PASSTHROUGH.` error in #2405 
2. Print new SQL warehouse data access config in the log when `databricks.sdk.errors.platform.InvalidParameterValue: enable_serverless_compute is required.` error in #2405 occurs. With that users can manually copy paste the config in UI to finish setting up the uber principal,

### DIFF
--- a/src/databricks/labs/ucx/aws/access.py
+++ b/src/databricks/labs/ucx/aws/access.py
@@ -12,6 +12,7 @@ from databricks.sdk import WorkspaceClient
 from databricks.sdk.errors import NotFound, ResourceDoesNotExist
 from databricks.sdk.service.catalog import Privilege
 from databricks.sdk.service.compute import Policy
+from databricks.sdk.service.sql import SetWorkspaceWarehouseConfigRequestSecurityPolicy
 
 from databricks.labs.ucx.assessment.aws import (
     AWSInstanceProfile,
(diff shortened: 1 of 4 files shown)


src/databricks/labs/ucx/aws/access.py
```diff
@@ -266,10 +267,15 @@ def _update_sql_dac_with_instance_profile(self, iam_instance_profile: AWSInstanc
                 f"workspace warehouse config. Do you want UCX to to update it with the uber instance profile?"
             ):
                 return
+        if warehouse_config.security_policy:
+            security_policy = SetWorkspaceWarehouseConfigRequestSecurityPolicy(warehouse_config.security_policy.value)
+        else:
```
nfx, line 272:
```suggestion
        if not warehouse_config.security_policy:
```
can it be simplified?

src/databricks/labs/ucx/azure/access.py
```diff
@@ -240,10 +240,24 @@ def _update_sql_dac_with_spn(
                     ),
                 ]
             )
-        self._ws.warehouses.set_workspace_warehouse_config(
-            data_access_config=sql_dac,
-            sql_configuration_parameters=warehouse_config.sql_configuration_parameters,
-        )
+        if warehouse_config.security_policy:
+            security_policy = SetWorkspaceWarehouseConfigRequestSecurityPolicy(warehouse_config.security_policy.value)
```
nfx, line 244: simplify

nfx (changes requested):

nfx (approved):
lgtm

PR #2421, opened 2024-08-14
Add support for DBFS root location resolution for HMS federation

vihangk-db:
This PR adds a DBFS Resolver which is used by HMS federation to resolve the DBFS root locations.

TODO:

- [ ] rewrite `ExternalLocations` to fetch from `mounts` table bypassing the `Mounts` crawler
- [ ] resolve special `/user/hive/metastore` as special case of DBFS mount
- [ ] use `LocationTrie` to come guess locations

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/federation.py
+++ b/src/databricks/labs/ucx/hive_metastore/federation.py
@@ -6,6 +6,8 @@ class DbfsRootResolver:
 
     def run(self) -> str:
         code = """
+// Warning: relying on an undocumented API here. This is brittle, as it may break.
+// TODO: switch to a publicly documented API when available.
 URI = spark._jvm.java.net.URI
 Some = spark._jvm.scala.Some
 HmsFedDbfsUtils = spark._jvm.com.databricks.sql.managedcatalog.connections.HmsFedDbfsUtils
--- a/src/databricks/labs/ucx/hive_metastore/federation.py
+++ b/src/databricks/labs/ucx/hive_metastore/federation.py
@@ -1,0 +1,22 @@
+from databricks.labs.blueprint.commands import CommandExecutor
+
+class DbfsRootResolver:
+    def __init__(self, command_executor: CommandExecutor):
+        self._command_executor = command_executor
+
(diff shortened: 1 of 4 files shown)


src/databricks/labs/ucx/hive_metastore/federation.py
```diff
@@ -0,0 +1,24 @@
+from databricks.labs.blueprint.commands import CommandExecutor
```
FastLee, line 1: Do we need a running cluster for this resolution?
Should we use it as part of the assessment?

vihangk-db: Yes a single user cluster

tests/integration/hive_metastore/test_federation.py
```diff
@@ -0,0 +1,10 @@
+from databricks.labs.blueprint.commands import CommandExecutor
+from databricks.labs.ucx.hive_metastore.federation import DbfsRootResolver
+
+
+def test_dbfs_root_resolver(ws, env_or_skip):
+    command_executor = CommandExecutor(ws.clusters, ws.command_execution, lambda: env_or_skip("DATABRICKS_CLUSTER_ID"))
+    root_resolver = DbfsRootResolver(command_executor)
```
nfx, line 7: add a unit test in the `tests/unit/hive_metastore` folder
```
command_executor_mock = create_autospec(CommandExecutor)
command_executor_mock.run.return_value = 's3://...'
```

example of a unit test:
https://github.com/databrickslabs/ucx/blob/main/tests/unit/hive_metastore/test_locations.py#L98-L114

nfx (commented):
please add a unit test and let's get it merged

tests/integration/hive_metastore/test_federation.py
vihangk-db: Sorry for the late response. I am not sure what the above unit test will actually be testing if the command executor itself a mock. I added a unit test as per your suggestion above. Let me know if you had something else in mind.

nfx:
closing in favor of #2947

PR #2454, opened 2024-08-19
Fallback mode enabled

vihangk-db:
### Linked issues

Resolves #..

### Functionality

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/labs.yml
+++ b/labs.yml
@@ -244,3 +244,6 @@ commands:
         description: workspace_ids which should join a collection. provide a comma separated list of workspace ids.
       - name: target-workspace-id
         description: (Optional) id of a workspace in the target collection. If not specified, ucx will prompt to select from a list
+
+  - name: create-federated-catalog
+    description: (EXPERIMENTAL) Create a federated catalog in the workspace
--- a/labs.yml
+++ b/labs.yml
@@ -247,3 +247,6 @@ commands:
 
   - name: create-federated-catalog
     description: (EXPERIMENTAL) Create a federated catalog in the workspace
+
+  - name: enable-hms-federation
+    description: (EXPERIMENTAL) Enable HMS federation based migration flow. When this is enabled, UCX will create a federated HMS catalog which syncs from the workspace HMS.
--- a/src/databricks/labs/ucx/azure/locations.py
+++ b/src/databricks/labs/ucx/azure/locations.py
@@ -20,12 +20,14 @@ class ExternalLocationsMigration:
         resource_permissions: AzureResourcePermissions,
         azurerm: AzureResources,
         principal_acl: PrincipalACL,
+        enable_hms_federation: bool
     ):
         self._ws = ws
         self._hms_locations = hms_locations
         self._resource_permissions = resource_permissions
         self._azurerm = azurerm
         self._principal_acl = principal_acl
+        self._enable_hms_federation = enable_hms_federation
 
     def _app_id_credential_name_mapping(self) -> tuple[dict[str, str], dict[str, str]]:
         # list all storage credentials.
@@ -120,7 +122,8 @@ class ExternalLocationsMigration:
     ) -> str | None:
         try:
             self._ws.external_locations.create(
-                name, url, credential, comment=comment, read_only=read_only, skip_validation=skip_validation
+                name, url, credential, comment=comment, read_only=read_only, skip_validation=skip_validation,
+                fallback=self._enable_hms_federation
             )
             return url
         except InvalidParameterValue as invalid:
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -502,5 +502,12 @@ def lint_local_code(
     ctx.local_code_linter.lint(prompts, None if path is None else Path(path))
 
 
+@ucx.command
+def create_federated_catalog(w: WorkspaceClient, prompts: Prompts):
+    """(Experimental) Create federated catalog from current workspace Hive Metastore."""
+    ctx = WorkspaceContext(w)
+    ctx.federation.run()
+
+
 if __name__ == "__main__":
     ucx()
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -508,6 +508,13 @@ def create_federated_catalog(w: WorkspaceClient, prompts: Prompts):
     ctx = WorkspaceContext(w)
     ctx.federation.run()
 
+@ucx.command
+def enable_hms_federation(w: WorkspaceClient, prompts: Prompts, ctx: WorkspaceContext | None = None):
+    """(Experimental) Create federated catalog from current workspace Hive Metastore."""
+    if not ctx:
+        ctx = WorkspaceContext(w)
+    ctx.federation_enabler.enable()
+
 
 if __name__ == "__main__":
     ucx()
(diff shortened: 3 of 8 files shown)


src/databricks/labs/ucx/hive_metastore/federation.py
```diff
@@ -0,0 +1,112 @@
...
+            return self._workspace_client.connections.create(
+                name=name,
+                connection_type=ConnectionType.HIVE_METASTORE,  # needs SDK change
+                options={"builtin": "true"},
+            )
+        except AlreadyExists:
+            for connection in self._workspace_client.connections.list():
+                if connection.name == name:
+                    return connection
+        raise NotFound(f'Connection {name} not found')
```
JCZuurmond, line 70: Make clear that should be raised when the for loop finishes with the except block
```suggestion
            else:
                raise NotFound(f'Connection {name} not found')
```

src/databricks/labs/ucx/hive_metastore/federation.py
```diff
@@ -0,0 +1,112 @@
...
+    def _get_authorized_paths(self) -> str:
+        existing = {}
+        for external_location in self._workspace_client.external_locations.list():
+            existing[external_location.url] = external_location
+        authorized_paths = []
+        current_user = self._workspace_client.current_user.me()
+        if not current_user.user_name:
+            raise ValueError('Current user not found')
+        for external_location_info in self._external_locations.snapshot():
+            location = external_location_info.location.rstrip('/').replace('s3a://', 's3://')
```
JCZuurmond, line 81: How do we know which prefixes need replacing? Add a comment for future understanding

src/databricks/labs/ucx/hive_metastore/federation.py
```diff
@@ -0,0 +1,112 @@
...
+                connection_type=ConnectionType.HIVE_METASTORE,  # needs SDK change
+                options={"builtin": "true"},
+            )
+        except AlreadyExists:
+            for connection in self._workspace_client.connections.list():
+                if connection.name == name:
+                    return connection
+        raise NotFound(f'Connection {name} not found')
+
+    def _get_authorized_paths(self) -> str:
```
JCZuurmond, line 72: Plural "paths" usually means a collection type

```suggestion
    def _get_authorized_paths(self) -> list[str]:
```

JCZuurmond: Note that this method does more than just getting the authorized paths, it also sets the missing permissions if needed, update the name accordingly or separate out the two functionalities:
1. Getting paths
2. Setting missing permissions

src/databricks/labs/ucx/cli.py
```diff
@@ -502,5 +502,19 @@ def lint_local_code(
...
 
+@ucx.command
+def create_federated_catalog(w: WorkspaceClient, prompts: Prompts):
+    """(Experimental) Create federated catalog from current workspace Hive Metastore."""
+    ctx = WorkspaceContext(w)
+    ctx.federation.run()
+
+@ucx.command
+def enable_hms_federation(w: WorkspaceClient, prompts: Prompts, ctx: WorkspaceContext | None = None):
+    """(Experimental) Create federated catalog from current workspace Hive Metastore."""
```
JCZuurmond, line 513: This command does not create anything it enables the flag in the settings to create the federated catalog

JCZuurmond: Why is this command needed? Can it be merged with the `create_federated_catalog` instead?

labs.yml
```diff
@@ -244,3 +244,9 @@ commands:
         description: workspace_ids which should join a collection. provide a comma separated list of workspace ids.
       - name: target-workspace-id
         description: (Optional) id of a workspace in the target collection. If not specified, ucx will prompt to select from a list
+
+  - name: create-federated-catalog
+    description: (EXPERIMENTAL) Create a federated catalog in the workspace
+
+  - name: enable-hms-federation
+    description: (EXPERIMENTAL) Enable HMS federation based migration flow. When this is enabled, UCX will create a federated HMS catalog which syncs from the workspace HMS.
```
JCZuurmond, line 252: When will ucx create the federated HMS catalog?

JCZuurmond (commented):
See comments

labs.yml
vihangk-db: Sorry for the late response. This PR should be on feat/hms-fed branch instead of main. I changed the base now.

The federated catalog is created [here](https://github.com/databrickslabs/ucx/blob/fcdc75a33990c98a2ec3dd1912feeca0a332835e/src/databricks/labs/ucx/hive_metastore/federation.py#L36) which is called from

`create_federated_catalog` [here](https://github.com/databrickslabs/ucx/blob/feat/hms-fed/src/databricks/labs/ucx/cli.py#L506)

src/databricks/labs/ucx/cli.py
vihangk-db: We need this command because the locations are created before the create_federated_catalog is executed and we want to have the fallback mode enable for the locations when they are created. This command can be independently used for create_federated_catalog in cases where the federated catalog is already created, and UCX needs to be used to create new locations post federation catalog creation.

src/databricks/labs/ucx/config.py
```diff
@@ -71,6 +71,9 @@ class WorkspaceConfig:  # pylint: disable=too-many-instance-attributes
     # [INTERNAL ONLY] Whether the assessment should capture only specific object permissions.
     include_object_permissions: list[str] | None = None
 
+    # [INTERNAL ONLY] When the HMS fedeation is enabled.
```
JCZuurmond, line 74:
```suggestion
    # [INTERNAL ONLY] When the HMS federation is enabled.
```

src/databricks/labs/ucx/contexts/workspace_cli.py
```diff
@@ -184,6 +186,10 @@ def notebook_loader(self) -> NotebookLoader:
     def federation(self):
         return HiveMetastoreFederation(self.workspace_client, self.external_locations, self.workspace_info)
 
+    @cached_property
+    def federation_enabler(self):
```
JCZuurmond, line 190:
```suggestion
    def federation_enabler(self) -> HiveMetastoreFederationEnabler:
```

src/databricks/labs/ucx/cli.py
```diff
@@ -508,6 +508,13 @@ def create_federated_catalog(w: WorkspaceClient, prompts: Prompts):
     ctx = WorkspaceContext(w)
     ctx.federation.run()
 
+@ucx.command
+def enable_hms_federation(w: WorkspaceClient, prompts: Prompts, ctx: WorkspaceContext | None = None):
```
JCZuurmond, line 512:
```suggestion
def enable_hms_federation(w: WorkspaceClient, prompts: Prompts, ctx: WorkspaceContext | None = None) -> None:
```

src/databricks/labs/ucx/azure/locations.py
```diff
@@ -120,7 +122,8 @@ def _create_external_location_helper(
     ) -> str | None:
         try:
             self._ws.external_locations.create(
-                name, url, credential, comment=comment, read_only=read_only, skip_validation=skip_validation
+                name, url, credential, comment=comment, read_only=read_only, skip_validation=skip_validation,
```
JCZuurmond, line 125: `make fmt`

JCZuurmond (commented):
The code looks good to me, but I am missing changes to the relevant documentation that explain the command being introduced. Also update the PR comment, I am missing the motivation for this PR

Issue #2336, opened 2024-07-31
[BUG]: `create-missing-principals` does not enforce role uniqueness on AWS


nfx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

currently we use `UCX-ROLE` as the default role name, but we do receive 
https://github.com/databrickslabs/ucx?tab=readme-ov-file#create-missing-pricipals-command-aws-only

### Expected Behavior

append metastore name as the ucx role suffix

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

nfx:
Do not use one predefined name for anything cloud-related - see the relevant fix for locations - https://github.com/databrickslabs/ucx/commit/505fe1e9d363245f122561263dfc8bc4ae8056f9

PR #2456, opened 2024-08-21
Fix issue when running create-missing-credential cmd tries to create the role again if already created

### LINKED ISSUE #2413: [BUG]: External location with S3A prefix create a conflict when attempting to create roles

### Current Behavior

If an external location with S3A bucket is discovered it is not matched properly for a role that is created in the create missing principals step.

### Expected Behavior

_No response_

### Steps To Reproduce

1. In sandbox create HMS tables with an S3A location reference.

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

HariGS-DB:
## Changes
When running create-missing-credential, code is trying to match locations with uc_roles_access.csv entries.
If a role is already created, running the command tries the create the role again. due to the line of code here
path = PurePath(external_location.location)
            matching_role = False
            for role in compatible_roles:
                if path.match(role.resource_path):

Ex:
path = PurePath(“s3://labs-things/a/b/”)
path.match(“s3://labs-things”)
the above code returns False, when it should be True due to which its trying to create the role again.

changed it to use location.startswith()

#2336 

Resolves #2413 

### Functionality

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/aws/access.py
+++ b/src/databricks/labs/ucx/aws/access.py
@@ -61,6 +61,8 @@ class AWSResourcePermissions:
         """
         roles: list[AWSUCRoleCandidate] = []
         missing_paths = self._identify_missing_paths()
+        if len(missing_paths) == 0:
+            return []
         s3_buckets = set()
         for missing_path in missing_paths:
             match = re.match(AWSResources.S3_BUCKET, missing_path)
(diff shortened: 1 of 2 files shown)


nfx (approved):
lgtm

PR #2479, opened 2024-08-22
draft

### LINKED ISSUE #2362: [BUG]:`migrate-credentials` commands have to cleanup/resumability
### Current Behavior

`migrate-credentials` commands creates residue roles/policies/etc.

### Expected Behavior

When the commands fails, we have to cleanup any residue created by those commands. 

### Cloud

AWS, Azure

HariGS-DB:
### Linked issues

Resolves #2362 

### Functionality

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/azure/access.py
+++ b/src/databricks/labs/ucx/azure/access.py
@@ -419,6 +419,49 @@ class AzureResourcePermissions:
             self._delete_uber_principal()  # Clean up dangling resources
             raise
 
+    def _delete_access_connectors(self) -> None:
+
+        def log_permission_denied(function: Callable[P, R], *, message: str) -> Callable[P, R | None]:
+            @wraps(function)
+            def wrapper(*args: Any, **kwargs: Any) -> R | None:
+                try:
+                    return function(*args, **kwargs)
+                except PermissionDenied:
+                    logger.error(message, exc_info=True)
+                    return None
+
+            return wrapper
+
+        message = "Missing permissions to load the configuration"
+        config = log_permission_denied(self._installation.load, message=message)(WorkspaceConfig)
+        if config is None or config.uber_spn_id is None:
+            return
+
+        secret_identifier = f"secrets/{config.inventory_database}/{self._UBER_PRINCIPAL_SECRET_KEY}"
+        storage_accounts = self._get_storage_accounts()
+
+        storage_account_ids = ' '.join(str(st.id) for st in storage_accounts)
+        message = f"Missing permissions to delete storage permissions for: {storage_account_ids}"
+        log_permission_denied(self._azurerm.delete_storage_permission, message=message)(
+            config.uber_spn_id, *storage_accounts, safe=True
+        )
+        message = f"Missing permissions to delete service principal: {config.uber_spn_id}"
+        log_permission_denied(self._azurerm.delete_service_principal, message=message)(config.uber_spn_id, safe=True)
+        if config.policy_id is not None:
+            message = "Missing permissions to revert cluster policy"
+            log_permission_denied(self._remove_service_principal_configuration_from_cluster_policy, message=message)(
+                config.policy_id, config.uber_spn_id, secret_identifier, storage_accounts
+            )
+        message = "Missing permissions to revert SQL warehouse config"
+        log_permission_denied(
+            self._remove_service_principal_configuration_from_workspace_warehouse_config, message=message
+        )(config.uber_spn_id, secret_identifier, storage_accounts)
+        message = "Missing permissions to delete secret scope"
+        log_permission_denied(self._safe_delete_scope, message=message)(config.inventory_database)
+        message = "Missing permissions to save the configuration"
+        config.uber_spn_id = None
+        log_permission_denied(self._installation.save, message=message)(config)
+
     def _delete_uber_principal(self) -> None:
 
         def log_permission_denied(function: Callable[P, R], *, message: str) -> Callable[P, R | None]:
@@ -512,6 +555,7 @@ class AzureResourcePermissions:
         thread_name = "Creating access connectors for storage accounts"
         results, errors = Threads.gather(thread_name, tasks)
         if len(errors) > 0:
+
             raise ManyError(errors)
         return list(results)
 


src/databricks/labs/ucx/azure/access.py
```diff
@@ -512,9 +512,20 @@ def create_access_connectors_for_storage_accounts(self) -> list[tuple[AccessConn
         thread_name = "Creating access connectors for storage accounts"
         results, errors = Threads.gather(thread_name, tasks)
         if len(errors) > 0:
+            logger.error("Error creating all access connectors. Deleting incomplete ones.")
```
JCZuurmond, line 515: The code does not delete the incomplete access connectors, it deletes the once which were completed

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -264,19 +265,26 @@ def _migrate_service_principals(
...
-            execution_result.append(validation_results)
+                storage_credential_info = self._storage_credential_manager.create_with_client_secret(spn)
+                validation_results = self._storage_credential_manager.validate(
+                    storage_credential_info,
+                    spn.permission_mapping.prefix,
+                )
+                execution_result.append(validation_results)
+            except (PermissionDenied, NotFound, BadRequest) as e:
+                logger.error("Error creating storage credentials")
+                self._resource_permissions.delete_storage_credentials(execution_result)
```
JCZuurmond, line 278: So, we require all SP to be migrated at once and do not allow for some to be migrated?

src/databricks/labs/ucx/azure/access.py
```diff
@@ -512,9 +512,20 @@ def create_access_connectors_for_storage_accounts(self) -> list[tuple[AccessConn
...
+            delete_access_connectors = [access_connector for access_connector, _ in results]
+            self.delete_access_connectors(delete_access_connectors)
             raise ManyError(errors)
         return list(results)
 
+    def delete_storage_credentials(self, storage_credentials: list[Any]):
```
JCZuurmond, line 521: Depending on your preference:

```suggestion
    def delete_storage_credential(self, *storage_credentials: Any):
```

Same below

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -288,19 +296,25 @@ def _create_storage_credentials_for_storage_accounts(self) -> list[StorageCreden
                 validation_results = self._storage_credential_manager.validate(storage_credential_info, url)
             except BadRequest:
                 logger.warning(f"Could not validate storage credential {storage_credential_info.name} for url {url}")
+            except (PermissionDenied, NotFound):
+                logger.error("Error creating all access connectors. Deleting incomplete ones.")
```
JCZuurmond, line 300: This is an error on validating the access connectors, not creating them

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -314,9 +328,14 @@ def run(self, prompts: Prompts, include_names: set[str] | None = None) -> list[S
         )
         sp_results = []
         if plan_confirmed:
-            sp_results = self._migrate_service_principals(sp_migration_infos)
+            try:
+                sp_results = self._migrate_service_principals(sp_migration_infos)
+            except (BadRequest, PermissionDenied, NotFound):
+                self._resource_permissions.delete_storage_credentials(execution_results)
```
JCZuurmond, line 334: Do we need to delete the access connectors if the SP migration fails? I would say that we prefer to stick with the access connectors in that case

JCZuurmond (commented):
See comments

src/databricks/labs/ucx/azure/credentials.py
HariGS-DB: In order to be a clean run, my suggestion is to either do all or roll back all. We don't want to leave it in in between state where it created 3 access connector but failed in the 4th one or it created all access connector but failed while creating storage credential for them. do you agree?

src/databricks/labs/ucx/azure/credentials.py
HariGS-DB: yes that makes sense. removing the code for access connectors roll back in this step

src/databricks/labs/ucx/azure/access.py
HariGS-DB: Done

src/databricks/labs/ucx/azure/credentials.py
HariGS-DB: Yes i have made the changes now.
 so all access connectors and associated storage credentials or roll back everything
migrate all service principal or rollback everything
but if access connectors and associated storage credential complete but migrate service principal fail, we only roll back migrate service principal ones

src/databricks/labs/ucx/azure/credentials.py
HariGS-DB: updated message

src/databricks/labs/ucx/azure/access.py
```diff
@@ -512,9 +512,23 @@ def create_access_connectors_for_storage_accounts(self) -> list[tuple[AccessConn
         thread_name = "Creating access connectors for storage accounts"
         results, errors = Threads.gather(thread_name, tasks)
         if len(errors) > 0:
+            logger.error(
+                "Error creating all access connectors. Please check the error msg and try again. "
+                "Rolling back access connectors which are created."
```
JCZuurmond, line 517:
```suggestion
                "Error creating access connectors. Please review the error message and resolve the issue before trying again. "
                f"Removing successfully created access connectors."
```

src/databricks/labs/ucx/azure/access.py
```diff
@@ -512,9 +512,23 @@ def create_access_connectors_for_storage_accounts(self) -> list[tuple[AccessConn
         thread_name = "Creating access connectors for storage accounts"
         results, errors = Threads.gather(thread_name, tasks)
```
JCZuurmond, line 513: What does the `.gather` return for result when there is an error?

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -264,19 +264,29 @@ def _migrate_service_principals(
...
-            execution_result.append(validation_results)
+                storage_credential_info = self._storage_credential_manager.create_with_client_secret(spn)
+                validation_results = self._storage_credential_manager.validate(
+                    storage_credential_info,
+                    spn.permission_mapping.prefix,
+                )
+                execution_result.append(validation_results)
+            except (PermissionDenied, NotFound, BadRequest) as e:
+                logger.error(
+                    "Error creating storage credentials. Please check error message and try again."
```
JCZuurmond, line 277: Similar update as above

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -288,6 +298,14 @@ def _create_storage_credentials_for_storage_accounts(self) -> list[StorageCreden
                 validation_results = self._storage_credential_manager.validate(storage_credential_info, url)
             except BadRequest:
                 logger.warning(f"Could not validate storage credential {storage_credential_info.name} for url {url}")
+            except (PermissionDenied, NotFound):
+                logger.error(
+                    "Error validating all storage credentials. Please check the error message and tr again."
```
JCZuurmond, line 303: Similar update as above

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -298,9 +316,9 @@ def run(self, prompts: Prompts, include_names: set[str] | None = None) -> list[S
             "[RECOMMENDED] Please confirm to create an access connector with a managed identity for each storage "
             "account."
         )
-        ac_results = []
+        execution_results: list[StorageCredentialValidationResult] = []
```
JCZuurmond, line 319: Rename is not consistent:
- sp_results -> service principal results
- ac_results -> access connector results
- execution_results = sp_results + ac_results

src/databricks/labs/ucx/azure/credentials.py
```diff
@@ -288,6 +298,14 @@ def _create_storage_credentials_for_storage_accounts(self) -> list[StorageCreden
                 validation_results = self._storage_credential_manager.validate(storage_credential_info, url)
             except BadRequest:
                 logger.warning(f"Could not validate storage credential {storage_credential_info.name} for url {url}")
+            except (PermissionDenied, NotFound):
+                logger.error(
+                    "Error validating all storage credentials. Please check the error message and tr again."
+                    "Rolling back access connectors and storage credentials created."
+                )
+                self._resource_permissions.delete_storage_credential(execution_results)
+                delete_access_connectors = [access_connector for access_connector, _ in access_connectors]
```
JCZuurmond, line 307: should be the urls right?
```suggestion
                access_connector_urls = [url for _, url in access_connectors]
```

JCZuurmond (commented):
Have some nits only now

src/databricks/labs/ucx/azure/access.py
HariGS-DB: This would return the output of _create_access_connector_for_storage_account function that is invoked from the thread so list of tuple[AccessConnector, str]

src/databricks/labs/ucx/azure/credentials.py
HariGS-DB: validation_results returns StorageCredentialValidationResult which contains the storage credential name and I use that to delete the storage credential on error

src/databricks/labs/ucx/azure/credentials.py
HariGS-DB: access_connectors returns list[tuple[AccessConnector, str]] and i use the AccessConnector data class to get the resource id of the access connector which is passed to delete the access connector

src/databricks/labs/ucx/azure/credentials.py
HariGS-DB: the url is for the storage account which could be like abfss://container1@storageaccount.dfs.windows.net
access_connector would be AccessConnector data class
which will contain values like /subscriptions/subscriptionID/resourceGroups/RGName/providers/Microsoft.Databricks/accessConnectors/connection_name

src/databricks/labs/ucx/azure/credentials.py
HariGS-DB: done

nfx (approved):

PR #2484, opened 2024-08-26
Add a check for No isolation shared clusters and MLR

### LINKED ISSUE #846: [FEATURE]: cluster assessment should assess MLR with shared mode cluster

### Problem statement

Right now the cluster assessment treat MLR as supported. However, if the customer has MLR cluster that are being shared by multiple users (HC cluster without table ACL or standard cluster), this cluster cannot be simply upgraded to shared mode cluster, since MLR is not supported in shared mode cluster.
Further discussion is required for upgrading this type of cluster:
- Is service principal cluster acceptable?
- Do users use unsupported ML libs? Can they use normal DBR and install required and supported ML libs as cluster or session level libraries?
- Can users separate their training code to a single user mode cluster with MLR and run non ML code in shared mode cluster? 

### Proposed Solution

1. At least we should call out all MLR clusters, and inform the customer that further discussion maybe required.
2. Furthermore, we can use similar solution in #810 to leverage the audit system table to verify if a MLR cluster is really being shared by multiple users. If shared, then we call out that upgrade this cluster need further discussion.

### Additional Context

_No response_

gubyb:
Resolves #846 

## Changes
Added check for No isolation shared clusters using MLR
image

### Functionality

- [ ] added new CLI command
- [ ] added a new workflow
- [x] modified existing workflow: Assessment workflow, cluster crawler
- [ ] added a new table

### Tests
- [x] manually tested
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/assessment/clusters.py
+++ b/src/databricks/labs/ucx/assessment/clusters.py
@@ -25,6 +25,7 @@ from databricks.labs.ucx.assessment.crawlers import (
     INIT_SCRIPT_LOCAL_PATH,
     azure_sp_conf_present_check,
     spark_version_compatibility,
+    is_mlr,
 )
 from databricks.labs.ucx.assessment.init_scripts import CheckInitScriptMixin
 from databricks.labs.ucx.framework.crawlers import CrawlerBase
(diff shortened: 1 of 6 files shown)


nfx (approved):
lgtm

PR #2485, opened 2024-08-26
Always escape SQL identifiers

### LINKED ISSUE #2483: [BUG]: `SHOW GRANTS ON TABLE history` fails

### Current Behavior

`SHOW GRANTS ON TABLE history` fails with : `[[PARSE_SYNTAX_ERROR](https://docs.microsoft.com/azure/databricks/error-messages/error-classes#parse_syntax_error)] Syntax error at or near 'hive_metastore': missing 'SHARE'. SQLSTATE: 42601`

### Expected Behavior

Should not fail

### Steps To Reproduce

1. Create a table called `history`
2. Trigger the show grants

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

JCZuurmond:
## Changes
Always escape SQL identifiers

### Linked issues

Resolves #2483

- [x] updated unit tests


src/databricks/labs/ucx/framework/utils.py
```diff
@@ -7,25 +7,18 @@
 _allowed_object_chars = set(string.ascii_letters + string.digits + '_')
 
 
-def escape_sql_identifier(path: str, optional: bool | None = True) -> str:
+def escape_sql_identifier(path: str) -> str:
```
JCZuurmond, line 10: @larsgeorge-db and @nfx : Do you recall why this was optional in the first place?

In retrospect, I conclude it is confusing to have the optional flag as (new) contributors read/write code like `self._exec(f"TRUNCATE TABLE {escape_sql_identifier(full_name)}")` expecting the SQL identifier will be escaped while it is not. Optionally escaping the sql identifier is more clear outside the function:

```python
if should_escape_sql_identifier:
    full_name = escape_sql_identifier(full_name)
```

ericvergnaud (approved):
lgtm

nfx (approved):
lgtm

nfx (approved):

PR #2495, opened 2024-08-27
Do not normalize cases when reformatting SQL queries in CI check

JCZuurmond:
Do not normalize cases when reformatting SQL queries in CI check. The CI was blocking due to changes when normalising the case sensitive columns, [example here](https://github.com/databrickslabs/ucx/actions/runs/10578058469/job/29307356485?pr=2485)


nfx (approved):
lgtm

Issue #1783, opened 2024-05-28
[FEATURE]: Enable download and upload of ucx files for a collection of workspaces


HariGS-DB:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Currently, for various ucx commands, there require an option for the user to edit some files (Ex: azure_storage_account_info, uc_roles_access, mapping.csv, aws_instance_profile_info). Currently these files are edited per workspace. For customers having multiple workspaces we have introduced the concept of collection where they could run cli cmd for a collection of workspaces. These customers would also like the option to edit the files for a collection at one go by downloading them locally , editing them, syncing them back in the respective workspace.  

### Proposed Solution

Add cli cmd to download a file for a given collection
download the file by combining the data from all the workspaces in a collection include workspace ID if not
user uploads the data through cli command
sync the data in each workspace based on workspace id

JCZuurmond:
Implemented; see linked PRs

Issue #2507, opened 2024-08-29
[TECH DEBT]: Merge functionality for running ucx accross workspaces 


JCZuurmond:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Currently we are adding and have added functionality to run ucx accross workspaces, see [this issue](#891) for an overview. The functionality to do so became somewhat dispersed and repeated:
- The account installer, specifically its [`get_workspace_contexts](https://github.com/databrickslabs/ucx/blob/fc554aeddba5cb173762d82a3826a06578ea4692/src/databricks/labs/ucx/install.py#L775)
- The [account workspaces](https://github.com/databrickslabs/ucx/blob/fc554aeddba5cb173762d82a3826a06578ea4692/src/databricks/labs/ucx/account/workspaces.py#L13)
- The [cli](https://github.com/databrickslabs/ucx/blob/fc554aeddba5cb173762d82a3826a06578ea4692/src/databricks/labs/ucx/cli.py#L29)

### Proposed Solution

TBD. First suggestion below

## Context
We started with an account installer for installing ucx across workspaces and a sync workspace info command to have the workspace name available in all workspaces (as account admin is required to fetch this information). Now, we have and plan for more commands to run across a collection of workspaces. 

Preferably we minimize the need for account admin, though, if you run a command over a collection of workspace you require to be admin to each of those workspaces.

# Approach
1. Move the logic for retrieving multiple workspace clients and contexts to the `AccountWorkspaces` class. 
2. Then we can access those via the `AccountContext` that has the `AccountWorkspaces` as (cached) attribute
3. We access it in the account intstaller or any of the cli commands

JCZuurmond:
Note that we also use `AccountWorkspaces` in the `AccountAggregate` class

JCZuurmond:
Furthermore, for the cli command, the `AccountClient` is initialized via a new code path when introducing the `get_contexts`. Before we initialized it (implicitly) by setting the `is_account` flag to `True` in the cli command. Now, we also initialize it in the `get_contexts`. 

Maybe this requires a change to blueprint, for example
1. Make the `is_account` a flag optionally set by the user: `databricks labs <project> <command> --on-account True`
2. Or, (try to) set the `WorkspaceClient` also when the `is_account` flag is `True`

That would allow us to try to use the `AccountClient` to get a list of workpaces (contexts/clients), otherwise fallback on the `WorkspaceClient` provided by the blueprint command

[...]

PR #2508, opened 2024-08-29
Add `upload` cli command to upload file to a collection of workspaces

JCZuurmond:
## Changes
Add a `upload` cli command to upload file to a collection of workspaces.

### Linked issues

Progresses #1783

### Functionality

- [ ] added new CLI command: `upload`

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/cli.py
```diff
@@ -103,6 +103,14 @@ def sync_workspace_info(a: AccountClient):
     ctx.account_workspaces.sync_workspace_info()
 
 
+@ucx.command(is_account=True)
+def upload(path: Path, w: WorkspaceClient, run_as_collection: bool = False, a: AccountClient | None = None):
+    """Upload a file to the (collection of) workspace(s)"""
+    logger.warning("The schema of CSV files is NOT validated, ensure it is correct")
```
JCZuurmond, line 109: What do we think about this warning? Do we want to extend the logic, thus complicating the logic, to verify which files could be uploaded in a ucx installation and what there schemas are? Or is this fine(ish)

tests/unit/test_cli.py
```diff
@@ -526,102 +581,100 @@ def test_migrate_locations_aws(ws, caplog):
...
-    ws.clusters.list.return_value = []
-    cluster_remap(ws, prompts)
+    ws1.clusters.list.return_value = []
+    cluster_remap(ws1, prompts)
     assert "No cluster information present in the workspace" in caplog.messages
 
 
-def test_revert_cluster_remap(ws, caplog, mocker):
+def test_revert_cluster_remap(caplog):
+    # TODO: What is this test supposed to test? Why do we expect a TypeError below?
```
JCZuurmond, line 638: @prajin-29: Do you remember what this function is supposed to test? It is not clear to me why the type error below is expected given the test setup

JCZuurmond (commented):
Focusing on integration tests now

src/databricks/labs/ucx/cli.py
```diff
@@ -532,6 +533,22 @@ def join_collection(a: AccountClient, workspace_ids: str):
     account_installer.join_collection(w_ids)
 
 
+@ucx.command
+def upload(
```
JCZuurmond, line 537: Note that I on purpose did not add the `a: AccountClient | None` parameter as it will never be used given the current logic to retrieve contexts for multiple workspaces. 
- When settings `a: AccountClient | None` in the signature, it will only be not `None` if `is_account=True` inside the `ucx.command` decorator. However, in that case the `w: WorkspaceClient` points to the accounts endpoints, which does not allow to upload a file when `run_as_collection=False`.
- In the other case, when `AccountClient` is `None`, we are able to use the `w: WorkspaceClient` (`run_as_collection=False`) or get all workspaces in the collection (`run_as_collection=True`) to upload the file to. **Note: in the later, we prompt the user for the account id instead of (possibly) using the account profile from the Databricks configuration as we do when `@ucx.command(is_account=True`). 

There is an issue to merge this logic #2507, with that issue we should improve the integration testing for cross-account compatibility

tests/integration/install/test_installation.py
```diff
@@ -467,3 +467,16 @@ def test_installation_with_dependency_upload(ws, installation_ctx, mocker):
 
     installation_ctx.deployed_workflows.repair_run("failing")
     assert installation_ctx.deployed_workflows.validate_step("failing")
+
+
+def test_account_installer_returns_workspace_contexts(env_or_skip, installation_ctx):
```
JCZuurmond, line 472: This integration test fails. Though, we need the workspace contexts for proper integration testing. As mentioned in the PR top comment, I think we should make the integration testing part of #2507

README.md
```diff
@@ -1409,6 +1411,27 @@ By default, it will revert all the clusters present in the backup folder
...
+## `upload` command
+
+```text
+$ databricks labs ucx upload --file <file_path> --run-as-collection True
+21:31:29 WARNING [d.labs.ucx] The schema of CSV files is NOT validated, ensure it is correct
+21:31:29 INFO [d.labs.ucx] Finished uploading: <file_path>
+```
+
+Upload a file to a single workspace (`--run-as-collection False`) or a collection of workspaces
+(`--run-as-collection True`). This command is especially useful when uploading the same file to multiple workspace.
```
ericvergnaud, line 1423: workspace -> workspaces

README.md
```diff
@@ -1409,6 +1411,27 @@ By default, it will revert all the clusters present in the backup folder
...
+
+## `download` command
+
+```text
+$ databricks labs ucx download --file <file_path> --run-as-collection True
+21:31:29 INFO [d.labs.ucx] Finished downloading: <file_path>
+```
+
+Download a csv file from a single workspace (`--run-as-collection False`) or a collection of workspaces
+(`--run-as-collection True`). This command is especially useful when downloading the same file from multiple workspace.
```
ericvergnaud, line 1433: workspace -> workspaces

labs.yml
```diff
@@ -265,3 +265,18 @@ commands:
       - name: target-workspace-id
         description: (Optional) id of a workspace in the target collection. If not specified, ucx will prompt to select from a list
 
+  - name: upload
+    description: upload file to all workspaces in the account where ucx is installed
+    flags:
+      - name: file
+        description: The file to upload
+      - name: run-as-collection
+        description: Run the command for the collection of workspace with ucx installed. Default is False.
```
ericvergnaud, line 274: workspace -> workspaces

labs.yml
```diff
@@ -265,3 +265,18 @@ commands:
...
+
+  - name: download
+    description: download file from all workspaces in the account where ucx is installed
+    flags:
+      - name: file
+        description: The file to download
+      - name: run-as-collection
+        description: Run the command for the collection of workspace with ucx installed. Default is False.
```
ericvergnaud, line 282: workspace -> workspaces

src/databricks/labs/ucx/cli.py
ericvergnaud: should we keep things simple for the user and create 2 commands i.e. `upload_to_workspace` and `upload_to_account` ?

src/databricks/labs/ucx/cli.py
```diff
@@ -532,6 +535,65 @@ def join_collection(a: AccountClient, workspace_ids: str):
...
+    contexts = _get_workspace_contexts(w, run_as_collection=run_as_collection, a=a)
+    logger.warning("The schema of CSV files is NOT validated, ensure it is correct")
+    for ctx in contexts:
+        ctx.installation.upload(file.name, file.read_bytes())
+    if len(contexts) > 0:
+        logger.info(f"Finished uploading {file}")
+
+
+@ucx.command
+def download(
```
ericvergnaud, line 556: I'm a bit skeptical with the concept of merging downloaded files from different workspaces... I suggest downloading from only 1 workspace

src/databricks/labs/ucx/cli.py
ericvergnaud: What's the behavior if the file already exists in a workspace ? Shouldn't the user have an `overwrite` flag ? If `overwrite` is `False`, shouldn't the command do a pre-flight check ?

ericvergnaud (changes requested):

src/databricks/labs/ucx/cli.py
JCZuurmond: There is no option to upload a csv to a Databricks account. We can get the workspaces from the account after which we upload to each workspaces.

For the `overwrite`, that could make sense. Currently we always overwrite. It creates the question on when to break the upload:
- Per workspace, do we **not** upload if the file already exists within that workspace and `overwrite=False`?
- Per collection of workspaces, do we **not** upload the file to **any** workspace if the file already exists within **one** of the workspaces and `overwrite=False`?

Given that we want the command to work on the full collection of workspaces and that all the files ucx uses can be re-generated, I am oke within always overwritting the file to avoid introducing more code complexity

nfx (approved):
lgtm

tests/unit/test_cli.py
```diff
@@ -120,28 +130,57 @@ def download(path: str) -> io.StringIO | io.BytesIO:
     return workspace_client
 
 
-def test_workflow(ws, caplog):
-    workflows(ws)
+@pytest.fixture
+def ws1() -> WorkspaceClient:
```
nfx, line 134: this diff is too big. rename `ws1` back to `ws` and add new fixture only for `ws2`.

nfx (changes requested):
reduce the size of the test diff, otherwise lgtm

nfx (approved):
lgtm

PR #2404, opened 2024-08-09
More crawler snapshot refactoring

asnare:
## Changes

This PR is stacked on top of #2402 and is further refactors the crawler snapshotting pattern to use common code without changing the existing system behaviour. This is intended as groundwork to the next step, whereby `snapshot()` will be updated to support refreshing the inventory.

### Linked issues

Progresses #2074, stacked on top of #2402.

### Tests

- existing unit tests (with minor updates)
- existing integration tests

PR #2535, opened 2024-09-04
Remove unnecessary crawler API: `Mounts.inventorize_mounts()`

asnare:
## Changes

This PR removes the `Mounts.inventorize_mounts()` method; it was only used by unit tests that could use the `.snapshot()` API instead. This was noted as part of #2404, and is part of general cleanup of patterns we use for the crawlers:

https://github.com/databrickslabs/ucx/blob/68ecb02df0f02fcef1b5dc44a8421ccf480431c6/src/databricks/labs/ucx/hive_metastore/locations.py#L240

### Linked issues

Relates to #2404.

### Tests

- updated unit tests


nfx (approved):
lgtm

PR #2546, opened 2024-09-06
Uses PermissionDenied instead of PermissionError

pritishpai:
## Changes
Uses PermissionDenied instead of PermissionError for handling Databricks' object errors

### Linked issues
Resolves #2407 

### Tests
Used existing tests

PR #2541, opened 2024-09-05
Bugfix: `test_delete_ws_groups_should_delete_renamed_and_reflected_groups_only`

asnare:
## Changes

This PR updates an integration test that started failing after the Databricks 0.32.0 release. In this test we check the error message associated with an expected failure, but the message changed with the new version of the SDK so the test now fails.

_Note: This is probably a bug introduced in the upstream SDK release._ For a call to `ws.groups.get(id)` where the ID doesn't exist:

 - Previously we would get the message: `Group with id $ID not found.`
 - As of the new release we get: `None request failed`

### Linked issues

Resolves #2539.

### Tests

- updated integration test

PR #2557, opened 2024-09-09
Bugfix: `test_running_real_remove_backup_groups_job`

asnare:
## Changes

This PR updates an integration test that has started failing since the Databricks 0.32.0 release. In this test we check the error message associated with an expected failure, but the message changed with the new version of the SDK so the test now fails.

### Linked issues

Relates to #2541, affects #2546.

Triggered by upstream databricks/databricks-sdk-py#749.

### Tests

- updated integration test


pritishpai (approved):

nfx (approved):
lgtm

PR #2558, opened 2024-09-09
Added linting for `spark.sql(...)` calls

### LINKED ISSUE #2551: [BUG]: Only table-migration linting occurs for queries ran via spark.sql("some query...")

### Current Behavior

The query passed for execution to `spark.sql(...)` is only analyzed for table names.

### Expected Behavior

The query should be fully linted, using all SQL linters.

### Steps To Reproduce

_No response_

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

ericvergnaud:
## Changes
When linting queries in `spark.sql(...)` , current SparkSqlPyLinter linter only checks for table migration but does not use other SQL linters (such as DirectFsAccess linters).
This PR fixes the issue.

### Linked issues
Resolves #2551

### DIFF
--- a/src/databricks/labs/ucx/source_code/base.py
+++ b/src/databricks/labs/ucx/source_code/base.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import codecs
 import locale
 import logging
-from abc import abstractmethod
+from abc import abstractmethod, ABC
 from collections.abc import Iterable
 from dataclasses import dataclass
 from pathlib import Path
@@ -176,12 +176,22 @@ class PythonLinter(Linter):
     def lint_tree(self, tree: Tree) -> Iterable[Advice]: ...
 
 
-class Fixer:
+class Fixer(ABC):
+    @abstractmethod
+    def can_fix(self, advice_code: str) -> bool: ...
+
+    @abstractmethod
+    def apply(self, advice_code: str, source_code: str) -> str: ...
+
+
+class NamedFixer(Fixer, ABC):
+
+    @property
     @abstractmethod
     def name(self) -> str: ...
 
-    @abstractmethod
-    def apply(self, code: str) -> str: ...
+    def can_fix(self, advice_code: str) -> bool:
+        return self.name == advice_code
 
 
 # The default schema to use when the schema is not specified in a table reference
--- a/src/databricks/labs/ucx/source_code/base.py
+++ b/src/databricks/labs/ucx/source_code/base.py
@@ -177,21 +177,13 @@ class PythonLinter(Linter):
 
 
 class Fixer(ABC):
-    @abstractmethod
-    def can_fix(self, advice_code: str) -> bool: ...
-
-    @abstractmethod
-    def apply(self, advice_code: str, source_code: str) -> str: ...
-
-
-class NamedFixer(Fixer, ABC):
 
     @property
     @abstractmethod
     def name(self) -> str: ...
 
-    def can_fix(self, advice_code: str) -> bool:
-        return self.name == advice_code
+    @abstractmethod
+    def apply(self, code: str) -> str: ...
 
 
 # The default schema to use when the schema is not specified in a table reference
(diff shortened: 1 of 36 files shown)


src/databricks/labs/ucx/source_code/base.py
```diff
@@ -176,12 +176,22 @@ def lint(self, code: str) -> Iterable[Advice]:
...
 
 
-class Fixer:
+class Fixer(ABC):
     @abstractmethod
-    def name(self) -> str: ...
+    def can_fix(self, advice_code: str) -> bool: ...
+
+    @abstractmethod
+    def apply(self, advice_code: str, source_code: str) -> str: ...
```
nfx, line 184: this is Fixer interface that has nothing to do with this PR. revert this change.

src/databricks/labs/ucx/source_code/linters/pyspark.py
```diff
@@ -353,3 +319,59 @@ def _find_matcher(self, node: NodeNG):
...
+
+    def can_fix(self, advice_code: str) -> bool:
+        return advice_code.startswith(self._PREFIX)
+
+    def lint_tree(self, tree: Tree) -> Iterable[Advice]:
+        for call_node, query in self._visit_call_nodes(tree):
+            for value in InferredValue.infer_from_node(query):
+                if not value.is_inferred():
+                    yield Advisory.from_node(
+                        code=f"{self._PREFIX}cannot-autofix-table-reference",
```
nfx, line 340:
```suggestion
                        code=f"cannot-autofix-table-reference",
```
it's still the same problem code.

src/databricks/labs/ucx/source_code/linters/pyspark.py
```diff
@@ -353,3 +319,59 @@ def _find_matcher(self, node: NodeNG):
...
+                        message=f"Can't migrate table_name argument in '{query.as_string()}' because its value cannot be computed",
+                        node=call_node,
+                    )
+                    continue
+                for advice in self._sql_linter.lint(value.as_string()):
+                    advice = advice.replace(code=f"{self._PREFIX}{advice.code}")
```
nfx, line 346:
```suggestion
```
i don't think it's required.

nfx (changes requested):
revert changes that are out of scope of this PR

src/databricks/labs/ucx/source_code/base.py
ericvergnaud: It's in scope because SparkSqlPyLinter is both a linter and a fixer. It needs to route `apply` to the correct SQL fixer.

ericvergnaud:
The situation is that we have a top-level linter (SparkSqlPyLinter) that relies on a collection of SQL linters.
The issue is that SparkSqlPyLinter is also a Fixer, and relies on a collection of SQL Fixers.
When `apply` is invoked, it needs to extract the SQL, and route the advice to the correct SQL Fixer.
My proposal is to use a prefix in the advice, that SparkSqlPyLinter manages, but I'm happy to consider other ideas.
We can also have a specific SparkSqlPyLinter per SQLLinter. which removes the need for routing.

src/databricks/labs/ucx/source_code/base.py
```diff
@@ -176,7 +176,9 @@ def lint(self, code: str) -> Iterable[Advice]:
     def lint_tree(self, tree: Tree) -> Iterable[Advice]: ...
 
 
-class Fixer:
+class Fixer(ABC):
+
+    @property
```
ericvergnaud, line 181: fixes an inconsistency

ericvergnaud:
> The situation is that we have a top-level linter (SparkSqlPyLinter) that relies on a collection of SQL linters. The issue is that SparkSqlPyLinter is also a Fixer, and relies on a collection of SQL Fixers. When `apply` is invoked, it needs to extract the SQL, and route the advice to the correct SQL Fixer. My proposal is to use a prefix in the advice, that SparkSqlPyLinter manages, but I'm happy to consider other ideas. We can also have a specific SparkSqlPyLinter per SQLLinter. which removes the need for internal routing.

@nfx I've refactored to the latter: 1 SparkSqlPyLinter per SQLLinter

nfx (approved):
lgtm

Issue #2542, opened 2024-09-05
[BUG]: Installer is broken, or missing helpful error message


sdanielzafar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```
dan.zafar@YFV9607WJY ~ % databricks -v
Databricks CLI v0.225.0
dan.zafar@YFV9607WJY ~ % databricks labs install ucx
09:44:58  INFO [d.l.ucx.install] Installing UCX v0.34.0
Traceback (most recent call last):
  File "/Users/dan.zafar/.databricks/labs/ucx/lib/src/databricks/labs/ucx/install.py", line 303, in configure
    config = self.installation.load(WorkspaceConfig)
             ^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/dan.zafar/.databricks/labs/ucx/lib/src/databricks/labs/ucx/install.py", line 169, in installation
    return self.product_info.current_installation(self.workspace_client)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dan.zafar/.databricks/labs/ucx/state/venv/lib/python3.11/site-packages/databricks/labs/blueprint/wheels.py", line 130, in current_installation
    return Installation.current(ws, self.product_name())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dan.zafar/.databricks/labs/ucx/state/venv/lib/python3.11/site-packages/databricks/labs/blueprint/installation.py", line 86, in current
    user_folder = cls._user_home_installation(ws, product)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dan.zafar/.databricks/labs/ucx/state/venv/lib/python3.11/site-packages/databricks/labs/blueprint/installation.py", line 423, in _user_home_installation
    me = ws.current_user.me()
         ^^^^^^^^^^^^^^^^^^^^
  File "/Users/dan.zafar/.databricks/labs/ucx/state/venv/lib/python3.11/site-packages/databricks/sdk/service/iam.py", line 2270, in me
[...]

PR #2561, opened 2024-09-10
Prevent UCX from using the 0.32.{0,1} releases of the Databricks SDK.

asnare:
## Changes

This PR updates the project dependencies so that we don't use the 0.32.0 or 0.32.1 releases of the SDK. One of the [changes](/databricks/databricks-sdk-py/pull/741) in that release seems to have introduced some regressions in existing error-handling. Although we can work around some of them, others (such as handling of permission-denied errors) we can't.

Given that this particular version of the SDK didn't introduce new features/behaviour that we rely upon, the safest path seems to be ensure that we don't install it specifically.

### Linked issues

Relates: databricks/databricks-sdk-py#749, databricks/databricks-sdk-py#755, 

Fixes: #2542


pyproject.toml
```diff
@@ -44,7 +44,7 @@ classifiers = [
     "Topic :: Utilities",
 ]
 
-dependencies = ["databricks-sdk~=0.30",
+dependencies = ["databricks-sdk~=0.30,!=0.32.0",
```
JCZuurmond, line 47: Please add a comment explaining why we exclude this version

JCZuurmond (commented):
See comment

asnare:
The 0.32.1 release has been made, which partly addresses the error-handling regressions. It doesn't address #2542 though.

asnare:
Version 0.32.2 of the SDK has been released, and it includes a fix for the remaining regression we noticed with error-handling. Rather than complicate the dependency-specification we'll just leave it as-is for now.

PR #2563, opened 2024-09-10
Escape column names in target tables of the table migration

### LINKED ISSUE #2544: [BUG]: migrate-tables fails to migrate table with a decimal as a column name

### Current Behavior

migrate-tables unable to migrate table with decimals as a column name, i.e. 10.0.

### Expected Behavior

Tables are migrated with non-standard column names wrapped in quotes.

### Steps To Reproduce

_No response_

### Cloud

Azure

### Operating System

Windows

### Version

latest via Databricks CLI

### Relevant log output

```shell
Failed to migrate table hive_metastore.[schema].[table] to [catalog].[schema].[table]: INVALID_PARAMETER_VALUE: Invalid input: RPC CreateTable Field managedcatalog.ColumnInfo.name: At columns.0: name "" is not a valid name
```

JCZuurmond:
## Changes
Escape column names in target tables of the table migration

### Linked issues
Resolves #2544

### Functionality

- [x] modified existing workflow: `-migration-`-ones

### Tests

- [x] added unit tests
- [x] changed integration tests


tests/integration/hive_metastore/test_migrate.py
```diff
@@ -663,6 +663,7 @@ def test_migrate_table_in_mount(
     src_external_table = runtime_ctx.make_table(
         schema_name=make_schema(catalog_name="hive_metastore", name=f'mounted_{env_or_skip("TEST_MOUNT_NAME")}').name,
         external_delta=f'dbfs:/mnt/{env_or_skip("TEST_MOUNT_NAME")}/a/b/{tbl_path}',
+        columns=[ColumnInfo(name="0", type_name=ColumnTypeName.STRING)],
```
asnare, line 666: The issue being resolved specifically has a problem with `.` in the column name, so I'd like to see that explicitly covered by the unit test.

In addition I'd like to see a few more examples (similar approach to the tests for `escape_sql_identifier()` routine). Aside from the usual suspects, I'd particularly like to see a column name with <code>&#x60;</code> somewhere in the middle of the name. (That's a corner-case that will fail at the moment, I think.)

tests/integration/hive_metastore/test_migrate.py
asnare, line 1: Great to see this is being covered by an integration test.

asnare (changes requested):

tests/integration/hive_metastore/test_migrate.py
JCZuurmond: Alright, thanks, those were indeed not handled. See my changes

src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -1041,6 +1044,18 @@ def create(  # pylint: disable=too-many-locals,too-many-arguments,too-many-state
         view_text = None
         full_name = f"{catalog_name}.{schema_name}.{name}".lower()
         ddl = f'CREATE {"VIEW" if view else "TABLE"} {full_name}'
+        if columns is None:
+            schema = "(id INT, value STRING)"
+        else:
+            schema = "("
+            for column in columns:
+                schema += escape_sql_identifier(column.name or "", maxsplit=0)
```
asnare, line 1052: That empty column name, is that valid? 🤔

tests/integration/hive_metastore/test_migrate.py
```diff
@@ -134,7 +138,11 @@ def test_migrate_external_table(
     make_mounted_location,
 ):
     src_schema = runtime_ctx.make_schema(catalog_name="hive_metastore")
-    src_external_table = runtime_ctx.make_table(schema_name=src_schema.name, external_csv=make_mounted_location)
+    src_external_table = runtime_ctx.make_table(
+        schema_name=src_schema.name,
+        external_csv=make_mounted_location,
+        columns=[ColumnInfo(name="`backtick", type_name=ColumnTypeName.STRING)],  # Test with column that needs escaping
```
asnare, line 144: The backtick (<code>&#x60;</code>) needs to be in the middle of the identifier, please. 😀 At either end I don't fully trust the escaper not to be stripping it.

src/databricks/labs/ucx/mixins/fixtures.py
JCZuurmond: yeahhh, I do not know what to do with this. The `ColumnInfo.name` is optional, but when does that happen? Should we use postitional column names in that case? If so, that does not work with creating the schema. Or should I skip them? Maybe skipping is better

JCZuurmond: Note that this is in a fixture, therefor it does not matter as much as it is just for test cases

JCZuurmond: Handled by raising a `ValueError` when the column name is not specified

tests/integration/hive_metastore/test_migrate.py
```diff
@@ -663,6 +672,8 @@ def test_migrate_table_in_mount(
     src_external_table = runtime_ctx.make_table(
         schema_name=make_schema(catalog_name="hive_metastore", name=f'mounted_{env_or_skip("TEST_MOUNT_NAME")}').name,
         external_delta=f'dbfs:/mnt/{env_or_skip("TEST_MOUNT_NAME")}/a/b/{tbl_path}',
+        # Test with column that needs escaping
+        columns=[ColumnInfo(name="1-0`.0-ugly-column", type_name=ColumnTypeName.STRING)],
```
asnare, line 676: 😍

tests/unit/mixins/test_fixtures.py
```diff
@@ -0,0 +1,36 @@
...
+# pylint: disable-next=unused-import
+from databricks.labs.ucx.mixins.fixtures import make_table, make_schema, make_random  # fmt: skip  # noqa: F401
+
+
+@pytest.fixture
+def ws():
+    workspace_client = create_autospec(WorkspaceClient)
+    yield workspace_client
+    # Circumvent linter warning; workspace client instance is never called at class level
+    workspace_client.assert_not_called()
```
asnare, line 17: I'm having trouble figuring out where this is used…?

JCZuurmond: Do you mean the fixture itself or the `assert_not_called`?

- Fixture itself: used implicitly by `make_table` fixture, see comment in imports
- The `assert_not_called` circumvents linter warning, see comment on line before

asnare: Ah. I missed what's going on here:

 - The import of those methods effectively means the fixtures are declared within this module.
 - One of them (`make_table`) in turns needs a `ws` fixture to be available.
 - The parameter names on the test methods shadow the (global) imports, which is why you need those suppressions.

nfx (changes requested):

nfx (approved):
lgtm

PR #2566, opened 2024-09-10
Crawler snapshots: allow refreshing

asnare:
## Changes

This PR updates the crawler internals so that when requesting a snapshot a refresh can be forced. In addition, the default semantics for crawlers have been updated so that they use an overwrite operation to write the data rather than appending (with the assumption that the table is already empty).

### Linked issues

- Progresses: #2074
- Resolves: #2576
- Closes: #2597

### Functionality

- modifies all existing crawlers

### Tests

- updated unit tests
- existing integration tests

### DIFF
--- a/src/databricks/labs/ucx/framework/crawlers.py
+++ b/src/databricks/labs/ucx/framework/crawlers.py
@@ -146,9 +146,3 @@ class CrawlerBase(ABC, Generic[Result]):
     def _update_snapshot(self, items: Sequence[Result], mode: Literal["append", "overwrite"] = "append") -> None:
         logger.debug(f"[{self.full_name}] found {len(items)} new records for {self._table}")
         self._backend.save_table(self.full_name, items, self._klass, mode=mode)
-
-    def _append_records(self, items: Sequence[Result]):
-        self._update_snapshot(items, mode="append")
-
-    def _overwrite_records(self, items: Sequence[Result]):
-        self._update_snapshot(items, mode="overwrite")
--- a/src/databricks/labs/ucx/framework/crawlers.py
+++ b/src/databricks/labs/ucx/framework/crawlers.py
@@ -138,7 +138,7 @@ class CrawlerBase(ABC, Generic[Result]):
                     return cached_results
             except NotFound:
                 pass
-        logger.debug(f"[{self.full_name}] crawling new batch for {self._table}")
+        logger.debug(f"[{self.full_name}] crawling new set of snapshot data for {self._table}")
         loaded_records = list(loader())
         self._update_snapshot(loaded_records, mode="overwrite")
         return loaded_records
--- a/src/databricks/labs/ucx/framework/crawlers.py
+++ b/src/databricks/labs/ucx/framework/crawlers.py
@@ -134,7 +134,7 @@ class CrawlerBase(ABC, Generic[Result]):
             logger.debug(f"[{self.full_name}] fetching {self._table} inventory")
             try:
                 cached_results = list(fetcher())
-                if len(cached_results) > 0:
+                if cached_results:
                     return cached_results
             except NotFound:
                 pass
--- a/src/databricks/labs/ucx/framework/crawlers.py
+++ b/src/databricks/labs/ucx/framework/crawlers.py
@@ -1,7 +1,7 @@
 import logging
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable, Sequence
-from typing import ClassVar, Generic, Protocol, TypeVar
+from typing import ClassVar, Generic, Literal, Protocol, TypeVar
 
 from databricks.labs.lsql.backends import SqlBackend
 from databricks.sdk.errors import NotFound
@@ -91,8 +91,8 @@ class CrawlerBase(ABC, Generic[Result]):
             return None
         return cls._valid(name)
 
-    def snapshot(self) -> Iterable[Result]:
-        return self._snapshot(self._try_fetch, self._crawl)
+    def snapshot(self, *, force_refresh: bool = False) -> Iterable[Result]:
+        return self._snapshot(self._try_fetch, self._crawl, force_refresh=force_refresh)
 
     @abstractmethod
     def _try_fetch(self) -> Iterable[Result]:
@@ -110,7 +110,7 @@ class CrawlerBase(ABC, Generic[Result]):
             Iterable[Result]: Records that capture the results of crawling the environment.
         """
 
-    def _snapshot(self, fetcher: ResultFn, loader: ResultFn) -> list[Result]:
+    def _snapshot(self, fetcher: ResultFn, loader: ResultFn, *, force_refresh: bool) -> list[Result]:
         """
         Tries to load dataset of records with `fetcher` function, otherwise automatically creates
         a table with the schema defined in the class of the first row and executes `loader` function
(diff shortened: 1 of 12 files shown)


src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -119,26 +119,36 @@ def _snapshot(self, fetcher: ResultFn, loader: ResultFn) -> list[Result]:
...
 
-    def _append_records(self, items: Sequence[Result]):
+    def _update_snapshot(self, items: Sequence[Result], mode: Literal["append", "overwrite"] = "append") -> None:
         logger.debug(f"[{self.full_name}] found {len(items)} new records for {self._table}")
-        self._backend.save_table(self.full_name, items, self._klass, mode="append")
+        self._backend.save_table(self.full_name, items, self._klass, mode=mode)
+
+    def _append_records(self, items: Sequence[Result]):
+        self._update_snapshot(items, mode="append")
+
```
asnare, line 152: PR marked as draft because although I thought we had some crawlers than incrementally append records, we don't seem to. I'm currently checking this and can probably remove this method once I've verified this.

asnare:
Removing draft status to trigger integration tests; not ready for merge though.

asnare:
> ❌ test_reflect_account_groups_on_workspace_recovers_when_group_already_exists: databricks.labs.blueprint.parallel.ManyError: Detected 1 failures: TimeoutError: Timed out after 0:20:00 (20m20.787s)

This is also failing on `main` and unrelated to this PR. (See #2559.)

src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -110,7 +110,7 @@ def _crawl(self) -> Iterable[Result]:
             Iterable[Result]: Records that capture the results of crawling the environment.
         """
 
-    def _snapshot(self, fetcher: ResultFn, loader: ResultFn) -> list[Result]:
+    def _snapshot(self, fetcher: ResultFn, loader: ResultFn, *, force_refresh: bool) -> list[Result]:
```
JCZuurmond, line 113: I slightly doubt the naming `force_refresh`. Why not just `refresh`; what does "force" add? Otherwise, maybe it should be `force_fetch` as that is forced. My preference goes to `crawler.snapshot(refresh=True)`.

Note that in dbt they use "full refresh" as they support incremental refreshing. In our case a "refresh" is always "full"

src/databricks/labs/ucx/hive_metastore/grants.py
```diff
@@ -205,9 +205,9 @@ def __init__(self, tc: TablesCrawler, udf: UdfsCrawler, include_databases: list[
         self._udf = udf
         self._include_databases = include_databases
 
-    def snapshot(self) -> Iterable[Grant]:
+    def snapshot(self, *, force_refresh: bool = False) -> Iterable[Grant]:
         try:
-            return super().snapshot()
+            return super().snapshot(force_refresh=force_refresh)
         except Exception as e:  # pylint: disable=broad-exception-caught
```
JCZuurmond, line 211: Something for outside this PR, but do not like this broad exception

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -359,7 +359,10 @@ def __init__(
             irrelevant_patterns.update(exclude_paths_in_mount)
         self._fiter_paths = irrelevant_patterns
 
-    def snapshot(self) -> list[Table]:
+    def snapshot(self, *, force_refresh: bool = True) -> list[Table]:
+        # TODO: Figure out what this should do (and refactor); non-trivial due to it overwriting the static table data.
```
JCZuurmond, line 363: yes!

tests/unit/framework/test_crawlers.py
```diff
@@ -61,17 +62,84 @@ def test_full_name():
...
+
+    result = cb.snapshot()
+
+    mock_fetcher.assert_called_once()
+    mock_loader.assert_not_called()
+    assert [Baz(first="first")] == result
+
+
+def test_snapshot_crawls_if_refresh_forced() -> None:
+    """Check that existing data is used (with no crawl) if the fetcher can load the snapshot data."""
```
JCZuurmond, line 108: Update docstring

tests/unit/framework/test_crawlers.py
```diff
@@ -61,17 +62,84 @@ def test_full_name():
...
+    cb = _CrawlerFixture[Baz](mock_backend, "a", "b", "c", Baz, fetcher=mock_fetcher, loader=mock_loader)
+
+    result = cb.snapshot(force_refresh=True)
+
+    mock_fetcher.assert_not_called()
+    mock_loader.assert_called_once()
+    assert [Baz(first="second")] == result
+
+
+def test_snapshot_force_refresh_replaces_prior_data() -> None:
```
JCZuurmond, line 121: Add docstring

tests/unit/framework/test_crawlers.py
```diff
@@ -61,17 +62,84 @@ def test_full_name():
...
+    mock_backend = MockBackend()
+    mock_fetcher = Mock(side_effect=RuntimeError("never called"))
+    mock_loader = Mock(return_value=[Baz(first="second")])
+    cb = _CrawlerFixture[Baz](mock_backend, "a", "b", "c", Baz, fetcher=mock_fetcher, loader=mock_loader)
+
+    cb.snapshot(force_refresh=True)
+
+    assert [Row(first="second", second=None)] == mock_backend.rows_written_for("a.b.c", mode="overwrite")
```
JCZuurmond, line 129: missing mock asserts

JCZuurmond (commented):
Needs some further research and cleaning as you mentioned, otherwise functionality looks good to me

src/databricks/labs/ucx/framework/crawlers.py
asnare: Naming is hard. ;)

The `force_` prefix is present to make the purpose unambiguous. Without this prefix it could be misleading: `refresh=False`, at face value, could be read to indicate that it's not allowed to perform a crawl. (However it will if the underlying table doesn't exist or is empty.) As such the name 'force_refresh' is intended to convey that if it's `True` we definitely will, but if it's `False` that doesn't mean we won't.

I don't like the idea of using `fetch` in the name: use of the term fetch is inconsistent, and internally within the class it means "load the snapshot from the table" instead of "perform a crawl" as you intended. For now cleaning up the use of this term is tech debt and outside the scope of this PR.

Does this make sense?

JCZuurmond: Fair enough, I agree with your reasoning. Naming is hard indeed :/

src/databricks/labs/ucx/hive_metastore/grants.py
asnare: I don't either.

tests/unit/framework/test_crawlers.py
asnare: 🙈

tests/unit/framework/test_crawlers.py
asnare: It was a conscious decision to leave them out: I felt that checking those interactions wasn't the target of the test and there are different unit tests just for them.

I've dropped them back in though because they don't do any harm and help the reader understand what's (supposed to be) happening.

asnare:
> ❌ test_migrate_table_in_mount: databricks.sdk.errors.platform.BadRequest: [SCHEMA_ALREADY_EXISTS] Cannot create schema `mounted_TEST_MOUNT_NAME` because it already exists. (1.748s)

This was due to concurrent CI runs and appears to also be an issue on `main`; #2570 will address this.

src/databricks/labs/ucx/hive_metastore/locations.py
asnare: @JCZuurmond: Do you think the changes in a624f31 are sufficient for now?

src/databricks/labs/ucx/framework/crawlers.py
asnare: All crawlers now use the common flow, so the append/overwrite methods have been removed.

src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -134,21 +134,15 @@ def _snapshot(self, fetcher: ResultFn, loader: ResultFn, *, force_refresh: bool)
             logger.debug(f"[{self.full_name}] fetching {self._table} inventory")
             try:
                 cached_results = list(fetcher())
-                if len(cached_results) > 0:
+                if cached_results:
```
JCZuurmond, line 137: Why do you remove the `len` check?

asnare: It's not needed: an empty list is `False` in a boolean context. (And conversely, a list with any elements is `True`.) This is the [pythonic way](https://peps.python.org/pep-0008/#programming-recommendations):

> For sequences, (strings, lists, tuples), use the fact that empty sequences are false:
```python
# Correct:
if not seq:
if seq:
# Wrong:
if len(seq):
if not len(seq):
```

PR #2604, opened 2024-09-12
[TECH DEBT]: Deal with some nits in the existing crawler code

asnare:
## Changes

This PR deals with some nits that I noticed amongst the crawlers while working on #2566, but omitted to avoid distracting from the focus of that PR. These do not result in any functional changes.


nfx (approved):
lgtm

PR #2631, opened 2024-09-13
Integration test reliability: update some tests to ensure they use unique group names

asnare:
## Changes

This PR updates two integration tests that were using a fixed name for the fixture objects that they create. These can sometimes (spuriously) fail due to concurrent test runs against the same infrastructure.

### Tests

 - updated integration tests

### DIFF
--- a/tests/integration/framework/test_fixtures.py
+++ b/tests/integration/framework/test_fixtures.py
@@ -26,8 +26,9 @@ def test_user(make_user):
     logger.info(f"created {make_user()}")
 
 
-def test_group(make_group, make_user):
-    logger.info(f'created {make_group(display_name="abc", members=[make_user().id])}')
+def test_group(make_group, make_user, make_random):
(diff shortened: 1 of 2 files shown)


tests/integration/framework/test_fixtures.py
```diff
@@ -26,8 +26,9 @@ def test_user(make_user):
     logger.info(f"created {make_user()}")
 
 
-def test_group(make_group, make_user):
-    logger.info(f'created {make_group(display_name="abc", members=[make_user().id])}')
+def test_group(make_group, make_user, make_random):
+    display_name = f"abc-{make_random(4)}"
+    logger.info(f'created {make_group(display_name=display_name, members=[make_user().id])}')
```
JCZuurmond, line 31: I understand it was not part of your change, but I would like to see an assert in the test

JCZuurmond (approved):
See comment

nfx (approved):
lgtm

PR #2624, opened 2024-09-13
Fix failing integration test: `test_reflect_account_groups_on_workspace_skips_groups_that_already_exists_in_the_workspace`

JCZuurmond:
## Changes
Also check for **workspace** groups to exists with the same name. Follow up on #2615

In hindsight, what might have happened before: the workspace group was not created before the account group was reflected in the workspace, thus the test passes. With the changes in PR #2615, we got more consistent behaviour because we wait for the workspace group to exists before running the to-be-tested code. However, the code did not handle that situation. This PR handles when a workspace group has the same name as an to-be-reflected-in-the-workspace account group

### Linked issues

Resolves #2623

### Functionality

- [x] modified existing workflow: `group-migration`

### Tests

- [x] modified integration tests: `test_reflect_account_groups_on_workspace_skips_account_groups_when_a_workspace_group_has_same_name`

PR #2632, opened 2024-09-13
Reliability fix: `test_group_matching_names`

asnare:
## Changes

This PR updates the `test_group_matching_names` integration test so that the temporary groups it creates won't be removed while the test is underway. (Without this the test can occasionally fail.)

### Tests

 - updated integration test


tests/integration/workspace_access/test_groups.py
```diff
@@ -199,14 +201,12 @@ def test_group_matching_names(
...
-    account_group_name = f"same_group_[{rand_elem}]"
+    purge_suffix = get_purge_suffix()
+    workspace_group_name = f"test_group_{rand_elem}_{purge_suffix}"
+    account_group_name = f"same_group_[{rand_elem}]_{purge_suffix}"
     user1 = make_user()
     members1 = [user1.id]
-    members2 = [user1.id]
-    if not same_user:
-        user2 = make_user()
-        members2 = [user2.id]
```
asnare, line 209: This change is here because the introduction of the `purge_suffix` variable above means the method had too many local variables for the linter. This was the easiest way to remove a local variable and avoid triggering the linter.

tests/integration/workspace_access/test_groups.py
```diff
@@ -199,14 +201,12 @@ def test_group_matching_names(
...
+    members2 = [user1.id] if same_user else [make_user().id]
```
asnare, line 209: This change is here because the introduction of the `purge_suffix` variable above means the method had too many local variables for the linter. This was the easiest way to remove a local variable and avoid triggering the linter.

asnare:
> ❌ test_reflect_account_groups_on_workspace_skips_groups_that_already_exists_in_the_workspace: […]

This is unrelated to this PR, and being addressed in #2624.

nfx (approved):

PR #2629, opened 2024-09-13
Fix test integration failure: `test_running_real_migrate_groups_job` 

JCZuurmond:
## Changes
Fix test integration failure: `test_running_real_migrate_groups_job` by waiting for the test group to be provisioned. Also the test assert failure is improved by first checking if the group is found.

### Linked issues
Resolves #2628

### Tests

- [x] modified integration tests: `test_running_real_migrate_groups_job`

PR #2634, opened 2024-09-13
Integration test reliability: improve group fixture consistency

asnare:
## Changes

This PR updates the way we create groups as part of integration tests to improve repeatability, eventually reliability although this might expose (existing) issues with some tests. The changes are:

 - The `wait_for_provisioning` implementation has been extended to use the double-check approach against both the `.get()` and `.list()` APIs.
 - We now default to waiting for group provisioning instead of returning immediately. This is a safer default, because it puts the onus on a test to explicitly declare that it doesn't need to wait for group provisioning to complete.
 - The `wait_for_provisioning` implementation is also available for tests to use themselves. Integration tests that create multiple groups have been updated so that they only wait _after_ creating all the groups rather than waiting sequentially after each individual group has been created. (This is an optimisation, intended to keep the run time of tests down.)

### Linked issues

Relates: #2629

### Tests

 - Updated integration tests.


src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -638,6 +639,33 @@ def _scim_values(ids: list[str]) -> list[iam.ComplexValue]:
     return [iam.ComplexValue(value=x) for x in ids]
 
 
+def wait_group_provisioned(interface: AccountGroupsAPI | GroupsAPI, *groups: Group) -> None:
+    # The groups API is eventually consistent, but not monotonically consistent. Here we try to compensate for
+    # the lack of monotonic consistency by requiring that two subsequent calls confirm it exists. REST API
+    # internals cache things for up to 60s, we see times close to this during tests. The retry timeout reflects
+    # this: if it's taking much longer then something else is wrong.
+
+    @retried(on=[NotFound], timeout=timedelta(seconds=90))
```
nfx, line 648: please also make the relevant PR into https://github.com/databrickslabs/pytester

src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -638,6 +639,33 @@ def _scim_values(ids: list[str]) -> list[iam.ComplexValue]:
     return [iam.ComplexValue(value=x) for x in ids]
 
 
+def wait_group_provisioned(interface: AccountGroupsAPI | GroupsAPI, *groups: Group) -> None:
```
nfx, line 642: encode it into `_make_group`, make waiting by default with opt-out

nfx (changes requested):

src/databricks/labs/ucx/mixins/fixtures.py
asnare: @nfx: I think this PR already does this?

This function is exposed for tests (and fixtures such as `make_ucx_group()`) that create multiple groups so that they can create them all and then wait on them all at once instead of having to wait (~60s) for each one before creating the next group and so forth.

I've added a docstring to emphasise that tests should only use this when creating multiple groups.

tests/integration/conftest.py
```diff
@@ -98,8 +98,13 @@ def sql_fetch_all(sql_backend):
 
 
 @pytest.fixture
-def make_ucx_group(make_random, make_group, make_acc_group, make_user):
-    def inner(workspace_group_name=None, account_group_name=None, **kwargs):
+def make_ucx_group(acc, ws, make_random, make_group, make_acc_group, make_user):
+    def inner(
+        workspace_group_name: str | None = None,
+        account_group_name: str | None = None,
+        wait_for_provisioning: bool = True,
```
nfx, line 105:
```suggestion
```
don't make it configurable and do it by default

tests/integration/conftest.py
```diff
@@ -110,20 +115,33 @@ def inner(workspace_group_name=None, account_group_name=None, **kwargs):
             display_name=workspace_group_name,
             members=members,
             entitlements=["allow-cluster-create"],
+            wait_for_provisioning=False,
+            **kwargs,
+        )
+        acc_group = make_acc_group(
+            display_name=account_group_name,
+            members=members,
+            wait_for_provisioning=False,
```
nfx, line 124:
```suggestion
            **kwargs,
        )
        acc_group = make_acc_group(
            display_name=account_group_name,
            members=members,
```
do it by default

tests/integration/hive_metastore/test_grants.py
```diff
@@ -2,19 +2,21 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 @retried(on=[NotFound, TimeoutError], timeout=timedelta(minutes=3))
-def test_all_grants_in_databases(runtime_ctx, sql_backend, make_group):
-    group_a = make_group()
-    group_b = make_group()
+def test_all_grants_in_databases(ws, runtime_ctx, sql_backend, make_group):
+    group_a = make_group(wait_for_provisioning=False)
```
nfx, line 18: wait by default

tests/integration/hive_metastore/test_grants.py
```diff
@@ -25,6 +27,8 @@ def test_all_grants_in_databases(runtime_ctx, sql_backend, make_group):
     view_d = runtime_ctx.make_table(schema_name=schema_a.name, view=True, ctas="SELECT id FROM range(10)")
     table_e = runtime_ctx.make_table(schema_name=schema_c.name)
 
+    wait_group_provisioned(ws.groups, group_a, group_b)
```
nfx, line 30: don't expose this function as public api, wait by default

nfx (changes requested):
don't expose `wait_group_provisioned` to other tests. wait by default

src/databricks/labs/ucx/mixins/fixtures.py
asnare: This method is no longer public, and the fixture always (unconditionally) invokes it after each group is created; callers can no longer defer the waiting.

tests/integration/conftest.py
asnare: Done.

tests/integration/conftest.py
asnare: Done.

tests/integration/hive_metastore/test_grants.py
asnare: Updated; this method is no longer exposed for tests to use.

tests/integration/hive_metastore/test_grants.py
asnare: Done.

src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -15,7 +15,7 @@
 import pytest
 from databricks.labs.lsql.backends import StatementExecutionBackend
 from databricks.labs.blueprint.commands import CommandExecutor
-from databricks.sdk import AccountClient, WorkspaceClient
+from databricks.sdk import AccountClient, WorkspaceClient, GroupsAPI, AccountGroupsAPI
```
JCZuurmond, line 18: isort

src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -638,6 +639,47 @@ def _scim_values(ids: list[str]) -> list[iam.ComplexValue]:
...
+def _wait_group_provisioned(interface: AccountGroupsAPI | GroupsAPI, group: Group) -> None:
+    """Wait for a group to be visible via the supplied group interface.
+
+    Due to consistency issues in the group-management APIs, new groups are not always visible in a consistent manner
+    after being created or modified. This method can be used to mitigate against this by checking that a group:
+
+     - Is visible via the `.get()` interface;
+     - Is visible via the `.list()` interface that enumerates groups.
+
+    Visibility is assumed when 2 calls in a row return the expected results.
```
JCZuurmond, line 651: Why is the second time needed?

tests/integration/hive_metastore/test_grants.py
```diff
@@ -12,7 +12,7 @@
 
 
 @retried(on=[NotFound, TimeoutError], timeout=timedelta(minutes=3))
-def test_all_grants_in_databases(runtime_ctx, sql_backend, make_group):
+def test_all_grants_in_databases(ws, runtime_ctx, sql_backend, make_group):
```
JCZuurmond, line 15: Could you pick this up also?
```suggestion
def test_all_grants_in_databases(ws, runtime_ctx, sql_backend, make_group) -> None:
```

JCZuurmond (commented):
code lgtm, hopefully fixes #2660. CI is failing though, could you retrigger it?

src/databricks/labs/ucx/mixins/fixtures.py
asnare: The second time is needed because it's not monotonically consistent: after a single call returns the correct result, another one after that can return the previous (stale) results.

The 2 calls is not a guarantee, but getting 2 in a row that succeed means it's much more likely that subsequent calls will also be consistent. Yes, after 2 there's a chance that a third will be wrong again but it's less likely. This approach is what we use elsewhere (during group migration) and is also the approach used by the terraform provider to mitigate the same problem. (And to be clear, it's mitigation not prevention.)

tests/integration/hive_metastore/test_grants.py
asnare: Oops, I've backed this out: it was left over after reverting some changes made earlier in the PR.

JCZuurmond: I think you still left some changes?

src/databricks/labs/ucx/mixins/fixtures.py
```diff
@@ -638,6 +639,47 @@ def _scim_values(ids: list[str]) -> list[iam.ComplexValue]:
     return [iam.ComplexValue(value=x) for x in ids]
 
 
+def _wait_group_provisioned(interface: AccountGroupsAPI | GroupsAPI, group: Group) -> None:
```
nfx, line 642: please do it into https://github.com/databrickslabs/pytester

nfx (changes requested):

src/databricks/labs/ucx/mixins/fixtures.py
asnare: Yes, I'll move these changes over there.

nfx: do we need this PR here then?

asnare: Nope; it can be closed.

asnare:
The changes in this PR will be implemented in https://github.com/databrickslabs/pytester instead.

PR #2640, opened 2024-09-16
Let `create-uber-principal` command run on collection of workspaces

### LINKED ISSUE #2605: [FEATURE] Let `create-uber-principal` command run as collection
### Summary

Enable `create-uber-principal` command to run as collection. See finished issues mentioned [here](https://github.com/databrickslabs/ucx/issues/1782) for reference.

JCZuurmond:
## Changes
Let `create-uber-principal` command run on a collection of workspaces.

### Linked issues

Resolves #2605

### Functionality

- [x] modified existing command: `databricks labs ucx create-uber-principal`

### Tests

- [x] manually tested
- [x] added unit tests
- [ ] ~added integration tests~ : Covering after #2507

### DIFF
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -298,22 +298,28 @@ def alias(
 
 @ucx.command
 def create_uber_principal(
-    w: WorkspaceClient,
     prompts: Prompts,
+    w: WorkspaceClient,
     ctx: WorkspaceContext | None = None,
+    run_as_collection: bool = False,
+    a: AccountClient | None = None,
     **named_parameters,
 ):
     """For azure cloud, creates a service principal and gives STORAGE BLOB READER access on all the storage account
     used by tables in the workspace and stores the spn info in the UCX cluster policy. For aws,
     it identifies all s3 buckets used by the Instance Profiles configured in the workspace.
     Pass subscription_id for azure and aws_profile for aws."""
-    if not ctx:
-        ctx = WorkspaceContext(w, named_parameters)
-    if ctx.is_azure:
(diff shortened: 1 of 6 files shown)


src/databricks/labs/ucx/cli.py
```diff
@@ -292,19 +292,25 @@ def create_uber_principal(
...
-    if ctx.is_aws:
-        return ctx.aws_resource_permissions.create_uber_principal(prompts)
-    raise ValueError("Unsupported cloud provider")
+    if ctx:
+        workspace_contexts = [ctx]
+    else:
+        workspace_contexts = _get_workspace_contexts(w, a, run_as_collection, **named_parameters)
+    for workspace_context in workspace_contexts:
+        if workspace_context.is_azure:
+            workspace_context.azure_resource_permissions.create_uber_principal(prompts)
```
JCZuurmond, line 309: @nfx : For the uber principal, I think we should (consider to) create one uber principal for all workspaces, instead of one per workspace. The word "uber" could (also) suggest "over" all workspaces (and storage accounts).

It means moving the for-loop over workspace contexts into the `create_uber_principal` invocation. However, it would require as that is part of the workspace context, which contains one workspace context by definition. A refactor is required to achieve this. (Linking #2507 to log this)

JCZuurmond (commented):
I will do manual testing for Azure only as I miss an environment to create the uber principal in AWS

src/databricks/labs/ucx/cli.py
JCZuurmond: Taking this concern out of the scope of this PR: https://github.com/databrickslabs/ucx/issues/2644

nfx (approved):
lgtm

PR #2630, opened 2024-09-13
Added task for linting queries

ericvergnaud:
## Changes
Workspace sql queries are not currently linted.
This PR fixes that by introducing a QueryLinter which mimics for dashboard queries what WorkflowLinter does for jobs

### Linked issues
Resolves #2350 

### Functionality
- [x] modified existing workflow: `experimental-workflow-linter`
- [x] added a new table `query_problems`

### Tests
- [x] added unit tests
- [x] added integration tests
- [x] manually tested schema upgrade:

Screenshot 2024-09-13 at 16 20 42

Please note that storage is simplified in PR https://github.com/databrickslabs/ucx/pull/2662

PR #2599, opened 2024-09-12
Collect direct filesystem access from queries

ericvergnaud:
## Changes
Adds support for extracting DirectFileSystemAccess records from workspace queries

### Linked issues
Resolves #2350

### Functionality
- [x] added a new table `directfs_in_queries`
- [x] added a new view `directfs` unioning `directfs_in_paths` with the above

### Tests
- [x] added unit tests
- [x] manually tested schema upgrade:

Screenshot 2024-09-13 at 16 17 22

Integration with ExperimentalWorkflowLinter will be done in a separate PR

PR #2646, opened 2024-09-16
Drop replace methods with builtin

ericvergnaud:
## Changes
Our code implements 'replace' methods on some dataclasses with no benefit over the builtin method.
This PR fixes that.

### Linked issues
None

### Functionality
None

### Tests
- [x] ran unit tests

### DIFF
--- a/src/databricks/labs/ucx/source_code/base.py
+++ b/src/databricks/labs/ucx/source_code/base.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import codecs
+import dataclasses
 import locale
 import logging
 from abc import abstractmethod, ABC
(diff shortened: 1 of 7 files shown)


src/databricks/labs/ucx/source_code/base.py
```diff
@@ -41,25 +42,6 @@ class Advice:
     end_line: int
     end_col: int
 
-    def replace(
```
nfx, line 44: what about directfs accesses?

nfx (changes requested):

src/databricks/labs/ucx/source_code/base.py
ericvergnaud: already done in https://github.com/databrickslabs/ucx/pull/2599 and https://github.com/databrickslabs/ucx/pull/2630

ericvergnaud: what is left re directfs is more focused methods that call dataclasses.replace

nfx (approved):
lgtm

PR #2659, opened 2024-09-17
refactor selection of linter from advice code

ericvergnaud:
## Changes
Our linters generate advices with advice codes that may vary depending on their analysis.
Since fixers are typically implemented as linters (see `FromTableSqlLinter`, `SparkSqlPyLinter`, `SparkTableNamePyLinter`), we need them to be used for any advice code generated during linting (that they can actually fix)
This is not achievable with the current implementation, because the `name` property can only return a single code (plus its purpose is obfuscated).
This PR fixes the above by renaming `name` and having it return multiple values

### Linked issues
None

### Functionality
None

### Tests
- [x] ran unit tests


src/databricks/labs/ucx/source_code/linters/context.py
```diff
@@ -83,14 +83,14 @@ def fixer(self, language: Language, diagnostic_code: str) -> Fixer | None:
         if language not in self._fixers:
             return None
         for fixer in self._fixers[language]:
-            if fixer.name == diagnostic_code:
+            if diagnostic_code in fixer.supported_advice_codes:
```
JCZuurmond, line 86: Diagnostic or advice?

JCZuurmond (commented):
One comment to clarify

src/databricks/labs/ucx/source_code/linters/context.py
ericvergnaud: it's the code found in Advice.code so the latter, maybe we need to also rename the parameter

JCZuurmond (approved):
lgtm

ericvergnaud:
@nfx not strategic, closing as discussed

Issue #2595, opened 2024-09-12
[FEATURE] Run the workflow static analysis as part of the `migration progress` workflow


JCZuurmond:
### Summary

Run the workflow static analysis on this daily migration progress workflow, `experimental-workflow-linter` workflow

PR #2678, opened 2024-09-19
Assess source code as part of the assessment

ericvergnaud:
## Changes
Adds 2 tasks to the assessment workflow:
 - assess source code from dashboards -> widgets -> SQL queries
 - assess source code from workflows -> jobs -> tasks -> notebooks/files

### Linked issues
Progresses #2595 

### Functionality
- [x] modified existing command: `databricks labs install ucx` by linting source code on top of existing assessment

### Tests
- [x] ran integration tests


src/databricks/labs/ucx/runtime.py
```diff
@@ -56,7 +55,6 @@ def all(cls):
                 ScanTablesInMounts(),
                 MigrateTablesInMounts(),
                 PermissionsMigrationAPI(),
-                ExperimentalWorkflowLinter(),
```
nfx, line 59: don't remove this, keep it

nfx (changes requested):

src/databricks/labs/ucx/runtime.py
ericvergnaud: So we'll be linting jobs and dashboards twice ?

nfx: we always need to keep a backup workflow for a release or two

tests/integration/assessment/test_workflows.py
```diff
@@ -5,7 +5,7 @@
 from databricks.sdk.service.iam import PermissionLevel
 
 
-@retried(on=[NotFound, InvalidParameterValue], timeout=timedelta(minutes=8))
+@retried(on=[NotFound, InvalidParameterValue], timeout=timedelta(minutes=25))
```
nfx, line 8: timeout too long

nfx (changes requested):

tests/integration/assessment/test_workflows.py
ericvergnaud: changed to 15. With the linting, 8 minutes almost always fails

ericvergnaud: 15 is not enough, using 20

nfx: same way as we do with schemas, add `include_jobs` / `include_dashboards` into `WorkspaceConfig`, so that it lints only what we want in scope of the integration test. we can't have this test to run longer than 9 minutes.

src/databricks/labs/ucx/runtime.py
ericvergnaud: @nfx using 20mn, it times out... I don't have a solution (other than increasing the timeout to 25mns)

nfx (commented):
if widgets are in the follow-up PR, then lgtm

PR #2679, opened 2024-09-19
Fix DFSA false positives from f-string fragments

ericvergnaud:
## Changes
The current code erroneously reports DFSA from f-string fragments such as `f"/Repos/{thing1}/sdk-{thing2}-{thing3}"`
This PR fixes that

### Linked issues
None

### Functionality
None

### Tests
- [x] added unit tests

Additional PRs will be submitted for other false positive patterns


nfx (approved):
fix `make fmt`

PR #2658, opened 2024-09-17
New workflow: `migration-progress-experimental`

asnare:
## Changes

This PR introduces a new workflow, `migration-progress-experimental`, along with a new `migration-progress` CLI command for triggering it. This new pipeline duplicates a lot (but not all of) the `assessment` pipeline and refreshes the inventory for:

 - Clusters
 - Grants
 - Jobs
 - Pipelines
 - Policies
 - Tables[^1]
 - TableMigrationStatus[^2]
 - UDFs

This job will eventually run on a schedule; this is out-of-scope for this PR.

### Linked issues

Resolves #2574, progresses #2074.

### Functionality

- [x] added relevant user documentation
- [X] added new CLI command: `databricks labs ucx migration-progress`
- [X] added a new workflow: `migration-progress-experimental`

### Tests

- [X] added unit tests
- [X] added integration tests
- [X] verified on staging environment (screenshot attached)

Staging screenshot: 
Successful run of the migration-progress-experimental workflow

[^1]: Differs from `assessment` by using `TablesCrawler`: the Scala-based crawler is about to be replaced by a new implementation, but that's not ready yet.
[^2]: Not part of the `assessment` workflow.

PR #2579, opened 2024-09-11
Added Py4j implementation of tables crawler to retrieve a list of HMS tables in the assessment workflow

pritishpai:
## Changes
Tests, fixes and uses the new FasterTableScanCrawler in the Assessment Job. The feature flag will determine whether the assessment uses the old scala code or the new python (using py4j) for better logging during table scans.

### Linked issues
Fix #2190 

### Functionality
- [x] modified existing workflow: `assesment.crawl_tables` now uses new py4j crawler over the scala one

### Tests
- [x] added integration tests

PR #2684, opened 2024-09-19
Fast table crawler: suggestions/fixes for describing tables

asnare:
## Changes

This PR updates the code in the new Py4J-based table crawler that is used to describe tables. Changes include:

 - Improved logging, including the python backtrace when an error occurs on the JVM side of the bridge.
 - Each table property is now fetched over the bridge on its own line, to help pinpoint the origin of errors that occur.
 - Fixes to unboxing the Scala `Option` values. (We now use a technique which only calls simple Scala methods.)
 - A small optimization when detecting if a table is partitioned: instead of materialising the collection we just check if it's non-empty.
 - The table's `.viewText()` property is properly handled as a Scala `Option`.
 - Explicitly verify that the supplied `catalog` argument is `hive_metastore`: that's the only value that is supported.

### Linked issues

Follows on from #2658, progresses #2579.

### Tests

- [X] manually tested
- [X] updated unit test

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/tables.py
+++ b/src/databricks/labs/ucx/hive_metastore/tables.py
@@ -536,6 +536,7 @@ class FasterTableScanCrawler(CrawlerBase):
         catalog and database.
         """
         full_name = f"{catalog}.{database}.{table}"
+        logger.debug(f"Fetching metadata for table: {full_name}")
         try:
             raw_table = self._external_catalog.getTable(database, table)
             table_format = raw_table.provider().getOrElse(None) or "UNKNOWN"
--- a/src/databricks/labs/ucx/hive_metastore/tables.py
+++ b/src/databricks/labs/ucx/hive_metastore/tables.py
@@ -499,6 +499,10 @@ class FasterTableScanCrawler(CrawlerBase):
         while iterator.hasNext():
             yield iterator.next()
 
+    @staticmethod
+    def _option_as_python(scala_option: typing.Any):
+        return scala_option.get() if scala_option.isDefined() else None
+
     def _all_databases(self) -> list[str]:
         if not self._include_database:
             return list(self._iterator(self._external_catalog.listDatabases()))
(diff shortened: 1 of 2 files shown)


nfx (approved):
lgtm

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -499,6 +499,10 @@ def _iterator(self, result: typing.Any) -> Iterator:
         while iterator.hasNext():
             yield iterator.next()
 
+    @staticmethod
+    def _option_as_python(scala_option: typing.Any):
```
JCZuurmond, line 503: Missing return type? Or we do not have this for scala objects?

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -536,28 +540,35 @@ def _describe(self, catalog, database, table) -> Table | None:
         catalog and database.
         """
         full_name = f"{catalog}.{database}.{table}"
```
JCZuurmond, line 542: Do we need add the backticks here, i.e. safe quoting?

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -536,28 +540,35 @@ def _describe(self, catalog, database, table) -> Table | None:
...
-            )
-        except Exception as e:  # pylint: disable=broad-exception-caught
-            logger.warning(f"Couldn't fetch information for table {full_name} : {e}")
+            is_partitioned = raw_table.partitionColumnNames().iterator().hasNext()
+            object_type = raw_table.tableType().name()
+            view_text = self._option_as_python(raw_table.viewText())
+            table_properties = list(self._iterator(raw_table.properties()))
+            formatted_table_properties = self._format_properties_list(table_properties)
+        except Exception:  # pylint: disable=broad-exception-caught
+            logger.warning(f"Couldn't fetch information for table: {full_name}", exc_info=True)
```
JCZuurmond, line 559: This has my preference since it is more clear which exception info is added to the log message. Though, it does not matter
```suggestion
        except Exception es e:  # pylint: disable=broad-exception-caught
            logger.warning(f"Couldn't fetch information for table: {full_name}", exc_info=e)
```

JCZuurmond (commented):
The approach LGTM, added some nits

src/databricks/labs/ucx/hive_metastore/tables.py
asnare: These are magic proxy objects, indeed, that bridge to JVM objects. I'd add the type but haven't been able to figure out (yet) what it is. In some ways they're the same as mocks: anything goes. Unlike mocks though, we definitely _don't_ have a type that represents the interface of the target JVM objects:

 - `scala_option` is: `scala.Option[T]`
 - The return type is: `T`

src/databricks/labs/ucx/hive_metastore/tables.py
asnare: This is only used for logging.

nfx (approved):
lgtm

PR #2690, opened 2024-09-19
[DOCUMENTATION] Improve documentation when installing UCX from a machine with restricted internet access

JCZuurmond:
Improve documentation when installing UCX from a machine with restricted internet access


nfx (approved):
lgtm

PR #2701, opened 2024-09-23
Harden configuration reading

### LINKED ISSUE #2581: [BUG]: TASK "assess azure sp" fails with 'str' object has no attribute 'get'

### Current Behavior

Assessment task "assess azure sp" fails with 'str' object has no attribute 'get' 

image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Cloud

Azure

### Operating System

Windows

### Version

latest via Databricks CLI

### Relevant log output

_No response_

JCZuurmond:
## Changes
Harden configuration reading by verifying the type before reading the "value" using `.get`

### Linked issues

Resolves #2581 (hopefully the second get is the issue, type hinting should cover that, but who knows)

### Functionality

- [x] modified existing workflow: `assessment`

### Tests

- [x] added unit tests

### DIFF
--- a/src/databricks/labs/ucx/assessment/secrets.py
+++ b/src/databricks/labs/ucx/assessment/secrets.py
@@ -30,7 +30,12 @@ class SecretsMixin:
             )
             return None
 
-    def _get_value_from_config_key(self, config: dict, key: str, get_secret: bool = True) -> str | None:
+    def _get_value_from_config_key(
+        self,
+        config: dict[str, str | dict[str, str]],
+        key: str,
+        get_secret: bool = True,
+    ) -> str | None:
         """Get a config value based on its key, with some special handling:
         If the key is prefixed with spark_conf, i.e. this is in a cluster policy, the actual value is nested
         If the value is of format {{secret_scope/secret}}, we extract that as well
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/assessment/secrets.py
```diff
@@ -30,13 +30,20 @@ def _get_secret_if_exists(self, secret_scope, secret_key) -> str | None:
...
+    ) -> str | None:
         """Get a config value based on its key, with some special handling:
         If the key is prefixed with spark_conf, i.e. this is in a cluster policy, the actual value is nested
         If the value is of format {{secret_scope/secret}}, we extract that as well
         """
         if re.search("spark_conf", key):
-            value = config.get(key, {}).get("value", "")
+            value = config.get(key, {})
+            if isinstance(value, dict):
+                value = value.get("value", "")
```
nfx, line 46: bug: otherwise `value` is `None`. add an else statement

nfx (changes requested):

src/databricks/labs/ucx/assessment/secrets.py
JCZuurmond: That could also happen in line 48 and the function signature allows return type `None`. It will break when `get_secret` is `True`

PR #2741, opened 2024-09-25
Fixes issue of circular dependency of migrate-location ACL

### LINKED ISSUE #2529: [BUG]: Circular dependency for migrate-locations

### Current Behavior

Using the migrate-locations cli command triggers the
AWSExternalLocationsMigration.run method
The method calls
apply_location_acl() method from PrincipalACL
The apply_location_acl() method is dependent on the existing UC External Locations which creates a circular dependency.
We need to redesign the flow to break this loop.

### Expected Behavior

migrate-locations should be able to create external locations and apply principal acls in both AWS/Azuree

### Steps To Reproduce

Run migrate locations in AWS where principal ACLs exist.

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

HariGS-DB:
## Changes

Resolves #2529 

### Functionality

- [ ] modified existing command: `databricks labs ucx migrate-locations`

### Tests

- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -259,7 +259,7 @@ make dev
 
 To use different python version, specify it in `HATCH_PYTHON` variable:
 ```shell
-HATCH_PYTHON=python3.10 make clean dev test
+HATCH_PYTHON="$(which python3.10)" make clean dev test
 ```
 
 Configure your IDE to use `.venv/bin/python` from the virtual environment when developing the project:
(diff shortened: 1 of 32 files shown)


nfx (commented):
did you try fetching location mapping when it's actually needed?
image

nfx (approved):
lgtm

Issue #2572, opened 2024-09-11
[FEATURE] Create a `ucx.history` table


JCZuurmond:
The history table should have the following schema:

| Column name | Data type | Comment |
|---------------|-----------|---------- |
| workspace_id | int | The identifier of the workspace where this record was generated. |
| run_id | int | An identifier of the workflow run that generated this record. |
| run_as | str | The identity of the account that ran the workflow that generated this record. |
| run_start_time | dt.datetime | When this record was generated. |
| snapshot_id | int | An identifier that is unique to the records produced for a given snapshot. |
| ucx_version | str | The UCX semantic version |
| failures | list[str] | The list of problems associated with the object that this inventory record covers. |
| object_type | str | The inventory table for which this record was generated. | 
| object_id | list[str] | The type-specific identifier for this inventory record. |
| object_data | str | Type-specific JSON-encoded data of the inventory record. |
| object_owner | str |The identity that has ownership of the object. |
| object_type_version | int | Versioning of inventory table, for forward compatibility. (Zero `0` if not specified.) |

JCZuurmond:
> *Preferably the object data is stored as a struct, however, since each object type has different attributes this type needs updating when updating one of the ucx dataclasses representing the object OR when adding a new object

@nfx and @asnare : Please give your opinion here

JCZuurmond:
To be determined: when do we create this table? The other tables are created during the `databricks labs install ucx`, i.e. the `install.py` in ucx. For a similar approach, I suggest creating this table also from a command line call and not in a workflow. It could become part of the #2571

asnare:
Some notes on the schema:

 - The `object_type` will correspond to the type of the inventory record. I think we have 2 choices here:

    - We use the name of the inventory table. Examples include: `tables`, `jobs`, `pipelines`, `migration_status`, etc.
    - We use the name of the data class. Corresponding examples include: `Table`, `JobInfo`, `PipelineInfo`, `TableMigrationStatus`, etc.

   My preference is for the former.

 - We're missing an `object_type_version` field, to be used for forward compatibility. An integer that defaults to 0.
 - The `object_id` field will contain a type-specific business key that identifies the inventory record. I'm okay with this being a string, but it could also be an `array<string>`. (Our crawlers don't currently identify their business keys: this is something they need to be updated for.)

[...]

Issue #2600, opened 2024-09-12
[FEATURE] Create a `ucx.workflow_runs` table


JCZuurmond:
### Summary

Create a `ucx.workflow_runs` table to persist UCX workflow runs, e.g. `assessment` or `migration-process`.

### Description

Table schema

| Column name | Data Type | Comment |
| -------------- | ---------- | ---------- |
| `started_at` | `dt.datetime` | The timestamp of the workflow run start. |
| `finished_at` | `dt.datetime` | The timestamp of the workflow run end. | 
| `workspace_id` |  `int` | The workspace id in which the workflow job ran. |
| `workflow_name` | `str` | The workflow name that ran, e.g. `"migration-progress"` or `"assessment"` | 
| `workflow_id` | `int` | The workflow id of the workflow that ran. |
| `workflow_run_id` | `int` |  The workflow run id. |
| `run_as` | `str` | The identity the workflow was run as` |
| `status` | `str` | The workflow status

JCZuurmond:
@nfx  and @asnare : In this table, we want to persist when a crawler did not return any objects to differentiate between a crawler not ran and no objects returned. How do you want to persist this information, i.e. what column schema?

JCZuurmond:
Also, do we need to keep this in the `ucx` catalog? Or should we keep this in the `hive_metastore`?

| | `ucx` | `hive_metastore` |
| Pro | Contains data over multiple workspace scans to which this is relevant. | Available during assessment |
| Con | Growing the new ucx catalog; we might want to be conservative with this. | Assessment should only be ran once, therefore irrelevant as it is only used when rerunning. |

asnare:
When we discussed this at the office the intent was: This will be in the `ucx` catalog and not the `hive_metastore`. (So yes, it's not available during assessment.)

I think a technical driver for this choice is that this table will be updated quite often with small updates, and that will behave better on the `ucx` catalog.

The purpose of the table is also to allow for interpretation of the history table when a crawl produces no records. Without this table we won't be able to handle that situation.

I was going to update the snapshot() functionality of the crawlers to use this table as an optimisation: if the loader returns 0 rows, consult this table if available to determine whether to actually return 0 rows or (as is currently the case) perform a crawl.

ericvergnaud:
Although we run the assessment only once, I suspect that we call snapshot() during migration progress.

JCZuurmond:
[...]

PR #2743, opened 2024-09-25
Introduce an optional history log, where crawler snapshots are journa…

### LINKED ISSUE #2573: [FEATURE] Persist the `migration-progress` crawled objects in the `ucx.history` table
## Description

Persist the `migration-process` crawled objects in the `ucx.history` table (#2572)

asnare:
## Changes

This PR introduces a history table where snapshots are journaled after each crawling operation.

### Linked issues

Progresses #2572, resolves #2573.

### Functionality

- [ ] added a new table

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/framework/history.py
```diff
@@ -0,0 +1,155 @@
...
+            self._object_type_version = getattr(klass, "_ucx_version") if hasattr(klass, "_ucx_version") else 0
+            self._key_from = key_from
+            self._persist = persist
+
+        @cached_property
+        def _workspace_id(self) -> int:
+            return self._ws.get_workspace_id()
+
+        @cached_property
+        def _owner(self) -> str:
```
nfx, line 112: this is not an object owner. it's defined differently for every object type. probably there'll need to be a separate PR to fix that and make it consistent.

src/databricks/labs/ucx/framework/history.py
```diff
@@ -0,0 +1,155 @@
...
+            current_user = self._ws.current_user.me()
+            owner = current_user.user_name or current_user.id
+            assert owner
+            return owner
+
+        def append_snapshot(self, records: Sequence[Record], *, run_start_time: dt.datetime) -> None:
+            # Equivalent entropy to a type-4 UUID.
+            snapshot_id = int.from_bytes(os.urandom(16), byteorder="big")
```
nfx, line 120: don't randomly generate it, inject it from `ctx.named_parameters.get("parent_run_id")`

see https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/runtime.py#L88-L89

src/databricks/labs/ucx/framework/history.py
```diff
@@ -0,0 +1,155 @@
...
+            historical_records = [
+                self._inventory_record_to_historical(record, snapshot_id=snapshot_id, run_start_time=run_start_time)
```
nfx, line 122: @JCZuurmond in the `parse_logs` task, create the `WorkflowRuns` entities, where you record run start time and run end time. otherwise the writer side gets too complicated. `parse_logs` is guaranteed to be executed at the end of every job run, so it's a good place to do so. or add one more task explicitly, which is a bit more explicit.

src/databricks/labs/ucx/framework/history.py
```diff
@@ -0,0 +1,155 @@
...
+            self,
+            record: Record,
+            *,
+            snapshot_id: int,
+            run_start_time: dt.datetime,
+        ) -> HistoricalRecord:
+            object_id = self._key_from(record)
+            object_as_dict = dataclasses.asdict(record)
+            object_as_json = json.dumps(object_as_dict)
+            # TODO: Get failures.
```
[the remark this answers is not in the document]
asnare: This will work for the classes that have failures. (Not all do: 5 of the 9 currently being updated during `migration-progress` have it, 4 do not.)

Aside from these, `ReconResult` has an `error_message` attribute instead of `failures`, but it fulfils the same purpose. This will need to be special-cased.

src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
JCZuurmond, line 1: Why is this required?

src/databricks/labs/ucx/framework/history.py
```diff
@@ -0,0 +1,158 @@
...
+class DataclassInstance(Protocol):
+    __dataclass_fields__: ClassVar[dict]
+    # TODO: Once all record types provide the property: key_fields: ClassVar[Sequence[str]]
+
+
+Record = TypeVar("Record", bound=DataclassInstance)
+
+
+class HistoryLog:
+    __slots__ = ("_ws", "_backend", "_run_id", "_catalog", "_schema", "_table")
```
JCZuurmond, line 61: This would be the first UCX class that have it, why introduce it here?

src/databricks/labs/ucx/framework/history.py
JCZuurmond: https://github.com/databrickslabs/ucx/pull/2754

src/databricks/labs/ucx/source_code/directfs_access.py
```diff
@@ -97,14 +98,14 @@ def __init__(self, backend: SqlBackend, schema: str, table: str):
         """
         super().__init__(backend=backend, catalog="hive_metastore", schema=schema, table=table, klass=DirectFsAccess)
 
-    def dump_all(self, dfsas: Sequence[DirectFsAccess]):
+    def dump_all(self, dfsas: Sequence[DirectFsAccess], crawl_start_time: dt.datetime):
```
JCZuurmond, line 101:
```suggestion
    def dump_all(self, dfsas: Sequence[DirectFsAccess], *, crawl_start_time: dt.datetime) -> None:
```

JCZuurmond (commented):
See comments

src/databricks/labs/ucx/framework/history.py
asnare: We've created #2761 to track this. That PR blocks progress on this one.

src/databricks/labs/ucx/framework/history.py
asnare: This is not the `run_id` from the workflow, which is on a different attribute. Rather it's a unique identifier that is specific to this crawl. We have workflow runs which perform (the same) crawl multiple times: the workflow run id is insufficient to identify a specific crawl.

src/databricks/labs/ucx/framework/crawlers.py
asnare: It lets a class use itself in type annotations, without resorting to strings.

src/databricks/labs/ucx/framework/history.py
asnare: It's not the first, but I agree it's rare. This is just a way of ensuring that attribute mistakes trigger errors instead of relying on ruff/mypy/pylint to flag them.

JCZuurmond: oke, I guess it is good practice, but also kinda "tech debt" for other classes. Could you create a tech debt issue for this? We might not do this, but let's track it

src/databricks/labs/ucx/framework/crawlers.py
JCZuurmond: ah, thanks, did not know that

src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -21,13 +24,22 @@ class DataclassInstance(Protocol):
...
 class CrawlerBase(ABC, Generic[Result]):
-    def __init__(self, backend: SqlBackend, catalog: str, schema: str, table: str, klass: type[Result]):
+    def __init__(
+        self,
+        backend: SqlBackend,
+        catalog: str,
+        schema: str,
+        table: str,
+        klass: type[Result],
+        history_log: HistoryLog | None = None,
```
nfx, line 34: **`... | None` constructor-time dependency in a singleton is an anti-pattern and it has led to undefined behaviors before.** 

If we look at things high level, the dependency diagram looks like

```mermaid
flowchart TD
    history -->|optional!!!| crawler
    crawler --> history_appender
    history --> history_appender
    crawler --> snapshot
    history_appender --> append_snapshot
    append_snapshot --> snapshot
```

and it causes some suboptimal assumptions related to `migration_status` table expressed in https://github.com/databrickslabs/ucx/issues/2600#issuecomment-2383202696.

This can be avoided if dependencies are built in a more explicit way:

```
    @job_task
    def crawl_tables(self, ctx: RuntimeContext) -> None:
        ctx.history_log.update_snapshot(ctx.tables_crawler)

    @job_task
    def crawl_udfs(self, ctx: RuntimeContext) -> None:
       ctx.history_log.update_snapshot(ctx.udfs_crawler)
```

and then `update_snapshot` being something like with clear and non-optional dependencies

```mermaid
flowchart TD
    history --> update_snapshot
    crawler --> update_snapshot

    update_snapshot -.-> _appender
    crawler -.-> _appender
    history -.-> _appender
    crawler -.-> snapshot
    snapshot -.-> _appender

```

```
def update_snapshot(self, crawler: CrawlerBase[T]):
    appender = self._create_appender(crawler.get_klass())
    appender.persist(crawler.snapshot(force_refresh=True))
```

this will fit better with things like `migration_status` and `dfsa_*`

src/databricks/labs/ucx/framework/history.py
nfx: we don't really do this in other places and it doesn't really add to readability. it's not quite consistent with the rest of the codebase.

src/databricks/labs/ucx/framework/history.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def __init__(
+        self,
+        ws: WorkspaceClient,
+        backend: SqlBackend,
+        run_id: int,
+        catalog: str,
+        schema: str,
+        table: str,
```
nfx, line 70:
```suggestion
```

src/databricks/labs/ucx/framework/history.py
```diff
@@ -0,0 +1,158 @@
...
+    ) -> None:
+        self._ws = ws
+        self._backend = backend
+        self._run_id = run_id
+        self._catalog = catalog
+        self._schema = schema
+        self._table = table
```
nfx, line 77:
```suggestion
        self._table = "history"
```

src/databricks/labs/ucx/framework/history.py
```diff
@@ -0,0 +1,158 @@
...
+
+        @cached_property
+        def _owner(self) -> str:
+            current_user = self._ws.current_user.me()
+            owner = current_user.user_name or current_user.id
+            assert owner
+            return owner
+
+        def append_snapshot(self, records: Sequence[Record], *, run_start_time: dt.datetime) -> None:
+            snapshot_id = int.from_bytes(os.urandom(7), byteorder="big")
```
nfx, line 120:
```suggestion
```

rely on workflow run id

src/databricks/labs/ucx/framework/history.py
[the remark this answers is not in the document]
asnare: This implementation is superseded by a new one. Encoding is handled by `HistoricalEncoder`: it treats `failures` as a special property, and does not include it in the object data. If the class doesn't have a `failures` attribute we set to this an empty list (`[]`) because it's a mandatory field.

src/databricks/labs/ucx/framework/history.py
asnare: In the new implementation (`HistoricalEncoder`) the owner property comes from the supplied `Ownership` implementation.

src/databricks/labs/ucx/framework/history.py
```diff
@@ -0,0 +1,155 @@
...
+            owner = current_user.user_name or current_user.id
+            assert owner
+            return owner
+
+        def append_snapshot(self, records: Sequence[Record], *, run_start_time: dt.datetime) -> None:
+            # Equivalent entropy to a type-4 UUID.
+            snapshot_id = int.from_bytes(os.urandom(16), byteorder="big")
+            historical_records = [
+                self._inventory_record_to_historical(record, snapshot_id=snapshot_id, run_start_time=run_start_time)
+                for record in records
```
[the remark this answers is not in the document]
asnare: Not present in revised implementation.

src/databricks/labs/ucx/framework/history.py
asnare: In the revised implementation, `run_id` is supplied and intended to be the job `run_id`.

src/databricks/labs/ucx/framework/crawlers.py
asnare: The implementation has been replaced to fit with this style: instead of being a part of the crawlers, the history mechanism will operate as a wrapper around a select few and controlled explicitly by the relevant workflow tasks.

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,153 @@
...
+from databricks.labs.ucx.framework.utils import escape_sql_identifier
+from databricks.labs.ucx.progress.install import Historical
+
+logger = logging.getLogger(__name__)
+
+
+class DataclassInstance(Protocol):
+    __dataclass_fields__: ClassVar[dict[str, Any]]
+
+    __id_fields__: ClassVar[Sequence[str]]
```
nfx, line 22: verify if this is not messing up with LSQL schema creation - check `deploy_schemas`

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,153 @@
...
+        ownership: Ownership[Record],
+        klass: type[Record],
+        run_id: int,
+        workspace_id: int,
+        catalog: str,
+        schema: str = "ucx",
+        table: str = "history",
+    ) -> None:
+        self._ws = ws
+        self._backend = backend
```
nfx, line 137:
```suggestion
        self._sql_backend = sql_backend
```
nit: consistency

nfx (approved):
lgtm, make sure it's part of the real workflow that should be scheduled to run once per day (at least on weekdays)

src/databricks/labs/ucx/assessment/clusters.py
```diff
@@ -46,6 +47,8 @@ class ClusterInfo:
     creator: str | None = None
     """User-name of the creator of the cluster, if known."""
 
+    __id_attributes__: ClassVar[Sequence[str]] = ("cluster_id",)
```
nfx, line 50: nit: do we really need `ClassVar` annotation in every single class? using a method in interface-on-read/protocol is a more concise option - https://typing.readthedocs.io/en/latest/spec/protocol.html

src/databricks/labs/ucx/contexts/workflow_task.py
```diff
@@ -116,10 +134,109 @@ def workflow_run_recorder(self) -> WorkflowRunRecorder:
...
+    def historical_clusters_log(self) -> HistoryLog[ClusterInfo]:
+        cluster_owner = ClusterOwnership(self.administrator_locator)
+        return HistoryLog(
+            self.sql_backend,
+            cluster_owner,
+            ClusterInfo,
+            int(self.named_parameters["parent_run_id"]),
+            self.workspace_id,
+            self.config.ucx_catalog,
+        )
```
nfx, line 159:
```suggestion
    @cached_property
    def cluster_ownership(self) -> ClusterOwnership:
        return ClusterOwnership(self.administrator_locator)
        
    @cached_property
    def historical_clusters_log(self) -> HistoryLog[ClusterInfo]:
        return HistoryLog(
            self.sql_backend,
            self.cluster_ownership,
            ClusterInfo,
            int(self.named_parameters["parent_run_id"]),
            self.workspace_id,
            self.config.ucx_catalog,
        )
```

we need `cluster_ownership` in context for https://github.com/databrickslabs/ucx/pull/2980. same for other things.

src/databricks/labs/ucx/progress/workflows.py
```diff
@@ -22,26 +22,34 @@ class MigrationProgress(Workflow):
...
-
-    @job_task(job_cluster="tacl")
-    def setup_tacl(self, ctx: RuntimeContext) -> None:
-        """(Optimization) Starts `tacl` job cluster in parallel to crawling tables."""
+        history_log = ctx.historical_udfs_log
+        udfs_snapshot = ctx.udfs_crawler.snapshot(force_refresh=True)
+        history_log.append_inventory_snapshot(udfs_snapshot)
 
-    @job_task(depends_on=[crawl_tables, crawl_udfs], job_cluster="tacl")
+    @job_task(depends_on=[verify_prerequisites, crawl_tables, crawl_udfs], job_cluster="table_migration")
```
nfx, line 52: did you verify that `table_migration` can crawl legacy TACL?

src/databricks/labs/ucx/progress/workflows.py
```diff
@@ -22,26 +22,34 @@ class MigrationProgress(Workflow):
     def __init__(self) -> None:
         super().__init__('migration-progress-experimental')
 
-    @job_task
+    @job_task(job_cluster="table_migration")
```
nfx, line 25: why do we need to use `table_migration` cluster instead of the default `main`?

nfx (changes requested):

src/databricks/labs/ucx/assessment/clusters.py
asnare: We are already using a protocol: https://github.com/databrickslabs/ucx/blob/2e5869d80811ce8a7449b9dc095b7a9423822ee1/src/databricks/labs/ucx/progress/history.py#L21

There are a few options here:

 - Inherit from the protocol. This allows the annotation to be left out (and implied from the base class). This works, except unfortunately the mock SQL backend has a bug that means our entity classes are not allowed to inherit from anything.
 - Leave out the annotation entirely. This should probably work, but mypy fails to recognise that the (class) field has the expected type for the protocol.
 - Just use the inner annotation. This should possibly work, but mypy fails to recognise that the (class) field has the expected type for the protocol.
 - Drop the `ClassVar` thing on the protocol. Aside from being incorrect (I think?) mypy rejects it.

src/databricks/labs/ucx/contexts/workflow_task.py
asnare: All the ownership components have been factored out.

src/databricks/labs/ucx/progress/workflows.py
asnare: Indeed, I was testing this (it's covered by integration tests).

It indeed fails, because the table-migration cluster is using Spark Connect and the fast crawler doesn't have the JVM-bridge into Spark available.

I've added a note on `RuntimeContext.tables_crawler` that as things stand not all runtime contexts support crawling.

src/databricks/labs/ucx/progress/workflows.py
asnare: I'm not sure about this myself: it was added in #2851.

I did notice that the default cluster is not UC-enabled, which means we can't write to the history file from there, whereas the table-migration cluster is UC-enabled.

src/databricks/labs/ucx/progress/workflows.py
asnare: So this created a bit of a problem, for which a workaround is now in place:

 - The TACL cluster, where we can crawl from, is not UC-enabled. So we can't write to the history log.
 - The table-migration cluster, where can write to the log, can't crawl tables.

The workaround for now is to split the table-refresh into 2 tasks:
1. Crawl from the `tacl` cluster, and just update the (HMS-based) inventory.
2. On the `table-migration` cluster, read the current inventory snapshot (assumably from step 1) and update the history log.

It's not perfect, in particular because if crawling produced no records (for whatever reason) then a crawl will be triggered during step 2.

src/databricks/labs/ucx/assessment/clusters.py
```diff
@@ -46,6 +47,8 @@ class ClusterInfo:
     creator: str | None = None
     """User-name of the creator of the cluster, if known."""
 
+    __id_attributes__: ClassVar[tuple[str, ...]] = ("cluster_id",)
```
JCZuurmond, line 50: typically dunder signals Python builtin. I prefer only leading `__` underscore

src/databricks/labs/ucx/contexts/application.py
```diff
@@ -40,15 +40,17 @@
     GrantsCrawler,
     MigrateGrants,
     PrincipalACL,
+    GrantOwnership,
```
JCZuurmond, line 43: Isort

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,216 @@
...
+
+    All attributes must be (non-optional) strings.
+    """
+
+
+Record = TypeVar("Record", bound=DataclassWithIdAttributes)
+
+
+class HistoricalEncoder(Generic[Record]):
+    _job_run_id: int
```
JCZuurmond, line 32: Why make these class attributes?

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,216 @@
...
+        Arguments:
+            klass: The record type.
+        Returns:
+            A tuple containing:
+                - A dictionary of fields to include in the object data, and their type.
+                - The type of the failures field, if present.
+        """
+        field_types = {field.name: field.type for field in dataclasses.fields(klass)}
+        failures_type = field_types.pop("failures", None)
+        if failures_type not in (None, str, list[str]):
```
JCZuurmond, line 80: I think we should not match `None` like that`:

```suggestion
        if not (failures_type is None or failures_type in {str, list[str]}:
```

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,216 @@
...
+    """The names of attributes (can be dataclass fields or ordinary properties) that make up the object identifier.
+
+    All attributes must be (non-optional) strings.
+    """
+
+
+Record = TypeVar("Record", bound=DataclassWithIdAttributes)
+
+
+class HistoricalEncoder(Generic[Record]):
```
JCZuurmond, line 31: Could you add a docstring on what this class does and why it is important/required?
 
 Important because: we cannot simply convert records to historical data by `json.dumps(dataclasses.asdict(record)))` because we:
 - Need to handle `failures` separately
 - ...

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,216 @@
...
+            job_run_id=self._job_run_id,
+            object_type=self._object_type,
+            object_id=self._object_id(record),
+            data=data,
+            failures=failures,
+            owner=self._ownership.owner_of(record),
+        )
+
+
+class HistoryLog(Generic[Record]):
```
JCZuurmond, line 188: A class is typically named as a persona:

```suggestion
class HistoryLogger(Generic[Record]):
```

Note: I am not sure about the `Logger` as it sounds like Python logging stuff. `Recorder` is maybe more accurate

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,216 @@
...
+
+    Attributes can be either a dataclass field or a property, for which the type must be a (non-optional) string.
+    """
+
+    def __init__(self, job_run_id: int, workspace_id: int, ownership: Ownership[Record], klass: type[Record]) -> None:
+        self._job_run_id = job_run_id
+        self._workspace_id = workspace_id
+        self._ownership = ownership
+        self._object_type = self._get_object_type(klass)
+        self._field_types, self._has_failures = self._get_field_types(klass)
```
JCZuurmond, line 61: `self._field_types` suggests a list of types to me. Maybe introduce as small data class or named tuple `Field` with `(name, type)`

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,216 @@
...
+                msg = f"Timestamp without timezone not supported in or within field {name}: {value}"
+                raise ValueError(msg)
+            # Always store with 'Z'.
+            ts_utc = value.astimezone(dt.timezone.utc)
+            return ts_utc.isoformat().replace("+00:00", "Z")
+
+        msg = f"Cannot encode {type(value)} value in or within field {name}: {value!r}"
+        raise TypeError(msg)
+
+    def _encode_field_value(self, name: str, value: Any | None) -> str | None:
```
JCZuurmond, line 141:
```suggestion
    def _encode_field_value(self, name: str, value: Any) -> str | None:
```

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,216 @@
...
+        if value is None:
+            return None
+        value_type = self._field_types[name]
+        if value_type in (str, (str | None)):
+            return value
```
JCZuurmond, line 146: To be sure?

```suggestion
            return str(value)
```

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,216 @@
...
+        encoded_value = json.dumps(
+            value,
+            allow_nan=False,
+            separators=(",", ":"),
```
JCZuurmond, line 150: Add a comment why this: no space behind default separator. (I had to Google this)

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,216 @@
...
+        return [getattr(record, field) for field in self._id_attribute_names]
+
+    @classmethod
+    def _encode_non_serializable(cls, name: str, value: Any) -> Any:
+        if isinstance(value, dt.datetime):
+            # Only allow tz-aware timestamps.
+            if value.tzinfo is None:
+                # Name refers to the outermost field, not necessarily a field on a (nested) dataclass.
+                msg = f"Timestamp without timezone not supported in or within field {name}: {value}"
+                raise ValueError(msg)
```
JCZuurmond, line 133: Should raise `TypeError` according to Json dumps doc

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,216 @@
...
+
+    @classmethod
+    def _as_dict(cls, record: Record) -> dict[str, Any]:
+        return dataclasses.asdict(record)
+
+    def _object_id(self, record: Record) -> list[str]:
+        return [getattr(record, field) for field in self._id_attribute_names]
+
+    @classmethod
+    def _encode_non_serializable(cls, name: str, value: Any) -> Any:
```
JCZuurmond, line 127: Wouldn't the `name` be coverred by `json.dumps`?

src/databricks/labs/ucx/progress/workflows.py
```diff
@@ -23,26 +23,49 @@ class MigrationProgress(Workflow):
...
+        """(Optimization) Allow the TACL job cluster to be started while we're verifying the prerequisites for
+        refreshing everything."""
+
+    @job_task(depends_on=[verify_prerequisites, setup_tacl], job_cluster="tacl")
     def crawl_tables(self, ctx: RuntimeContext) -> None:
         """Iterates over all tables in the Hive Metastore of the current workspace and persists their metadata, such
         as _database name_, _table name_, _table type_, _table location_, etc., in the table named
         `$inventory_database.tables`. The metadata stored is then used in the subsequent tasks and workflows to, for
         example, find all Hive Metastore tables that cannot easily be migrated to Unity Catalog."""
+        # The TACL cluster is not UC-enabled, so the snapshot cannot be written immediately to the history log.
```
JCZuurmond, line 45: Good comment!

src/databricks/labs/ucx/progress/workflows.py
```diff
@@ -23,26 +23,49 @@ class MigrationProgress(Workflow):
...
+        # Step 1 of 2: Just refresh the inventory.
         ctx.tables_crawler.snapshot(force_refresh=True)
 
-    @job_task
+    @job_task(depends_on=[verify_prerequisites, crawl_tables], job_cluster="table_migration")
+    def update_tables_history_log(self, ctx: RuntimeContext) -> None:
+        """Update the history log with the latest tables inventory snapshot."""
+        # The table migration cluster is not legacy-ACL enabled, so we can't crawl from here.
+        # Step 2 of 2: Assuming the inventory was refreshed, capture into the history log.
```
JCZuurmond, line 53: "assuming" : given its `depends_on`

JCZuurmond (commented):
Herewith my review of the code, have yet to go through the tests

src/databricks/labs/ucx/progress/history.py
asnare: This is one way to annotate the types of instance variables, although as a convenience we often do it during assignment from the initialiser (or even more commonly omit it entirely, leaving it to mypy to infer from the assigned value).

In this case I wanted to attach docstrings and this is the way to do that.

src/databricks/labs/ucx/progress/history.py
asnare: It turns out there's a subtle bug in there, which I've fixed.

src/databricks/labs/ucx/progress/history.py
asnare: Excellent suggestion. I've added this, does it seem right to you?

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+class HistoricalEncoder(Generic[Record]):
+    """An encoder for dataclasses that will be stored in our history log.
+
+    Our history records are designed with several goals in mind:
+     - Records can be of different types, meaning we have to store heterogenous types using a homogenous schema.
+     - We partially shred the records to allow for easier SQL-based querying.
+     - Flexibility from the start, because schema changes in the future will be very difficult.
+     - Records will be shared (and queried) across workspaces with different versions of UCX installed.
```
JCZuurmond, line 38: @nfx : Suggested that we need to enforce the same UCX version across workspaces (somehow)

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,240 @@
...
+
+    With this in mind:
+     - We have an 'object-type' (discriminator) field for holding the type of the record we're storing. This allows for
```
JCZuurmond, line 41:
```suggestion
     - We have an `object_type` (discriminator) field for holding the type of the record we're storing. This allows for
```

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,240 @@
...
+       type-specific logic when querying.
+     - We have a (type-specific) field for storing the business key (identifier) of each record.
+     - We have special handling of the 'failures' attribute that is normally present in the types we are storing.
```
JCZuurmond, line 44:
```suggestion
     - We have special handling of the 'failures' attribute that is sometimes present in the types we are storing.
```

Note on the failures: it is present when we "assess" objects not when we "crawl" objects. At least, that is the common discriminator. In practice, we are not so strict about this

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,240 @@
...
+     - The object-data is an unconstrained key-value map, allowing arbitrary attributes to be stored and (if top-level)
+       queried directly. Complex values are JSON-encoded.
+     - The ownership mechanism is used to associate each record with a user.
```
JCZuurmond, line 47: Link

src/databricks/labs/ucx/progress/history.py
```diff
@@ -0,0 +1,240 @@
...
+     - To help with forward and backward compatibility the UCX version used to encode a record is included in each
+       record.
+     - The associated workspace and job run is also included in each record.
```
JCZuurmond, line 50: Mention cross-workspace and progress

src/databricks/labs/ucx/progress/history.py
JCZuurmond: It looks good! I have some pointers, but general looks good

JCZuurmond (commented):
`HistoricalEncoder` docstring review

src/databricks/labs/ucx/progress/history.py
asnare: I've renamed it… does that help?

src/databricks/labs/ucx/progress/history.py
asnare: I need a bit more context on that… can you point me to this?

At the very least any enforcement needs to be relaxed enough to allow for upgrades.

src/databricks/labs/ucx/progress/history.py
asnare: I've updated the wording here.

src/databricks/labs/ucx/progress/history.py
asnare: I've added a sphinx-style cross-reference. (IntelliJ/PyCharm recognise it, I'm not sure if we have something else we should be using?)

src/databricks/labs/ucx/progress/history.py
asnare: This is the top-level name, and is used as the key in the `object_data` map. These are strings, so they don't need to be JSON-encoded.

It's passed through to here to: a) allow for use in the exception message; b) if we ever need to subclass/override, this allows field-specific encoding.

Nested names (within `value`) are indeed part of the JSON-encoding.

src/databricks/labs/ucx/progress/history.py
asnare: How strongly do you feel about this?

The wording of the `json.dumps()` [documentation](https://docs.python.org/3.10/library/json.html#json.dump) states that in isolation, but I believe the intent is really:

 - `TypeError` if the _type_ of the value cannot be handled.
 - `ValueError` if the type is fine, but there's a problem with the value itself. (See  the `allow_nan` for the equivalent situation with `float`).

(The internal implementation has _no_ handling for exceptions: it simply propagates. For what it's worth, this code is explicitly covered by unit tests.)

src/databricks/labs/ucx/progress/history.py
asnare: That's a good point.

The style of this encoded is to use the declared types (from the hints, checked during initialisation) to drive encoding. But there was no check that the value during encoding matches the declared type.

I've added this check, and decided to handle this situation by raising an exception (failing early). The type-checker should be catching situations where the value doesn't match the type-hint, but to be safe I'd rather we fail early (and hopefully catch this during CI) rather than have junk values end up in history tables.

What do you think?

src/databricks/labs/ucx/assessment/clusters.py
asnare: We can't use just a leading `__`: that makes it a private attribute, and it can't be accessed from outside the instance. (Dunder attributes are special: although they start with `__` that are not subject to mangling like ordinary fields. Arguably they signal a protocol in use behind the curtains, which indeed is what Python mainly uses them for.)

In this case alternatives are:

 - `_id_attributes: ClassVar[tuple[str, ...]]`: Protected, mildly cranky linter ("access to protected member of a client class").
 - `id_attributes: ClassVar[tuple[str, ...]]`: Public, which doesn't feel right for what is essentially metadata that we're trying to attach to the instance.

A bit further from home, is something like the SQLAlchemy approach where metadata (such as this) is placed in a sibling-class. That's a pretty large hammer to hit this particular nail though.

Any further thoughts?

nfx: this confusion could have been avoided by using a method-based protocol...

nfx (approved):
let's fix forward

src/databricks/labs/ucx/assessment/clusters.py
JCZuurmond: Let's stick with this for now

src/databricks/labs/ucx/progress/history.py
JCZuurmond: sure, different to how we do it in the remainder code base, but fine

src/databricks/labs/ucx/progress/history.py
JCZuurmond: Let's move this conversation out of this PR https://github.com/databrickslabs/ucx/issues/3043

JCZuurmond (approved):

PR #2744, opened 2024-09-25
Add UCX history schema and table for storing UCX's artifact

### LINKED ISSUE #2572: [FEATURE] Create a `ucx.history` table

JCZuurmond:
## Changes
Add UCX history schema and table for storing UCX's artifact. The artifacts are, amongst other things, going to be used for migration progress tracking.

### Linked issues

Resolves #2572 

### Functionality

- [x] modified existing command: `databricks labs ucx create-ucx-catalog`

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests


JCZuurmond (commented):
@nfx and @asnare : Please provide input. Especially on the dataclass attributes and the naming

JCZuurmond:
Missing feature on lsql to reuse `SchemaDeployer` here: https://github.com/databrickslabs/lsql/issues/294

asnare:
> @nfx and @asnare : Please provide input. Especially on the dataclass attributes and the naming

This is the current version that I'm using (as an intermediate version) on #2743:
```python
@dataclass(frozen=True, kw_only=True)
class HistoricalRecord:
    workspace_id: int
    """The identifier of the workspace where this record was generated."""

    run_id: str
    """An identifier of the workflow run that generated this record."""

    snapshot_id: str
    """An identifier that is unique to the records produced for a given snapshot."""

    run_start_time: dt.datetime
    """When this record was generated."""

    object_type: str
    """The inventory table for which this record was generated."""

    object_type_version: int
    """Versioning of inventory table, for forward compatibility."""

    object_id: list[str]
    """The type-specific identifier for this inventory record."""

    object_data: str
    """Type-specific JSON-encoded data of the inventory record."""

    failures: list[str]
    """The list of problems associated with the object that this inventory record covers."""

    owner: str
    """The identity of the account that created this inventory record."""
```

asnare:
I've also added some notes on the feature issue: https://github.com/databrickslabs/ucx/issues/2572#issuecomment-2371635046 &
https://github.com/databrickslabs/ucx/issues/2572#issuecomment-2373443999

JCZuurmond:
Both `_id

> > @nfx and @asnare : Please provide input. Especially on the dataclass attributes and the naming
> 
> This is the current version that I'm using (as an intermediate version) on #2743:
> 
> ```python
> @dataclass(frozen=True, kw_only=True)
> class HistoricalRecord:
>     workspace_id: int
>     """The identifier of the workspace where this record was generated."""
> 
>     run_id: str
>     """An identifier of the workflow run that generated this record."""
> 
>     snapshot_id: str
>     """An identifier that is unique to the records produced for a given snapshot."""
> 
>     run_start_time: dt.datetime
>     """When this record was generated."""
> 
>     object_type: str
>     """The inventory table for which this record was generated."""
> 
>     object_type_version: int
>     """Versioning of inventory table, for forward compatibility."""
> 
>     object_id: list[str]
>     """The type-specific identifier for this inventory record."""
> 
>     object_data: str
>     """Type-specific JSON-encoded data of the inventory record."""
> 
>     failures: list[str]
>     """The list of problems associated with the object that this inventory record covers."""
> 
>     owner: str
>     """The identity of the account that created this inventory record."""
> ```

Thanks! I have update the data class: https://github.com/databrickslabs/ucx/pull/2744/commits/e80556c046218f61eed42b51dd46c135b4f8029d

Note:
- The `run_id` and `snapshot_id` are `int`. Do you know if they should be `str` instead?
- On `owner` see: https://github.com/databrickslabs/ucx/issues/2572#issuecomment-2373851457

asnare:
> * The `run_id` and `snapshot_id` are `int`. Do you know if they should be `str` instead?

`run_id`: Can definitely be an integer.

`snapshot_id`: I was populating this with a UUID, but I'm also fine with it being an int.

src/databricks/labs/ucx/progress/install.py
```diff
@@ -0,0 +1,75 @@
...
+    object_type: str
+    """The inventory table for which this record was generated."""
+
+    object_type_version: int
+    """Versioning of inventory table, for forward compatibility."""
+
+    object_id: list[str]
+    """The type-specific identifier for this inventory record."""
+
+    object_data: str
```
nfx, line 44:
```suggestion
    data: dict[str,str]
```
we'll impose a constraint on the writer, so that readers are easier.

src/databricks/labs/ucx/progress/install.py
```diff
@@ -0,0 +1,75 @@
...
+    """Type-specific JSON-encoded data of the inventory record."""
+
+    object_owner: str
```
nfx, line 47:
```suggestion
    owner: str
```
is it user ID or user name? 
https://databricks-sdk-py.readthedocs.io/en/latest/dbdataclasses/iam.html#databricks.sdk.service.iam.User.id

src/databricks/labs/ucx/progress/install.py
```diff
@@ -0,0 +1,81 @@
...
+from databricks.sdk.errors import InternalError
+from databricks.sdk.retries import retried
+
+from databricks.labs.ucx.framework.utils import escape_sql_identifier
+
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, kw_only=True)
```
nfx, line 15: let's separate job runs and records

```python
@dataclass(frozen=True, kw_only=True)
class WorkflowRun:
    started_at: dt.datetime
    finished_at: dt.datetime
    workspace_id: int
    workflow_name: str # migration-progress, assessment, etc, etc
    workflow_id: int
    workflow_run_id: int # <=== this column to be used for joining with historical records, create a view to pre-join these. i think you called it "snapshot" or something.
    run_as: str
    status: str
```

essentially, it'll be

```
WITH level0 AS (
  SELECT workspace_id, workflow_id, MAX(finished_at) AS finished_at FROM workflow_runs GROUP BY 1,2
), level1 AS (
  SELECT * FROM level0 LEFT JOIN workflow_runs USING (workspace_id, workflow_id, finished_at)
)
SELECT * FROM level1
```

src/databricks/labs/ucx/progress/install.py
nfx: also add application state migration

src/databricks/labs/ucx/contexts/workspace_cli.py
```diff
@@ -179,6 +180,10 @@ def iam_role_creation(self):
     def notebook_loader(self) -> NotebookLoader:
         return NotebookLoader()
 
+    @cached_property
+    def history_installation(self) -> HistoryInstallation:
```
nfx, line 184: not sure if it's the best name.

src/databricks/labs/ucx/progress/install.py
```diff
@@ -0,0 +1,75 @@
...
+    run_start_time: dt.datetime
+    """When this record was generated."""
+
+    object_type: str
+    """The inventory table for which this record was generated."""
+
+    object_type_version: int
+    """Versioning of inventory table, for forward compatibility."""
+
+    object_id: list[str]
```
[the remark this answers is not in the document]
asnare: We have a few things where the object_id is a composite from a few fields. My initial thinking here is that just using them as the `object_id` side-steps having to deal with corner cases like "what happens when I join these fields to make a string but the fields can contain the separator?".

I'm going through the various crawler record types now to see what this looks like in practice.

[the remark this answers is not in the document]
nfx: please be concrete about what things do require this.

src/databricks/labs/ucx/progress/install.py
nfx: `klazz(**object_data)` works just fine.

src/databricks/labs/ucx/progress/install.py
```diff
@@ -0,0 +1,75 @@
...
+    snapshot_id: int
+    """An identifier that is unique to the records produced for a given snapshot."""
+
+    run_start_time: dt.datetime
+    """When this record was generated."""
+
+    object_type: str
+    """The inventory table for which this record was generated."""
+
+    object_type_version: int
```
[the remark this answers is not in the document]
nfx: we can 
```
version_map[tuple[int,int,int], int] = {
   (0,37,0): 1,
   (0,39,0): 2,
   (0,52,0): 3,
}
```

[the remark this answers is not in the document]
asnare: Does this handle future versions, including later versions installed in other workspaces attached to the same metastore?

If we (a reader) are on version `0,54,0` then we'll assume version 3 (which is fine) but break if `0,55,0` introduced version 4 because that's indistinguishable to us: we will treat it was 3 (and fail) because only `0,55,0` or later knows version 4 exists.

Just encoding the integer avoids this because we can detect a future record (and that it cannot be decoded by the current version).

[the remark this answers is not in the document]
nfx: actually, better option would be:

```
migrate_after_versions[tuple[int,int,int], Callable[[dict[str,str]], dict[str,str]] = {
   (0,37,0): lambda x: x | {'foo': "bar"},
   (0,39,0): remove_x_column,
   (0,52,0): convert_some_datatype,
}
```

we sort by version tuple and get a diff between `from databricks.labs.ucx.__about__ import __version__` and apply those. if we get a version that we don't understand - we warn and skip.

src/databricks/labs/ucx/progress/install.py
JCZuurmond: > is it user ID or user name?

What is more consistent incl. programmatic users? I prefer user name, but if that does not work well for SP and 
IAM, then id

JCZuurmond: > also add application state migration

What do you refer to with this?

src/databricks/labs/ucx/progress/install.py
JCZuurmond: That is a separate issue: https://github.com/databrickslabs/ucx/issues/2600

src/databricks/labs/ucx/contexts/workspace_cli.py
JCZuurmond: Me neither, I copied the "Installation" part from other objects, like `WorkspaceInstallation` or `AccountInstallation`, then added `run` method to run the installation.

I actually prefer the word "artifacts", which is not very descriptive and some of the objects in the history table are not the "byproduct" of UCX but the "main product". On the other hand, it describes what we are doing here: installing the resources for UCX artifacts (like scan runs, errors and the history table).

What do you think?

src/databricks/labs/ucx/progress/install.py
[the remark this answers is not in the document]
asnare: Just to aid my own understanding, and try to work through ambiguities, as far as I can tell there are two ways of interpreting a version tied to the UCX release:

 - The records contain the _first version_ that understands the serialised form. That is, version 0.37.0, 0.38.0, etc. write 0.37.0 in the record, until version 0.39.0 at which point version 0.39.0, 0.40, etc. all start writing 0.39.0. This continues until 0.52.0, etc.

   Under this protocol we can detect future/unknown versions because they're higher than `__about__.__version__`. We can also read past versions. It does however have a fairly large caveat: for all data classes we need to know the earliest version at which the 'current' version was introduced. This needs to be known in advance of the software release while we're writing and testing it and hard-coded somewhere. (We also need something special during development to ensure that we can read our own records.)

 - The records contain the _actual version_ that wrote the record.

   Under this protocol we can detect past versions, but not handle future/unknown versions because there's no way to know if a future release introduced a new version in some way. For safety we could just always ignore (and warn on) versions later than `__about__.__version__`. The main caveat here relates to when a downgrade happens. Mixed installs with a shared metastore could end up quite noisy (if not filtering out records from other workspaces).

@nfx: Is this also your understanding?

src/databricks/labs/ucx/progress/install.py
nfx: then we need that one land `main` first, as history needs `workflow_run_id` to join on. other PR struggles with `started_at` and `finished_at` fields. same for `run_as`, actually

src/databricks/labs/ucx/progress/install.py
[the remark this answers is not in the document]
asnare: We have the following classes that use composite keys:
 * `Grant`: `object_type`, `object_key`, `action_type`, `principal`
 * `Permissions`: `object_type`, `object_id`
 * `ReconResult`: `src_schema`, `src_table`
 * `Table`: `catalog`, `database`, `name`
 * `TableMigrationStatus`: `src_schema`, `src_table`
 * `TableSize`: `catalog`, `database`, `name`
 * `Udf`: `catalog`, `database`, `name`

(There's also the special case of `SubmitRunInfo` whose key, `run_ids`, is already a pseudo-composite key because it contains a JSON-encoded array of run identifiers.)

If we are going to just use `str` then I propose that the encoding is just the JSON-encoding of these fields as an array.

An alternative would be to use `map[str,str]`.

If we want to stick with `str` but not use JSON-encoding then we will need to come up with our composite encoding scheme, bearing in mind that the field values aren't under our control and an escaping mechanism will be needed.

src/databricks/labs/ucx/progress/install.py
asnare: We haven't covered what the values in the map should contain.

Until recently our inventory classes were 'flat', and a common workaround is to have a field that encodes structured data:
 * Of the 19 inventory classes, 7 have a `failures` field which is a JSON-encoded array.[^1]
 * `Permissions` has a generic `raw` field that contains arbitrary JSON-encoded data.
 * `ReconResult` has `schema_comparison` and `data_comparison`: also JSON-encoded.
 * `SubmitRunInfo` has JSON-encoded `run_ids` (also mentioned elsewhere in the discussion object identifiers).
 * `Table` has `storage_properties`, which is a custom-encoded `dict[str,str]`.

The recent exception to this is `DirectFsAccess`, which is the first non-flat data inventory class using the recently introduced support in LSQL for complex data types.

With this in mind, there are a few options for what the values in the `object_data` map should be:
 * Blindly JSON-encode it. This will work, and is simple. The main caveats are that: string values will be quoted, and the JSON-encoded fields will end up double-encoded.
 * As above, but: if the field type is string, store as-is. This eliminates unnecessary encoding of all the string fields. The JSON-encoded fields remain as such.

@nfx: How does this sound? Does anything else come to mind?

[^1]: Elsewhere we propose removing this field from classes that have it and storing this directly on the `failures` attribute.

src/databricks/labs/ucx/progress/install.py
asnare: I've also just surveyed all our inventory classes. Of the 19 classes, only 5 currently have a creator field:
 * `ClusterInfo`: `creator`
 * `GlobalInitScriptInfo`: `script_id`
 * `JobInfo`: `creator`
 * `PipelineInfo`: `creator_name`
 * `PolicyInfo`: `creator`

All of these are optional.

asnare: Following some discussion, we've concluded that all classes that end up in the history table need to have an associated owner. This will require class-specific logic. When there is no logical (or available) owner, the workspace admin should be designated as owner.

src/databricks/labs/ucx/progress/install.py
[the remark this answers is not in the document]
asnare: Following discussion, #2761 has been created. The broad outcome of that discussion was that:
 - Every object needs to have an identifiable owner. This is mandatory.
 - Often the owner is indeed a composite of other attributes.
 - Encoding/decoding the owner will be type-specific.
 - Therefore `str` is fine.

src/databricks/labs/ucx/progress/install.py
asnare: Note: #2761 has been created to track the owner logic.

src/databricks/labs/ucx/progress/install.py
JCZuurmond: https://github.com/databrickslabs/ucx/pull/2754

src/databricks/labs/ucx/progress/install.py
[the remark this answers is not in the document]
JCZuurmond: @nfx: What do you think?

src/databricks/labs/ucx/progress/install.py
JCZuurmond: @nfx: Please respond to the above

src/databricks/labs/ucx/progress/install.py
asnare: Owner logic has landed in #2774.

src/databricks/labs/ucx/progress/install.py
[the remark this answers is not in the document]
nfx: okay, ~7 cases our of ~21 crawlers (33%) is a good share to have `object_id: list[str]`.

src/databricks/labs/ucx/progress/install.py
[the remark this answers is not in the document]
nfx: > The records contain the actual version that wrote the record.

this is the intended way. let's store the version as `ucx_version: list[int]` and source it from `__about__.__version__` and use `databricks.sdk.mixins.compute.SemVer.parse` to parse it.

src/databricks/labs/ucx/progress/install.py
nfx: 1) `failures` has to become a top-level field, because we'll perform many different SQL operations on it (grouping, counting, distinct, etc) and it shouldn't get to `data: dict[str,str]` field
2) we don't sync `Permissions`, `ReconResult`, and `SubmitRunInfo` to multiworkspace database
3) to simplify things, let's not put `storage_properties` as we're not actually relying on it for anything.

to get to type safety, each dataclass instance that needs syncing has to get an explicit method

```
def to_historical(self) -> HistoricalRecord:
  return HistoricalRecord(...)
```
and some of the classes, like `Table`, would need a `from_historical` class method:

```
@classmethod
def from_historical(cls, h: HistoricalRecord):
  ...
```

src/databricks/labs/ucx/progress/install.py
nfx: owner is tracked separately

nfx: i meant "populating the owner" was tracked separately

src/databricks/labs/ucx/progress/install.py
asnare: @nfx: That's clear.

We still need to define the semantics of `object_data` values to cover the `source_lineage` field of the `SourceInfo` subclasses: `DirectFsAccess` and `UsedTable`.

Options are:

 - JSON encoded, as suggested above.
 - Ignored (like `failures`)

Or maybe something else…?

src/databricks/labs/ucx/progress/install.py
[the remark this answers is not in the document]
asnare: That's clear, thanks.

src/databricks/labs/ucx/progress/install.py
```diff
@@ -1,12 +1,49 @@
...
+    """The identity that has ownership of the object."""
+
+    ucx_version: str = __version__
+    """The UCX semantic version."""
+
+    snapshot_id: str | None = None
+    """An identifier that is unique to the historical records produced for a given snapshot."""
+
+    object_type_version: int = 0
+    """Versioning of inventory table, for forward compatibility."""
```
nfx, line 44:
```suggestion
```
we don't need these two

nfx (changes requested):

nfx (approved):
lgtm

Issue #2737, opened 2024-09-24
[FEATURE]: Add `create-ucx-catalog` to `assign-metastore` command


JCZuurmond:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Add `create-ucx-catalog` to `assign-metastore` command so that UCX catalog is setup when attaching the metastore.

Note: The [this PR](https://github.com/databrickslabs/ucx/pull/2694) did an attempt of adding the `create-ucx-catalog` to the `assign-metastore, however, [some redsign](#2507) is required to run workspace level commands within the `AccountContext`

### Proposed Solution

See above

FastLee:
Are we sure we want to do it at the same occasion. The challenge is that we recommend against setting a default storage location at the metastore, which means that we will have to provide a managed location at the catalog level. This will complicate an otherwise simple task.

Issue #2577, opened 2024-09-11
[FEATURE] Skip the `migration-progress` run when the `ucx` catalog does not exists


JCZuurmond:
### Summary

Skip the `migration-process` run when the `ucx` catalog does not exists.

### Implementation

At the start of the `migration-process` workflow:
- Check if the `ucx` catalog exists
- If not, stop the workflow
- If so, continue the workflow

JCZuurmond:
See `verification.py` for inspiration: https://github.com/databrickslabs/ucx/blob/6a6f2f4dd2f3c37878dd3596a0ed202859afa650/src/databricks/labs/ucx/hive_metastore/verification.py#L9

PR #2754, opened 2024-09-26
Add and populate UCX `workflow_runs` table

### LINKED ISSUE #2600: [FEATURE] Create a `ucx.workflow_runs` table

JCZuurmond:
## Changes
Add and populate workflow runs table

### Linked issues

Resolves #2600 

### Functionality

- [x] added relevant user documentation
- [x] modified existing workflow: `migration-process-experimental`

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests

### TODO
- [ ] Handle concurrent writes, [see](https://github.com/databrickslabs/ucx/pull/2754#discussion_r1778259805)
- [x] Decide on getting workflow run status from `parse_log_task` --> only add it to the migration progress workflow for now

### DIFF
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -611,7 +611,7 @@ def create_ucx_catalog(w: WorkspaceClient, prompts: Prompts, ctx: WorkspaceConte
     """
     workspace_context = ctx or WorkspaceContext(w)
     workspace_context.catalog_schema.create_ucx_catalog(prompts)
-    workspace_context.history_installation.run()
+    workspace_context.progress_tracking_installer.run()
 
 
 @ucx.command
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -611,6 +611,7 @@ def create_ucx_catalog(w: WorkspaceClient, prompts: Prompts, ctx: WorkspaceConte
     """
     workspace_context = ctx or WorkspaceContext(w)
     workspace_context.catalog_schema.create_ucx_catalog(prompts)
+    workspace_context.history_installation.run()
 
 
 @ucx.command
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -611,7 +611,7 @@ def create_ucx_catalog(w: WorkspaceClient, prompts: Prompts, ctx: WorkspaceConte
     """
     workspace_context = ctx or WorkspaceContext(w)
     workspace_context.catalog_schema.create_ucx_catalog(prompts)
-    workspace_context.progress_tracking_installer.run()
+    workspace_context.progress_tracking_installation.run()
 
 
 @ucx.command
--- a/src/databricks/labs/ucx/contexts/workflow_task.py
+++ b/src/databricks/labs/ucx/contexts/workflow_task.py
@@ -120,4 +120,5 @@ class RuntimeContext(GlobalContext):
             workflow_id=int(self.named_parameters["job_id"]),
             workflow_run_id=int(self.named_parameters["parent_run_id"]),
             workflow_run_attempt=int(self.named_parameters.get("attempt", 0)),
+            workflow_start_time=self.named_parameters["workflow_start_time"],
         )
--- a/src/databricks/labs/ucx/contexts/workflow_task.py
+++ b/src/databricks/labs/ucx/contexts/workflow_task.py
@@ -116,7 +116,6 @@ class RuntimeContext(GlobalContext):
         return WorkflowRunRecorder(
             self.workspace_client,
             self.sql_backend,
-            workflow_name=self.named_parameters["job_name"],
             workflow_id=int(self.named_parameters["job_id"]),
             workflow_run_id=int(self.named_parameters["parent_run_id"]),
             workflow_run_attempt=int(self.named_parameters.get("attempt", 0)),
--- a/src/databricks/labs/ucx/contexts/workflow_task.py
+++ b/src/databricks/labs/ucx/contexts/workflow_task.py
@@ -16,6 +16,7 @@ from databricks.labs.ucx.hive_metastore import TablesInMounts
 from databricks.labs.ucx.hive_metastore.table_size import TableSizeCrawler
 from databricks.labs.ucx.hive_metastore.tables import FasterTableScanCrawler
 from databricks.labs.ucx.installer.logs import TaskRunWarningRecorder
+from databricks.labs.ucx.progress.workflow_runs import WorkflowRunRecorder
 
 
 class RuntimeContext(GlobalContext):
@@ -108,4 +109,15 @@ class RuntimeContext(GlobalContext):
             self.sql_backend,
             self.inventory_database,
             int(self.named_parameters.get("attempt", "0")),
+        )
+
+    @cached_property
+    def workflow_run_recorder(self):
+        return WorkflowRunRecorder(
+            self.workspace_client,
+            self.sql_backend,
+            workflow_name=self.named_parameters["job_name"],
+            workflow_id=int(self.named_parameters["job_id"]),
+            workflow_run_id=int(self.named_parameters["parent_run_id"]),
+            workflow_run_attempt=int(self.named_parameters.get("attempt", 0)),
         )
--- a/src/databricks/labs/ucx/contexts/workflow_task.py
+++ b/src/databricks/labs/ucx/contexts/workflow_task.py
@@ -121,5 +121,5 @@ class RuntimeContext(GlobalContext):
             workflow_id=int(self.named_parameters["job_id"]),
             workflow_run_id=int(self.named_parameters["parent_run_id"]),
             workflow_run_attempt=int(self.named_parameters.get("attempt", 0)),
-            workflow_start_time=self.named_parameters["workflow_start_time"],
+            workflow_start_time=self.named_parameters["start_time"],
         )
--- a/src/databricks/labs/ucx/contexts/workflow_task.py
+++ b/src/databricks/labs/ucx/contexts/workflow_task.py
@@ -116,6 +116,7 @@ class RuntimeContext(GlobalContext):
         return WorkflowRunRecorder(
             self.workspace_client,
             self.sql_backend,
+            self.config.ucx_catalog,
             workflow_id=int(self.named_parameters["job_id"]),
             workflow_run_id=int(self.named_parameters["parent_run_id"]),
             workflow_run_attempt=int(self.named_parameters.get("attempt", 0)),
--- a/src/databricks/labs/ucx/contexts/workflow_task.py
+++ b/src/databricks/labs/ucx/contexts/workflow_task.py
@@ -114,9 +114,9 @@ class RuntimeContext(GlobalContext):
     @cached_property
     def workflow_run_recorder(self):
         return WorkflowRunRecorder(
-            self.workspace_client,
             self.sql_backend,
             self.config.ucx_catalog,
+            workspace_id=int(self.named_parameters["workspace_id"]),
             workflow_name=self.named_parameters["workflow"],
             workflow_id=int(self.named_parameters["job_id"]),
             workflow_run_id=int(self.named_parameters["parent_run_id"]),
--- a/src/databricks/labs/ucx/contexts/workflow_task.py
+++ b/src/databricks/labs/ucx/contexts/workflow_task.py
@@ -112,7 +112,7 @@ class RuntimeContext(GlobalContext):
         )
 
     @cached_property
-    def workflow_run_recorder(self):
+    def workflow_run_recorder(self) -> WorkflowRunRecorder:
         return WorkflowRunRecorder(
             self.sql_backend,
             self.config.ucx_catalog,
--- a/src/databricks/labs/ucx/contexts/workflow_task.py
+++ b/src/databricks/labs/ucx/contexts/workflow_task.py
@@ -117,6 +117,7 @@ class RuntimeContext(GlobalContext):
             self.workspace_client,
             self.sql_backend,
             self.config.ucx_catalog,
+            workflow_name=self.named_parameters["workflow"],
             workflow_id=int(self.named_parameters["job_id"]),
             workflow_run_id=int(self.named_parameters["parent_run_id"]),
             workflow_run_attempt=int(self.named_parameters.get("attempt", 0)),
--- a/src/databricks/labs/ucx/contexts/workflow_task.py
+++ b/src/databricks/labs/ucx/contexts/workflow_task.py
@@ -116,7 +116,7 @@ class RuntimeContext(GlobalContext):
         return WorkflowRunRecorder(
             self.sql_backend,
             self.config.ucx_catalog,
-            workspace_id=int(self.named_parameters["workspace_id"]),
+            workspace_id=self.workspace_client.get_workspace_id(),
             workflow_name=self.named_parameters["workflow"],
             workflow_id=int(self.named_parameters["job_id"]),
             workflow_run_id=int(self.named_parameters["parent_run_id"]),
--- a/src/databricks/labs/ucx/contexts/workspace_cli.py
+++ b/src/databricks/labs/ucx/contexts/workspace_cli.py
@@ -18,7 +18,7 @@ from databricks.labs.ucx.azure.locations import ExternalLocationsMigration
 from databricks.labs.ucx.azure.resources import AzureAPIClient, AzureResources
 from databricks.labs.ucx.contexts.application import CliContext
 from databricks.labs.ucx.hive_metastore.table_migration_status import TableMigrationIndex
-from databricks.labs.ucx.progress.install import ProgressTrackingInstaller
+from databricks.labs.ucx.progress.install import ProgressTrackingInstallation
 from databricks.labs.ucx.source_code.base import CurrentSessionState
 from databricks.labs.ucx.source_code.linters.context import LinterContext
 from databricks.labs.ucx.source_code.linters.files import LocalFileMigrator, LocalCodeLinter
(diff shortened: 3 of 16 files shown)


JCZuurmond:
- [ ] TODO: Populate the `workflows_runs` table

src/databricks/labs/ucx/progress/install.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class ArtifactInstallation:
+    """Install resources for UCX's artifacts."""
+
+    _SCHEMA = "artifacts"
+
+    def __init__(self, sql_backend: SqlBackend, ucx_catalog: str) -> None:
+        # `mod` is required parameter, but it's not used in this context.
+        self._schema_deployer = SchemaDeployer(sql_backend, self._SCHEMA, mod=None, catalog=ucx_catalog)
```
JCZuurmond, line 46: Requires lsql release with: https://github.com/databrickslabs/lsql/pull/296

src/databricks/labs/ucx/progress/install.py
```diff
@@ -0,0 +1,51 @@
+import datetime as dt
```
JCZuurmond, line 1: Note that this is part of "progress", but it will cover more than (migration) progress

src/databricks/labs/ucx/contexts/workspace_cli.py
```diff
@@ -179,6 +180,10 @@ def iam_role_creation(self):
     def notebook_loader(self) -> NotebookLoader:
         return NotebookLoader()
 
+    @cached_property
+    def artifact_installation(self) -> ArtifactInstallation:
+        return ArtifactInstallation(self.sql_backend, self.config.ucx_catalog)
```
nfx, line 185: artifacts generally means "wheels", "jars", etc

nfx: perhaps "progress tracking"?

src/databricks/labs/ucx/progress/install.py
JCZuurmond: I am moving it

src/databricks/labs/ucx/progress/install.py
```diff
@@ -0,0 +1,51 @@
...
+    """The identity the workflow was run as`"""
+
+    status: str
+    """The workflow run final status"""
+
+
+class ArtifactInstallation:
+    """Install resources for UCX's artifacts."""
+
+    _SCHEMA = "artifacts"
```
nfx, line 42: i think we wanted to call this database `multiworkspace`, no?

src/databricks/labs/ucx/progress/install.py
nfx: released in v0.12.1

tests/unit/progress/test_install.py
```diff
@@ -0,0 +1,14 @@
+from databricks.labs.ucx.progress.install import ArtifactInstallation
+
+
+def test_history_installation_run_creates_history_schema(mock_backend) -> None:
+    installation = ArtifactInstallation(mock_backend, "ucx")
+    installation.run()
+    assert "CREATE SCHEMA IF NOT EXISTS ucx.history" in mock_backend.queries[0]
```
nfx, line 7:
```suggestion
    assert "CREATE SCHEMA IF NOT EXISTS ucx.multiworkspace" in mock_backend.queries[0]
```

src/databricks/labs/ucx/progress/install.py
```diff
@@ -0,0 +1,51 @@
...
+
+from databricks.labs.lsql.backends import SqlBackend
+from databricks.labs.lsql.deployment import SchemaDeployer
+
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, kw_only=True)
+class WorkflowRun:
```
nfx, line 13: please add a task that runs after all tasks in the MigrationProgress workflow https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/progress/workflows.py#L5

pick the `ctx.named_parameters.get("parent_run_id")` & `job_id` to populate things

https://github.com/databrickslabs/ucx/blob/main/src/databricks/labs/ucx/progress/workflows.py#L5

nfx (changes requested):

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -55,9 +55,12 @@
 
 TEST_RESOURCE_PURGE_TIMEOUT = timedelta(hours=1)
 TEST_NIGHTLY_CI_RESOURCES_PURGE_TIMEOUT = timedelta(hours=3)  # Buffer for debugging nightly integration test runs
+# See https://docs.databricks.com/en/jobs/parameter-value-references.html#supported-value-references
 EXTRA_TASK_PARAMS = {
     "job_id": "{{job_id}}",
+    "job_name": "{{job.name}}",
```
nfx, line 61: the name should be ucx name, e.g. migration-progress, assessment - not the job name "[UCX][prod] assessment"

src/databricks/labs/ucx/progress/workflow_runs.py
```diff
@@ -0,0 +1,79 @@
...
+        """Record a workflow run in the database."""
+        workflow_run = WorkflowRun(
+            started_at=dt.datetime.fromisoformat(self._workflow_start_time),
+            finished_at=dt.datetime.now(),
+            workspace_id=self._ws.get_workspace_id(),
+            workflow_name=self._workflow_name,
+            workflow_id=self._workflow_id,
+            workflow_run_id=self._workflow_run_id,
+            workflow_run_attempt=self._workflow_run_attempt,
+            run_as="UNKOWN",  # TODO Update this
```
nfx, line 71: just fetch it via `self._ws.jobs.get(...)`

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -55,9 +55,12 @@
...
     "run_id": "{{run_id}}",
+    "start_time": "{{job.start_time.iso_datetime}}",
```
nfx, line 63: is it different from the info gotten from API?

JCZuurmond (commented):
Forgot to hit submit on this review, maybe outdated

src/databricks/labs/ucx/installer/workflows.py
JCZuurmond: Combining my answer with this comment and the one above. See [this comment](https://github.com/databrickslabs/ucx/pull/2754#discussion_r1777319587) also.

I do not know if the time is different. However, what I like, is that this task avoids any API calls just like the `LogRecorder`. This would make it easier to add to any workflow. However, if we do not want that anyhow, we can also call the jobs API

src/databricks/labs/ucx/progress/install.py
```diff
@@ -0,0 +1,51 @@
...
+    workflow_id: int
+    """"The workflow id of the workflow that ran."""
+
+    workflow_run_id: int
+    """The workflow run id."""
+
+    run_as: str
+    """The identity the workflow was run as`"""
+
+    status: str
```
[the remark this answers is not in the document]
JCZuurmond: I removed the status

JCZuurmond:
Note that this task will fail when there is no UCX catalog with the `multiworkspace` schema and `workflow_runs` table. The following issue checks at the start of the assessment workflow if the catalog exists: #2577

src/databricks/labs/ucx/progress/workflow_runs.py
JCZuurmond: decided to exclude these instead

src/databricks/labs/ucx/progress/install.py
```diff
@@ -0,0 +1,51 @@
...
+    workflow_name: str
+    """The workflow name that ran."""
+
+    workflow_id: int
+    """"The workflow id of the workflow that ran."""
+
+    workflow_run_id: int
+    """The workflow run id."""
+
+    run_as: str
```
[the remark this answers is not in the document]
JCZuurmond: decided to remove this

JCZuurmond:
> Note that this task will fail when there is no UCX catalog with the `multiworkspace` schema and `workflow_runs` table. The following issue checks at the start of the assessment workflow if the catalog exists: #2577

Added `try-except` with `NotFound` and log error. The log crawler will pick that up

src/databricks/labs/ucx/progress/workflows.py
```diff
@@ -111,3 +111,17 @@ def refresh_table_migration_status(self, ctx: RuntimeContext) -> None:
...
+        depends_on=[
+            crawl_grants,
+            assess_jobs,
+            assess_clusters,
+            assess_pipelines,
+            crawl_cluster_policies,
+            refresh_table_migration_status,
+        ]
+    )
+    def record_workflow_run(self, ctx: RuntimeContext) -> None:
```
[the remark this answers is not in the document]
JCZuurmond: There is something to be said for that. The easiest way to do so, is add the `WorkflowRecord` similarly to the `TaskRunWarningRecorder`: https://github.com/databrickslabs/ucx/blob/d14a2cf98c745224aee01be65a4e12162db4fd85/src/databricks/labs/ucx/installer/workflows.py#L800-L808

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -55,11 +55,14 @@
...
 TEST_RESOURCE_PURGE_TIMEOUT = timedelta(hours=1)
 TEST_NIGHTLY_CI_RESOURCES_PURGE_TIMEOUT = timedelta(hours=3)  # Buffer for debugging nightly integration test runs
+# See https://docs.databricks.com/en/jobs/parameter-value-references.html#supported-value-references
 EXTRA_TASK_PARAMS = {
     "job_id": "{{job_id}}",
     "run_id": "{{run_id}}",
+    "start_time": "{{job.start_time.iso_datetime}}",
     "attempt": "{{job.repair_count}}",
     "parent_run_id": "{{parent_run_id}}",
+    "workspace_id": "{{workspace.id}}",
```
JCZuurmond, line 65: @nfx: Note this new parameter I added to avoid the cli call in `WorkflowRunRecorder` to get the workspace ID

src/databricks/labs/ucx/progress/workflows.py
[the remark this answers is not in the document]
JCZuurmond: Thinking about getting the workflow status during the `WorkflowRunRecorder.record` An approach could be to add the corresponding task after the `parse_logs` task (to every workflow), then we get the result of the `parse_logs` task with a [workflow parameter](https://docs.databricks.com/en/jobs/parameter-value-references.html#supported-value-references)

>  {{tasks.<task_name>.result_state}} The result state of task <task_name>. The possible values are success, failed, excluded, canceled, evicted, timedout, upstream_canceled, upstream_evicted, and upstream_failed.

The `parse_logs` task status is the only way to reliably get the status of the workflow as some tasks will log the error but still continue and finish successfully.

src/databricks/labs/ucx/contexts/workflow_task.py
```diff
@@ -109,3 +110,16 @@ def task_run_warning_recorder(self):
             self.inventory_database,
             int(self.named_parameters.get("attempt", "0")),
         )
+
+    @cached_property
+    def workflow_run_recorder(self):
+        return WorkflowRunRecorder(
+            self.sql_backend,
+            self.config.ucx_catalog,
+            workspace_id=int(self.named_parameters["workspace_id"]),
```
nfx, line 119:
```suggestion
            workspace_id=self.workspace_client.get_workspace_id(),
```
why do you need to pollute the whole named parameters stack just to add a parameter for one task?

src/databricks/labs/ucx/installer/workflows.py
nfx: where the CLI call is coming from? it's RuntimeContext, CLI can't reach it by design.

src/databricks/labs/ucx/cli.py
```diff
@@ -611,6 +611,7 @@ def create_ucx_catalog(w: WorkspaceClient, prompts: Prompts, ctx: WorkspaceConte
     """
     workspace_context = ctx or WorkspaceContext(w)
     workspace_context.catalog_schema.create_ucx_catalog(prompts)
+    workspace_context.progress_tracking_installation.run()
```
nfx, line 614: we should reflect this in the `assign_metastore`

nfx (changes requested):

src/databricks/labs/ucx/installer/workflows.py
JCZuurmond: Meant API call: `self._ws.get_workspace_id()`

src/databricks/labs/ucx/contexts/workflow_task.py
JCZuurmond: The reasoning was to avoid any API calls in the `Recorder`, see [this comment](https://github.com/databrickslabs/ucx/pull/2754#discussion_r1778109572)

src/databricks/labs/ucx/cli.py
JCZuurmond: See this issue on this: https://github.com/databrickslabs/ucx/issues/2737

src/databricks/labs/ucx/contexts/workflow_task.py
JCZuurmond: To avoid any API calls in the `WorkflowRunRecorder` makes it less likely to break

src/databricks/labs/ucx/progress/workflow_runs.py
```diff
@@ -0,0 +1,79 @@
...
+
+    workflow_id: int
+    """"The workflow id of the workflow that ran."""
+
+    workflow_run_id: int
+    """The workflow run id."""
+
+    workflow_run_attempt: int
+    """The workflow run attempt."""
+
```
asnare, line 34: As noted [here](https://github.com/databrickslabs/ucx/issues/2600#issuecomment-2383202696), we also need to have the snapshot identifier to disambiguate between crawls that have occurred within a single workflow run:
```suggestion

    snapshot_id: int
    """The unique identifier of the snapshot that took place, used to identify the records associated with the same snapshot."""
```

src/databricks/labs/ucx/progress/workflow_runs.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+        try:
+            self._sql_backend.save_table(
+                self._full_table_name,
+                [workflow_run],
+                WorkflowRun,
+                mode="append",
+            )
+        except NotFound as e:
+            logger.error(f"Workflow run table not found: {self._full_table_name}", exc_info=e)
```
asnare, line 79: I don't think `NotFound` will be thrown if the table doesn't exist: the semantics of `.save_table()` are that it will be created if it doesn't already exist.

That said, I think we will probably trigger `NotFound` if the catalog or schema don't exist.

src/databricks/labs/ucx/progress/workflows.py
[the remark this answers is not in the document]
asnare: I'm a little sceptical about recording the workflow run like this as an explicit task: some of the tables here are _also_ modified by other workflows.

Once the history table is available, all crawls need to be captured or we end with the inventory file and the history fill out-of-sync.

As such I'd much prefer that this record be created and appended to the table as part of the crawler process where it also appends the records the history table: these two updates go hand-in-hand and are tightly-coupled.

src/databricks/labs/ucx/progress/workflow_runs.py
```diff
@@ -0,0 +1,79 @@
...
+            finished_at=dt.datetime.now(),
+            workspace_id=self._ws.get_workspace_id(),
+            workflow_name=self._workflow_name,
+            workflow_id=self._workflow_id,
+            workflow_run_id=self._workflow_run_id,
+            workflow_run_attempt=self._workflow_run_attempt,
+            run_as="UNKOWN",  # TODO Update this
+            status="RUNNING",  # Always RUNNING as it is called during a running workflow
+        )
+        self._sql_backend.save_table(
```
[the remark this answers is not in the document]
JCZuurmond: https://github.com/databrickslabs/lsql/pull/303

src/databricks/labs/ucx/progress/workflow_runs.py
JCZuurmond: The point is indeed to catch for a catalog or schema not found. However, I am logging here that the table is not found with the error, from which we can deduce what is missing: catalog/schema

src/databricks/labs/ucx/progress/workflow_runs.py
nfx: as per https://github.com/databrickslabs/ucx/pull/2743#pullrequestreview-2341125401, we don't need `snapshot_id`

nfx (approved):
lgtm

src/databricks/labs/ucx/progress/workflows.py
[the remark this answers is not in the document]
JCZuurmond: We decided to add the workflow runs at the end of a workflow as in the PR at this moment

src/databricks/labs/ucx/progress/workflow_runs.py
JCZuurmond: We decided to not add the `snapshot_id` to the workflow runs table

JCZuurmond:
Integration test fails due to a time out on the assessment job. This is happening also for the [assessment integration test](https://github.com/databrickslabs/ucx/issues/2637), but went unnoticed as the [CI is not able to create issues anymore](https://github.com/databrickslabs/ucx/issues/2768). This is blocking the PR at the moment

PR #2766, opened 2024-09-27
Change logic of direct filesystem access linting

ericvergnaud:
## Changes
Our current logic detects all string constants that match a DFSA pattern, excluding false positives on a per use case basis.
This leaves many false positives. Practically, we only care about DFSA if called from `spark` or `dbutils` modules.
This PR implements this change.

### Linked issues
None

### Functionality
None

### Tests
- [x] added unit tests


src/databricks/labs/ucx/source_code/linters/directfs.py
```diff
@@ -72,43 +72,36 @@ class _DetectDirectFsAccessVisitor(TreeVisitor):
...
-        if isinstance(source_node.parent, JoinedStr):
-            return
+    def _check_str_arg(self, call_node: Call, arg_node: NodeNG, inferred: InferredValue):
         value = inferred.as_string()
         for pattern in DIRECT_FS_ACCESS_PATTERNS:
             if not pattern.matches(value):
                 continue
-            # avoid false positives with relative URLs
-            if self._is_http_call_parameter(source_node):
+            # don't capture calls not originating from spark or dbutils
```
JCZuurmond, line 96: I can see this from the code, but do not understand **why** we should not capture calls from spark or dbutils. Could you update the comment to tell the "why"?

src/databricks/labs/ucx/source_code/linters/directfs.py
```diff
@@ -72,43 +72,36 @@ class _DetectDirectFsAccessVisitor(TreeVisitor):
...
+            is_from_spark = False
+            is_from_db_utils = Tree(call_node).is_from_module("dbutils")
+            if not is_from_db_utils:
+                is_from_spark = Tree(call_node).is_from_module("spark")
```
JCZuurmond, line 100: Why can `is_from_spark` only be `True` when `is_from_db_utils` is `False`. I would expect those variables to be independent. If that it is the case, this dependency introduced on line 99 creates a bug on line 104

JCZuurmond:
> Our current logic detects all string constants that match a DFSA pattern, excluding false positives on a per use case basis.

"excluding" should be "including"

JCZuurmond:
> Practically, we only care about DFSA if called from spark or dbutils modules.

Is that case? Why shouldn't all DFSA calls be detected and migrated?

JCZuurmond (changes requested):
See comments

ericvergnaud:
> > Our current logic detects all string constants that match a DFSA pattern, excluding false positives on a per use case basis.
> 
> "excluding" should be "including"?

No. The above is correct. We match all patterns, including very broad ones (anything starting with '/'), then we filter out on a per use case basis (such as api call)

src/databricks/labs/ucx/source_code/linters/directfs.py
ericvergnaud: We **_are_** capturing calls from spark or dbutils. We're only capturing those, because there is no other known way to manipulate data directly from file system

src/databricks/labs/ucx/source_code/linters/directfs.py
ericvergnaud: They are independent, but if either is False we never reach line 104

ericvergnaud:
> > Practically, we only care about DFSA if called from spark or dbutils modules.
> 
> Is that case? Why shouldn't all DFSA calls be detected and migrated?
Can you give an example of direct filesystem access call that happens outside spark or dbutils ?

src/databricks/labs/ucx/source_code/linters/directfs.py
JCZuurmond: Could you explain the later to the comments?

> because there is no other known way to manipulate data directly from file system

src/databricks/labs/ucx/source_code/linters/directfs.py
JCZuurmond: Do you mean:

> They are independent, but if **both are** False we never reach line 104

That is what I read: the "and" in the code with negations on both sides

ericvergnaud: Yup sorry

ericvergnaud: The purpose here is to minimize calls to `is_from_module` because they're not cheap

src/databricks/labs/ucx/source_code/linters/directfs.py
ericvergnaud: done

src/databricks/labs/ucx/source_code/linters/directfs.py
JCZuurmond: > The purpose here is to minimize calls to is_from_module because they're not cheap

Could you add that as a comment? It is fine with me, just want to be sure the logic is correct though

JCZuurmond:
> you give an example of direct filesystem access call that happens outside spark or dbutils ?

Isn't the following a DFSA call outside spark or dbutils `SELECT * FROM parquet.`dbfs:/mnt/my_table/` (in SQL) or `files = os.listdir('/dbfs/mnt/my_table/*.parquet'); df = spark.read.parquet(files)`

src/databricks/labs/ucx/source_code/linters/directfs.py
ericvergnaud: Done

src/databricks/labs/ucx/source_code/linters/directfs.py
```diff
@@ -72,43 +72,37 @@ class _DetectDirectFsAccessVisitor(TreeVisitor):
...
-            # avoid false positives with relative URLs
-            if self._is_http_call_parameter(source_node):
+            # only capture calls originating from spark or dbutils
+            # because there is no other known way to manipulate data directly from file system
+            is_from_spark = False
+            is_from_db_utils = Tree(call_node).is_from_module("dbutils")
+            if not is_from_db_utils:
+                is_from_spark = Tree(call_node).is_from_module("spark")
+            if not is_from_db_utils and not is_from_spark:
                 return
```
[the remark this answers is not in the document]
ericvergnaud: It is but it requires making 2 calls to is_from_module in all cases

src/databricks/labs/ucx/source_code/linters/directfs.py
```diff
@@ -72,43 +72,37 @@ class _DetectDirectFsAccessVisitor(TreeVisitor):
...
-            return
+    def _check_str_arg(self, call_node: Call, arg_node: NodeNG, inferred: InferredValue):
         value = inferred.as_string()
         for pattern in DIRECT_FS_ACCESS_PATTERNS:
             if not pattern.matches(value):
                 continue
-            # avoid false positives with relative URLs
-            if self._is_http_call_parameter(source_node):
+            # only capture calls originating from spark or dbutils
+            # because there is no other known way to manipulate data directly from file system
```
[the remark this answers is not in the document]
ericvergnaud: done

ericvergnaud:
> > you give an example of direct filesystem access call that happens outside spark or dbutils ?
> 
> Isn't the following a DFSA call outside spark or dbutils `SELECT * FROM parquet.dbfs:/mnt/my_table/` (in SQL) or `files = os.listdir('/dbfs/mnt/my_table/*.parquet'); df = spark.read.parquet(files)` (in Python)

SQL is out of scope of this PR which is only about Python. `os.listdir` does not access data, so not sure it needs to be covered ?

nfx (approved):
lgtm

PR #2767, opened 2024-09-27
Improve scan tables in mounts

### LINKED ISSUE #2540: [TECH DEBT]: `TablesInMounts` crawler (`scan-tables-in-mounts-experimental`) issues

### Problem statement

The `TablesInMounts` crawler, invoked as part of the `scan-tables-in-mounts-experimental` workflow` has a few issues with its current implementation:

 - It assumes the table crawler has already run and accesses the inventory directly rather than via the `.snapshot()` interface.
 - It assumes that known tables that can _also_ be located via mounts have a `mounted_` prefix for their name.
 - Rather than augment the table inventory with anything it finds, it simply overwrites the inventory with what it found.

### Proposed Solution

There are a few potential solutions.

### Delete the workflow and crawler.

They were implemented for a single customer, who are done. Currently we don't think anyone is using this.

### "Fix" the crawler

The simplest fix is for the crawler to:

1. Use its own table (`tables_direct` ?) instead of trying to share the `tables` inventory.
2. Use the `.snapshot()` API from the tables crawler instead of just accessing the table directory.
3. Deduplicate against all known tables, not just those that are named with a `mounted_` prefix.

The crawling can't be made part of the existing tables crawler: crawling the mounts for many environments means using credentials that aren't available until after the `assessment` workflow.

An unsolved problem in this solution is how consumers of the `tables` inventory can reliably be updated to (also) use the `tables_direct` inventory: given that it's experimental and running the workflow is the exception rather than the norm, most of the time this inventory doesn't exist. (The normal `.snapshot()` API shouldn't be used because we don't want to trigger a crawl just because the inventory isn't present: most of the time it's not supposed to be.)

### Additional Context

This crawler (and its code) is experimental was written for a single (specific) customer. The customer:

 - Generally used direct filesystem references for most of their tables.
 - Didn't register most of their tables.
 - (I think that) if a table was registered, it used the `mounted_` prefix for the name.

For this landscape, this crawler was fine: replacing the tables inventory entirely with what it found via the mounts was harmless because there weren't any other tables.

new issue: deal with table in mounts crawler cleanup. Options are nuke it, or store it in a different table so that it doesn't clobber the existing inventory. Can't really run it as part of the static table crawler because (aside from performance) it might often need a SP that isn't available until post-assessment.

sdanielzafar:
## Changes
Make the `scan-tables-in-mounts` functionality much more robust and comprehensive, including adding new unit tests and fixing various bugs.

We wanted to use this for a current UC migration for a large retailer and when we ran the job originally it skipped most of the directories and found 8 total tables. With my changes we now parse many many tables. I also worked through a number of bugs.

I did my best to make minimal revisions to the existing code, but I suggest we refactor this to use `os` to list directories in the future instead of `dbutils` so that we can parallelize the operations making it scalable. 

Fixes #2540

### Functionality

- [X] modified existing workflow: scan-tables-in-mounts-experimental

### Tests

- [X] manually tested
- [X] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -456,45 +456,67 @@ def _find_delta_log_folders(self, root_dir: str, delta_log_folders=None) -> dict
...
-
-            if previous_entry:
-                # Happens when first folder was _delta_log and next folders are partitioned folder
-                if previous_entry.format == "DELTA" and self._is_partitioned(file_info.name):
-                    delta_log_folders[root_path] = TableInMount(format=previous_entry.format, is_partitioned=True)
-                # Happens when previous entries where partitioned folders and the current one is delta_log
-                if previous_entry.is_partitioned and table_in_mount and table_in_mount.format == "DELTA":
-                    delta_log_folders[root_path] = TableInMount(format=table_in_mount.format, is_partitioned=True)
-                continue
+        if file_infos:
```
JCZuurmond, line 459: Is this necessary? Is it `None` instead of an empty list, if so, make it an empty list or return early:

```
if not file_infos:
    return delta_log_folder
```

or 

```
file_infos = self._dbutils.fs.ls(root_dir) or []
```

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -456,45 +456,67 @@ def _find_delta_log_folders(self, root_dir: str, delta_log_folders=None) -> dict
...
+                    logger.debug(f"Added {table_in_mount.format} table for {root_path} (normal delta)")
 
-            if not table_in_mount:
-                self._find_delta_log_folders(file_info.path, delta_log_folders)
-                continue
+                else:
+                    delta_log_folders[root_path] = table_in_mount
+                    logger.debug(f"Added {table_in_mount.format} table for {root_path} (general)")
+                    if self._is_recursible_dir(file_info):
+                        self._find_delta_log_folders(file_info.path, delta_log_folders)
```
JCZuurmond, line 508: Why search for more delta log folders if this one is already added?

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -530,3 +552,23 @@ def _is_json(self, file_name: str) -> bool:
...
+        path_dirs = [file.name for file in self._dbutils.fs.ls(file_info.path) if file.size == 0]
+        return 'commits/' in path_dirs and 'offsets/' in path_dirs
+
+    def _is_recursible_dir(self, file_info: FileInfo) -> bool:
+        # Rules for recursing into a folder
+        # - should be size 0 (usually a directory)
+        # - should not be the _delta_log directory
+        # - file should not be partitioned
+        # - file name should not start with part-
+        # - brackets are not allowed in dbutils.fa.ls
```
JCZuurmond, line 566: Does this make a directory non-recursible or should it be handled when calling the `ls`?

```suggestion
        # - brackets are not allowed in dbutils.fs.ls
```

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -530,3 +552,23 @@ def _is_json(self, file_name: str) -> bool:
...
+        return (
+            file_info.size == 0
+            and file_info.name != "_delta_log/"
+            and not self._is_partitioned(file_info.name)
+            and not file_info.name.startswith("part-")
+            and not any([char in file_info.name for char in "[]"])
+            and not self._is_streaming_checkpoint_dir(file_info)
```
JCZuurmond, line 573: This is not mentioned in the comment above

src/databricks/labs/ucx/source_code/known.json
```diff
@@ -237,6 +237,14 @@
     "audioread.rawread": [],
```
JCZuurmond, line 237: Remove changes in `known.json`

tests/unit/hive_metastore/test_locations.py
```diff
@@ -1,6 +1,7 @@
 from unittest.mock import Mock, call, create_autospec
 
 import pytest
+import sys
```
JCZuurmond, line 4: Move to first import block, is a built-in package

tests/unit/hive_metastore/test_locations.py
```diff
@@ -306,9 +307,9 @@ def test_match_table_external_locations():
 def test_mount_listing_multiple_folders():
     client = create_autospec(WorkspaceClient)
 
-    first_folder = FileInfo("dbfs:/mnt/test_mount/table1/", "table1/", "", "")
-    second_folder = FileInfo("dbfs:/mnt/test_mount/table2/", "table2/", "", "")
-    folder_table1 = FileInfo("dbfs:/mnt/test_mount/table1/_delta_log/", "_delta_log/", "", "")
+    first_folder = FileInfo("dbfs:/mnt/test_mount/table1/", "table1/", 0, "")
```
JCZuurmond, line 310: Should have been picked up by the linter, could you trigger the linter for this tests and the tests below by adding a type hint:

```
def test_mount_listing_multiple_folders() -> None
```

tests/unit/hive_metastore/test_locations.py
```diff
@@ -718,3 +719,184 @@ def my_side_effect(path, **_):
     assert results[0].location == "adls://bucket/table1"
     assert results[1].location == "dbfs:/mnt/test_mount/table2"
     assert results[2].location is None
+
+
+def test_mount_listing_misplaced_flat_file():
```
JCZuurmond, line 724: What should happen with the misplaced flat file? Should it be ignored by the tables in mounts?

tests/unit/hive_metastore/test_locations.py
```diff
@@ -718,3 +719,184 @@ def my_side_effect(path, **_):
...
+    client.dbutils.fs.ls.side_effect = my_side_effect
+    backend = MockBackend(
+        rows={
+            '`hive_metastore`.`test`.`tables`': [],
+            '`test`.`mounts`': MOUNT_STORAGE[("/mnt/test_mount", "adls://bucket/")],
+        }
+    )
+
+    original_limit = sys.getrecursionlimit()
+    sys.setrecursionlimit(150)  # Temporarily lower recursion limit to catch errors early
```
JCZuurmond, line 852: Make this a fixture

tests/unit/hive_metastore/test_locations.py
```diff
@@ -718,3 +719,184 @@ def my_side_effect(path, **_):
...
+    try:
+        mounts = Mounts(backend, client, "test")
+        results = TablesInMounts(backend, client, "test", mounts).snapshot()
+    except RecursionError:
+        pytest.fail("Recursion depth exceeded, possible infinite loop.")
```
JCZuurmond, line 857: Should this be an assert?

tests/unit/hive_metastore/test_locations.py
```diff
@@ -718,3 +719,184 @@ def my_side_effect(path, **_):
...
+        }
+    )
+    mounts = Mounts(backend, client, "test")
+    results = TablesInMounts(backend, client, "test", mounts).snapshot()
+    assert len(results) == 1
+    assert results[0].table_format == "DELTA"
+    assert results[0].is_partitioned
+
+
+def test_mount_infinite_loop():
```
JCZuurmond, line 824: What should happen with an inifinte loop?

tests/unit/hive_metastore/test_locations.py
```diff
@@ -718,3 +719,184 @@ def my_side_effect(path, **_):
...
+        results = TablesInMounts(backend, client, "test", mounts).snapshot()
+    except RecursionError:
+        pytest.fail("Recursion depth exceeded, possible infinite loop.")
+    finally:
+        sys.setrecursionlimit(original_limit)  # Restore the original limit after test
+
+    assert len(results) == 0
+
+
+def test_mount_exclude_checkpoint_dir():
```
JCZuurmond, line 864:
```suggestion
def test_mount_exclude_streaming_checkpoint():
```

OR

```suggestion
def test_mount_exclude_streaming_checkpoint_directory():
```

JCZuurmond (changes requested):
Thanks for submitting the PR. Generally looks good, have some comments.

@ericvergnaud, @nfx or @asnare  : Probably know a good way to use a UCX API for listing files on Databricks as you mention in the PR comment

src/databricks/labs/ucx/hive_metastore/locations.py
sdanielzafar: fixed

src/databricks/labs/ucx/hive_metastore/locations.py
sdanielzafar: this is just part of the recursive algorithm. For instance, let's say at the user's top level directory there is a spare `test.csv` thrown in somewhere. You don't want to exit saying that the top level dir is a CSV file.

src/databricks/labs/ucx/hive_metastore/locations.py
sdanielzafar: yes- dbutils cannot handle brackets. Something I encountered in the wild during testing on a customer's mount directory.

src/databricks/labs/ucx/source_code/known.json
sdanielzafar: done

tests/unit/hive_metastore/test_locations.py
sdanielzafar: No, there is no simple heuristic to ignore it, so we file it as a table in case it's important.

tests/unit/hive_metastore/test_locations.py
sdanielzafar: It should be caught before going to PROD! This was a bug fix. I created this test to help identify an fix an infinite loop that sometimes appeared for files that had an `=` in them.

tests/unit/hive_metastore/test_locations.py
sdanielzafar: no need?

sdanielzafar:
Thanks for the insightful comments @JCZuurmond! I have pushed some changes in between my PTO adventures. Sorry for the latency.

src/databricks/labs/ucx/source_code/known.json
```diff
@@ -237,6 +237,14 @@
...
     "audioread.version": []
   },
+  "autocommand": {
```
nfx, line 240: revert changes to this file

tests/unit/hive_metastore/test_locations.py
nfx: this may break other tests, as all of them are executed in multiple threads. not quite happy about this. is there another option?

nfx (changes requested):

tests/unit/hive_metastore/test_locations.py
sdanielzafar: I'm not understanding, test does have a type hint (return `None`)?

src/databricks/labs/ucx/source_code/known.json
sdanielzafar: done- shows this file is not being modified now, sorry abt that!

sdanielzafar:
@nfx - addressed your review comments, thanks! Let me know if you need anything else. I _think_ the recursion limit issue will be fixed with a pytest fixture, so I added that. If not, I'm happy to get rid of that test since it's a pointed bug fix test.

tests/unit/hive_metastore/test_locations.py
JCZuurmond: I do not see a type hint for `test_mount_listing_multiple_folders`? 

If a method does not have a type hint in its signature then the body will not be linted

sdanielzafar: @JCZuurmond - Not sure I follow. First, there are no methods here, only functions. Second, this does have a return type hint of `None` can you please be clear about what you are looking for?

https://github.com/databrickslabs/ucx/blob/9c8b22be4dd29ecba8ebb9c5ebcf9061b076cd79/tests/unit/hive_metastore/test_locations.py#L307

JCZuurmond: Did you push your changes?

Screenshot 2024-10-07 at 09 24 24

JCZuurmond: I probably made the mistake of looking at the comment in context, thus missing that commit. Sorry for the confusion

sdanielzafar:
@nfx - did the rebase and resolved the test issues you mentioned. I'm still getting these errors with `make fmt` which don't make sense to me as I did not modify these files. Thoughts?:
```
(.venv) dan.zafar@YFV9607WJY ucx % make fmt              
hatch run fmt
cmd [1] | black . --extend-exclude 'tests/unit/source_code/samples/'
All done! ✨ 🍰 ✨
336 files left unchanged.
cmd [2] | ruff check . --fix
All checks passed!
cmd [3] | mypy --disable-error-code 'annotation-unchecked' --exclude 'tests/unit/source_code/samples/*' --exclude dist .
venv/bin/activate_this.py:27: error: "str" has no attribute "decode"; maybe "encode"?  [attr-defined]
venv/bin/activate_this.py:30: error: Module has no attribute "real_prefix"  [attr-defined]
Found 2 errors in 1 file (checked 337 source files)
make: *** [fmt] Error 1
```

tests/unit/hive_metastore/test_locations.py
sdanielzafar: used a pytest fixure.

nfx (changes requested):
diff is too big and impossible to review. please rebase on the latest `main` from origin and force-push
image

sdanielzafar:
@nfx - was able to get the rebase sorted out after all, you can see only two files are being changed now. Can we merge this?

tests/unit/hive_metastore/test_locations.py
```diff
@@ -734,3 +735,188 @@ def test_resolve_dbfs_root_in_hms_federation():
...
+    results = TablesInMounts(backend, client, "test", mounts).snapshot(force_refresh=True)
+    assert len(results) == 1
+    assert results[0].table_format == "DELTA"
+    assert results[0].is_partitioned
+
+
+@pytest.fixture
+def set_recursion_limit():
+    original_limit = sys.getrecursionlimit()
+    sys.setrecursionlimit(150)
```
nfx, line 843: please rewrite algorithm using the iterative approach, e.g. managing stack via list. if this hack is required on the testing side, it'll blow up in the customer side.

sdanielzafar: This was just a test to catch an infinite loop in the code, which was a bug I had to solve when I picked up this code initially. I can get rid of the test if it's causing too much confusion.

JCZuurmond (commented):
LGTM besides the recursion limit, please remove and ask for a new review

nfx (approved):
let's merge this

Issue #2761, opened 2024-09-26
[FEATURE]: Identify owners for inventory types that we track the history of


asnare:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

We will shortly be tracking history for inventory types that are routinely refreshed during migration. The history journal that we maintain will require that for each record an owner is available: the owner is the person (or group) responsible for the underlying resource being migrated. If this is not available, the workspace administrator should be used instead.

### Proposed Solution

Each crawler that is responsible for a refreshable class will need to be updated to have code that can identify the owner for its `Result` type.

### Documentation 
- [ ] #3067

### Additional Context

Related issues:

 - #2572
 - #2744 

Blocks:

 - #2573
 - #2743

asnare:
We need to track owners for the following inventory types:

 - `ClusterInfo`:

    1. Existing `creator` field, which is optional.
    2. Workspace admin.

 - `DirectFsAccess`:

    1. For:

       - _Queries_: the `user` attribute of the query[^1], which is optional.
       - _Jobs_: the `owner` of the DBFS or Workspace path[^1][^2] for the notebook or source, which is optional.

    2. Workspace admin.

 - `Grant`, `Table` and `UDF`:

    1. Workspace admin.

    _Note: No attempt to determine creator via table/UDF properties._

 - `JobInfo`:

    1. Existing `creator` field, which is optional.
    3. Workspace admin.

- `PipelineInfo`:

    1. Existing `creator_name` field, which is optional.
    2. Workspace admin.

- `PolicyInfo`:

    1. Existing `creator` field, which is optional.
    2. Workspace admin.

- `TableMigrationStatus`:

    1. The owner of the source table, as defined above.



Where the workspace admin is needed (because a more appropriate owner cannot be determined) the algorithm is as follows:

1. Query all active workspace admins, sort alphabetically by user-name, use the first one.
4. If there are no workspace admins, query all active account admins associated with the workspace, sort alphabetically by user-name, use the first one.
5. Raise an error if we still don't have an identity. (Note: this is possible, especially due to accounts being decommissioned and leaving a workspace without an active admin.)

[^1]: This is not yet available on the `DirectFsAccess` instances; a schema change will probably be required along with code to expose this information.
[^2]: The APIs for DBFS and Workspace paths don't expose the owner/creator information, so this information is unavailable. If it were available, this would first be exposed via the `.owner` attribute of our `pathlib` emulation.

PR #2774, opened 2024-10-01
Crawler: support for object ownership

asnare:
## Changes

As part of #2761 we need to have a way for determining the user responsible for some of our inventory types. This PR updates the crawler framework so that:

 - There is a way to identify an owner of a resource referred to by inventory records.
 - When the owner cannot be identified, a workspace or account administrator is used instead.

### Linked issues

Progresses #2761.

### Functionality

- A component for locating an administrator user.
- Ownership information for the following inventory types:

  - [X] `ClusterInfo`
  - [x] `DirectFsAccess` (stubbed)
  - [X] `Grant`
  - [x] `JobInfo`
  - [x] `PipelineInfo`
  - [X] `PolicyInfo`
  - [x] `Table`
  - [x] `TableMigrationStatus`
  - [x] `UDF`

### Tests

- [x] added unit tests
- [x] added integration tests


src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -107,6 +117,56 @@ def snapshot(self, *, force_refresh: bool = False) -> Iterable[Result]:
...
+        return self._result_owner(result) or self._workspace_admin
+
+    @cached_property
+    def _workspace_admin(self) -> str:
+        # Avoid repeatedly hitting the shared cache.
+        return self._find_administrator_for(self._ws)
+
+    @classmethod
+    @final
+    def _find_administrator_for(cls, ws: WorkspaceClient) -> str:
```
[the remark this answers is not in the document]
asnare: This was for lazily determining the workspace administrator; it's been removed from here, and is now part of the `Ownership` component.

src/databricks/labs/ucx/framework/utils.py
```diff
@@ -22,6 +28,55 @@ def escape_sql_identifier(path: str, *, maxsplit: int = 2) -> str:
     return ".".join(escaped)
 
 
+def _has_role(user: User, role: str) -> bool:
```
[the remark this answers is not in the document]
asnare: I've created an `Ownership` component: the base class will hold the common logic for locating an administrator, and type-specific subclasses will do the per-type handling that we need for the types that will be captured in the history table.

src/databricks/labs/ucx/framework/utils.py
```diff
@@ -22,6 +28,55 @@ def escape_sql_identifier(path: str, *, maxsplit: int = 2) -> str:
...
+def find_workspace_admins(ws: WorkspaceClient) -> Iterable[User]:
+    """Enumerate the active workspace administrators in a given workspace.
+
+    Arguments:
+        ws (WorkspaceClient): The client for the workspace whose administrators should be enumerated.
+    Returns:
+        Iterable[User]: The active workspace administrators, if any.
+    """
+    all_users = ws.users.list(attributes="id,active,userName,roles")
+    return (user for user in all_users if user.active and _has_role(user, "workspace_admin"))
```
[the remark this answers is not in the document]
asnare: No, this was just a guess while I figured out how it's supposed to be done. (And indeed, it now uses membership of the `admins` workspace group.)

src/databricks/labs/ucx/framework/utils.py
```diff
@@ -22,6 +28,55 @@ def escape_sql_identifier(path: str, *, maxsplit: int = 2) -> str:
...
+    account administrator is sought, again returning the first alphabetically by user-name if there is more than one.
+
+    Arguments:
+        ws (WorkspaceClient): The client for the workspace for which an administrator should be located.
+    Returns:
+        the first (alphabetically by user-name) active workspace or account administrator, or `None` if neither can be
+        found.
+    """
+    first_user = functools.partial(min, default=None, key=lambda user: user.name)
+    return first_user(find_workspace_admins(ws)) or first_user(find_account_admins(ws))
```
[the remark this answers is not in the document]
asnare: This implementation is already sorting[^1]: `head(sort(…))` → `min(…)`


[^1]: It was using the wrong attribute; it should be `user.user_name` and that's now fixed.

src/databricks/labs/ucx/assessment/clusters.py
```diff
@@ -154,17 +156,20 @@ def _assess_clusters(self, all_clusters):
         for cluster in all_clusters:
             if cluster.cluster_source == ClusterSource.JOB:
                 continue
-            if not cluster.creator_user_name:
+            creator = cluster.creator_user_name
+            if not creator:
                 logger.warning(
                     f"Cluster {cluster.cluster_id} have Unknown creator, it means that the original creator "
                     f"has been deleted and should be re-created"
```
JCZuurmond, line 163: should the User be recreated or the cluster?

src/databricks/labs/ucx/assessment/clusters.py
```diff
@@ -154,17 +156,20 @@ def _assess_clusters(self, all_clusters):
...
                 )
+                # Normalize empty creator.
```
JCZuurmond, line 165: Normalize?

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -0,0 +1,204 @@
...
+
+class DataclassInstance(Protocol):
+    __dataclass_fields__: ClassVar[dict]
+
+
+Record = TypeVar("Record")
+
+
+class AdministratorFinder(ABC):
+    def __init__(self, ws: WorkspaceClient):
```
JCZuurmond, line 23:
```suggestion
    def __init__(self, ws: WorkspaceClient) -> None:
```

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -0,0 +1,204 @@
...
+        admin_users = [user for user in all_users if user.user_name and self._is_active_admin(user)]
+        logger.debug(f"Verifying membership of the 'admins' workspace group for users: {admin_users}")
+        candidate_group_ids = set()
+        for user in admin_users:
+            if not user.groups:
+                continue
+            for group in user.groups:
+                if group.display == "admins" and group.value:
+                    candidate_group_ids.add(group.value)
+        admin_groups = list(self._filter_workspace_groups(candidate_group_ids))
```
JCZuurmond, line 82: Still ids:
```suggestion
        admin_groups_ids = list(self._filter_workspace_groups(candidate_group_ids))
```

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -0,0 +1,204 @@
...
+                continue
+            seen.add(group_id)
+            try:
+                group = self._ws.groups.get(group_id)
+            except NotFound:
+                continue
+            if group.meta and group.meta.resource_type == "WorkspaceGroup":
+                yield group_id
+
+    def find_admin_users(self) -> Iterable[User]:
```
JCZuurmond, line 63:
```suggestion
    def find_admin_users(self) -> list[User]:
```

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -0,0 +1,204 @@
...
+        return found_admin
+
+
+class Ownership(ABC, Generic[Record]):
+    """Determine an owner for a given type of object."""
+
+    def __init__(self, administrator_locator: AdministratorLocator) -> None:
+        self._administrator_locator = administrator_locator
+
+    @final
```
JCZuurmond, line 178: What does this do?

src/databricks/labs/ucx/hive_metastore/table_size.py
```diff
@@ -23,20 +22,25 @@ class TableSize:
 
 
 class TableSizeCrawler(CrawlerBase[TableSize]):
-    def __init__(self, backend: SqlBackend, schema, include_databases: list[str] | None = None):
+    # TODO: Ensure TablesCrawler and FasterTableScanCrawler share a common interface.
```
JCZuurmond, line 25: @pritishpai : Maybe could be part of: https://github.com/databrickslabs/ucx/pull/2819

JCZuurmond (commented):
My first comments

src/databricks/labs/ucx/framework/owners.py
asnare: It's a decorator that tells type checkers that this method/attribute may not be overridden in a subclass.

Reference: https://docs.python.org/3.10/library/typing.html?highlight=final#typing.final

src/databricks/labs/ucx/assessment/clusters.py
```diff
@@ -179,6 +184,16 @@ def _try_fetch(self) -> Iterable[ClusterInfo]:
...
 
 
+class ClusterOwnership(Ownership[ClusterInfo]):
+    """Determine ownership of clusters in the inventory.
+
+    This is the cluster creator (if known), or otherwise an administrator.
+    """
+
+    def _get_owner(self, record: ClusterInfo) -> str | None:
+        return record.creator
```
nfx, line 194: where's the `or otherwise an administrator` logic? 
(edit) found it later in code.

Remove the `or otherwise an administrator` string and rename the method to `maybe_direct_owner`. here `direct` explicitly tells that it is on the record

src/databricks/labs/ucx/assessment/clusters.py
```diff
@@ -210,7 +226,7 @@ def _assess_policies(self, all_policices) -> Iterable[PolicyInfo]:
             except KeyError:
                 spark_version = None
             policy_name = policy.name
-            creator_name = policy.creator_user_name
+            creator_name = policy.creator_user_name or None
```
nfx, line 229: i wonder if `or None` makes sense here

src/databricks/labs/ucx/assessment/jobs.py
```diff
@@ -106,11 +108,14 @@ def _prepare(all_jobs) -> tuple[dict[int, set[str]], dict[int, JobInfo]]:
...
             job_assessment[job.job_id] = set()
-            if not job.creator_user_name:
+            creator_user_name = job.creator_user_name
+            if not creator_user_name:
                 logger.warning(
                     f"Job {job.job_id} have Unknown creator, it means that the original creator has been deleted "
                     f"and should be re-created"
                 )
+                # Normalization.
+                creator_user_name = None
```
nfx, line 118: is it when `job.creator_user_name == ""`?

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -0,0 +1,204 @@
...
+        """Determine if a user is an active administrator."""
+        return bool(user.active) and self._member_of_group_named(user, "admins")
+
+    def _filter_workspace_groups(self, identifiers: Iterable[str]) -> Iterable[str]:
+        """Limit a set of identifiers to those that are workspace groups."""
+        seen = set()
+        for group_id in identifiers:
+            if group_id in seen:
+                continue
+            seen.add(group_id)
```
nfx, line 55:
```suggestion
        for group_id in set(identifiers):
```
you want to process each group only once, even if it's supplied twice, right?

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -0,0 +1,204 @@
...
+                if group.display == "admins" and group.value:
+                    candidate_group_ids.add(group.value)
+        admin_group_ids = list(self._filter_workspace_groups(candidate_group_ids))
+        match admin_group_ids:
+            case []:
+                return ()
+            case [admin_group]:
+                return (user for user in admin_users if self._member_of_group(user, admin_group))
+            case _:
+                msg = f"Multiple 'admins' workspace groups found; something is wrong: {admin_group_ids}"
```
nfx, line 89: how this is possible? pick only `resource_type == "WorkspaceGroup"`

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -0,0 +1,204 @@
...
+        return self._ws.get_workspace_id()
+
+    @cached_property
+    def _found_admin(self) -> str | None:
+        # Lazily instantiate and query the finders in an attempt to locate an admin user.
+        finders = (finder(self._ws) for finder in self._finders)
+        # If a finder returns multiple admin users, use the first (alphabetically by user-name).
+        first_user = functools.partial(min, default=None, key=lambda user: user.user_name)
+        found_admin_users: Iterable[User | None] = (first_user(finder.find_admin_users()) for finder in finders)
+        return next((user.user_name for user in found_admin_users if user), None)
```
nfx, line 156: it's a bit hard to follow - doesn't pass my jetlag test, honestly. could you rewrite it as a for-loop?

```
for klass in self._finders:
  finder = klass(self._ws)
  users = sorted(finder.find_admin_users(), key=lambda _: _.user_name)
  for user in users:
    return user.user_name
return None
```

same amount of lines, but more readable.

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -0,0 +1,204 @@
...
+
+    @property
+    def workspace_administrator(self) -> str:
```
nfx, line 159:
```suggestion
    def get_workspace_administrator(self) -> str:
```
let's not make public `@property` with side-effects, like calling APIs.

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -0,0 +1,204 @@
...
+class Ownership(ABC, Generic[Record]):
+    """Determine an owner for a given type of object."""
+
+    def __init__(self, administrator_locator: AdministratorLocator) -> None:
+        self._administrator_locator = administrator_locator
+
+    @final
+    @property
+    def administrator_locator(self):
+        return self._administrator_locator
```
nfx, line 181:
```suggestion
```
`GlobalContext` has to be the only dependency injection source of truth. this "property" breaks this rule and implies false assumptions.

src/databricks/labs/ucx/assessment/jobs.py
```diff
@@ -140,6 +145,16 @@ def _check_jar_task(self, all_task: list[RunTask]) -> list[str]:
...
 
 
+class JobOwnership(Ownership[JobInfo]):
+    """Determine ownership of jobs (workflows) in the inventory.
+
+    This is the pipeline creator (if known), or otherwise an administrator.
+    """
+
+    def _get_owner(self, record: JobInfo) -> str | None:
+        return record.creator
```
nfx, line 155: what about `run_as`?

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -0,0 +1,204 @@
...
+
+    @final
+    def owner_of(self, record: Record) -> str:
```
nfx, line 184: i am struggling to find where this method is called in production code. shouldn't it be called from history recorder?

nfx (changes requested):

src/databricks/labs/ucx/assessment/jobs.py
asnare: Indeed. The existing code treated `""` and `None` as equivalent. This was intended to ensure we only end up with `None`.

I like the `or None` suggestion above better.

src/databricks/labs/ucx/framework/owners.py
asnare: Absolutely: this method is not being accessed yet. It will be called from the history recorder, but this PR blocks the history recorder (#2743). The history recorder will be updated to use this interface.

src/databricks/labs/ucx/assessment/clusters.py
asnare: The prior code had a code-path where `""` and `None` were both treated as missing, and really they're the same thing so normalising to `None` makes sense. It's been slightly simplified now.

src/databricks/labs/ucx/assessment/clusters.py
asnare: My reasoning here:

 * We have the same code-path on other types[^1] that treats both `""` and `None` as missing.
 * For all practical purposes, if this is ever `""` then indeed the creator is missing.
 * Rather than have downstream usage deal with these (potentially) differently, normalise them during crawling so that `None` means missing and anything else is not missing.

An alternative would be to treat `""` as present and for consistency update the other crawlers to not warn if this were to ever occur.

Either way is fine, so long as we're consistent and the semantics are well-defined.

[^1]: `ClusterInfo`, `JobInfo`, `PipelineInfo`

src/databricks/labs/ucx/framework/owners.py
asnare: Indeed, this was clumsy. I'm simplified it.

src/databricks/labs/ucx/assessment/clusters.py
asnare: The cluster should either have its owner changed (there is an API for this) or be recreated.

Interestingly there's a mix of terminology here: the property here is `creator_user_name` but the REST API for changing it is `…/clusters/change-owner`, and such requests have a `owner_username` property.

src/databricks/labs/ucx/framework/owners.py
asnare: Agreed, I've refactored for readability.

src/databricks/labs/ucx/assessment/jobs.py
asnare: This is interesting. This is not being gathered by the crawler, which means either: a) updating the crawler to include it; b) doing it here.

Wherever this is done, it looks more complicated than at first glance: the `run_as` property is documented on the SDK as write-only. I'll do some tests to see if this is really the case or not, but if that's correct then it needs to be found somewhere else and I'm not (yet) sure where?

Either way I'd suggest covering this as a separate PR.

src/databricks/labs/ucx/framework/owners.py
asnare: These are lazy; they don't return lists. (Although sometimes they do return an empty tuple.)

tests/integration/assessment/test_clusters.py
```diff
@@ -39,6 +45,41 @@ def test_cluster_crawler_no_isolation(ws, make_cluster, inventory_schema, sql_ba
...
+    # Find the crawled records for our clusters.
+    cluster_record_with_owner = next(record for record in records if record.cluster_id == cluster_with_owner.cluster_id)
+    cluster_record_without_owner = next(
+        record for record in records if record.cluster_id == cluster_without_owner.cluster_id
+    )
+
+    # Verify ownership is as expected.
+    ownership = ClusterOwnership(runtime_ctx.administrator_locator)
+    assert ownership.owner_of(cluster_record_with_owner) == ws.current_user.me().user_name
+    assert "@" in ownership.owner_of(cluster_record_without_owner)
```
nfx, line 80: exact user name match is preferred

tests/integration/assessment/test_pipelines.py
```diff
@@ -42,3 +42,23 @@ def test_pipeline_with_secret_conf_crawler(ws, make_pipeline, inventory_schema,
 
     assert len(results) >= 1
     assert results[0].pipeline_id == created_pipeline.pipeline_id
+
+
+def test_pipeline_ownership(ws, runtime_ctx, make_pipeline, inventory_schema, sql_backend) -> None:
+    """Verify the ownership can be determined for crawled pipelines."""
```
nfx, line 48:
```suggestion
def test_verify_ownership_determined_for_crawled_pipelines(ws, runtime_ctx, make_pipeline, inventory_schema, sql_backend) -> None:
```

tests/integration/hive_metastore/test_grants.py
```diff
@@ -108,3 +113,25 @@ def test_all_grants_for_other_objects(
...
+    # Produce the crawled records.
+    crawler = GrantsCrawler(table_crawler, udf_crawler, include_databases=[schema.name])
+    records = crawler.snapshot(force_refresh=True)
+
+    # Find the crawled record for the grant we made.
+    grant_record = next(record for record in records if record.this_type_and_key() == ("DATABASE", schema.full_name))
+
+    # Verify ownership can be made.
+    ownership = GrantOwnership(runtime_ctx.administrator_locator)
+    assert "@" in ownership.owner_of(grant_record)
```
nfx, line 137: rewrite `assert "@" in ownership.owner_of(grant_record)` to check for concrete usernames as much as possible, as these tests don't test enough - e.g. hiding the problem

nfx (commented):
rewrite `assert "@" in ownership.owner_of(grant_record)` to check for concrete usernames as much as possible, as these tests don't test enough - e.g. hiding the problem

nfx (approved):
lgtm

PR #2813, opened 2024-10-03
[TECH DEBT] Add type hints for cached properties

JCZuurmond:
## Changes
Add type hints for cached properties and resolve the linting issues found after adding the type hints. Would have caught the integration test issues introduced in #2772

PR #2819, opened 2024-10-03
Adds type hints with `TablesCrawler` for the table crawlers where required #2819

pritishpai:
\## Changes
Adds type hints with `TablesCrawler` for the table crawlers where required #2819

FasterTableCrawler now extends TablesCrawler

### Linked issues
(https://github.com/databrickslabs/ucx/pull/2813/files#r1786418356)

### DIFF
--- a/src/databricks/labs/ucx/contexts/workflow_task.py
+++ b/src/databricks/labs/ucx/contexts/workflow_task.py
@@ -12,10 +12,9 @@ from databricks.labs.ucx.assessment.jobs import JobsCrawler, SubmitRunsCrawler
 from databricks.labs.ucx.assessment.pipelines import PipelinesCrawler
 from databricks.labs.ucx.config import WorkspaceConfig
 from databricks.labs.ucx.contexts.application import GlobalContext
-from databricks.labs.ucx.framework.crawlers import CrawlerBase
 from databricks.labs.ucx.hive_metastore import TablesInMounts
 from databricks.labs.ucx.hive_metastore.table_size import TableSizeCrawler
-from databricks.labs.ucx.hive_metastore.tables import FasterTableScanCrawler, Table
+from databricks.labs.ucx.hive_metastore.tables import FasterTableScanCrawler
 from databricks.labs.ucx.installer.logs import TaskRunWarningRecorder
 from databricks.labs.ucx.progress.workflow_runs import WorkflowRunRecorder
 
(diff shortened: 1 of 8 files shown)


src/databricks/labs/ucx/contexts/workflow_task.py
```diff
@@ -84,7 +84,7 @@ def global_init_scripts_crawler(self):
         return GlobalInitScriptCrawler(self.workspace_client, self.sql_backend, self.inventory_database)
 
     @cached_property
-    def tables_crawler(self):
+    def tables_crawler(self) -> BaseTablesCrawler:
```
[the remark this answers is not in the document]
JCZuurmond: This is even better!

JCZuurmond:
> why do we need this?

I request Pritish to solve the missing type hint for the table crawlers, what you suggested in a comment is a good approach

pritishpai:
Used superclass CrawlerBase for type hints. Had to add a placeholder method 'load_one' for one of the usages of TableCrawler

src/databricks/labs/ucx/framework/crawlers.py
```diff
@@ -126,6 +126,14 @@ def _crawl(self) -> Iterable[Result]:
...
         """
 
+    def load_one(self, schema_name: str, table_name: str) -> Result | None:
+        """
+        Placeholder method to using super class for type hints of TableCrawler
+        :param schema_name:
+        :param table_name:
+        :return:
+        """
+
```
asnare, line 136: This isn't really a satisfactory way of solving the problem: it basically says "all crawlers support this method" but for most it doesn't do anything.

asnare (commented):
The problem with type hints for this is that we have two different types and they have different interfaces (due to `load_one()`).

Without this difference, either of these would work as the hint for the component:

 - `CrawlerBase[Table]`
 - `TablesCrawler | FasterTableScanCrawler`

I think the proper fix is to recognise that the difference between these classes is limited to the crawling implementation: the "fetcher" is the same for both. With this in mind, my version of this PR would:

 - Refactor `TablesCrawler` to be abstract, and contain:

    - An implementation for `_try_fetch()` and `load_one()`.
    - Leave `_crawl()` unimplemented (and therefore abstract).

 - Update `FasterTableScanCrawler`:

    - Inherit from `TablesCrawler`.
    - Implement `_crawl()`, which should just be what we have now.
    - Drop the `_try_fetch()` implementation: this just comes from the parent.

 - Introduce `SlowTableScanCrawler`:

    - Inherit from `TablesCrawler`.
    - Copy the current SQL-based implementation for `.crawl()` from what we currently have in `TablesCrawler`.
    - No `_try_fetch()` implementation needed: this just comes from the parent.
    - Existing unit tests that target `TablesCrawler` need to target this class instead.

This will reduce code duplication and allow the type hints on the contexts (and other places where we need this hint) to be `TablesCrawler`.

src/databricks/labs/ucx/framework/crawlers.py
nfx: okay, let's do take two: let's extend the Py4J crawler from `TablesCrawler` and then we can avoid this `fetch_one` hack here. that way, we won't have to change the dependencies in other components that depend on tables crawler.

```
class FasterTableScanCrawler(TablesCrawler):
```

nfx (changes requested):

nfx (approved):
lgtm

PR #2837, opened 2024-10-04
Temporarily ignore `MANAGED` HMS tables on external storage location

HariGS-DB:
### Linked issues

Relates to #2838 

image

### Functionality

### Tests

- [ ] added unit tests
- [ ] added integration tests


FastLee (changes requested):
I'm good with the changes. We need to add a warning in the readme AND the log file that these are managed table and shouldn't be dropped from HMS.

nfx (approved):

PR #2856, opened 2024-10-07
Unit test updates: workspace client fixture (`ws`)

asnare:
## Changes

This PR introduces a `ws` fixture for the unit tests, and refactors some of the existing unit tests that were using their own custom fixtures.

We now have an upstream dependency (`pytester`) that provides a fixture of the same name, but it is intended for use with integration tests. Without `ws` as part of the local `conftest.py` for unit tests it is easy to inadvertently inject the upstream fixture (which doesn't work as intended.)

### Tests

- updated unit tests


nfx (approved):
lgtm

PR #2852, opened 2024-10-07
Replace `greenlit` with unknown library in integration tests

JCZuurmond:
## Changes
`greenlit` library is added to the known list in #2830, thus not unknown anymore

### Linked issues
Resolves #2845
Resolves #2844

### Tests
- [x] modified integration tests: `test_job_spark_python_task_linter_unhappy_path`

PR #2863, opened 2024-10-08
Upgrade minimum required version of `pytester`: 0.2.1 → 0.3.0

asnare:
## Changes

This PR updates the minimum required version of `pytester` from 0.2.1 to 0.3.0. As of #2852 our integration tests depends on changes introduced in databrickslabs/pytester#60 (and released with 0.3.0).


nfx (approved):
lgtm

PR #2885, opened 2024-10-09
Tech debt: fix situations where `next()` isn't being used properly

asnare:
## Changes

This PR updates/fixes the way Python's builtin [`next()`](https://docs.python.org/3/library/functions.html#next) function is used. Specifically, in a few places the code assumed that `None` will be returned if there is no next value but this is incorrect.

### DIFF
--- a/src/databricks/labs/ucx/source_code/notebooks/cells.py
+++ b/src/databricks/labs/ucx/source_code/notebooks/cells.py
@@ -289,9 +289,11 @@ class CellLanguage(Enum):
 
     @classmethod
     def of_language(cls, language: Language) -> CellLanguage:
-        # TODO: Should this not raise a ValueError if the language is not found?
-        #  It also  causes a GeneratorExit exception to be raised. Maybe an explicit loop is better.
-        return next((cl for cl in CellLanguage if cl.language == language))
+        for cell_language in CellLanguage:
+            if cell_language.language == language:
+                return cell_language
+        msg = f"Unsupported cell language: {language}"
+        raise ValueError(msg)
 
     @classmethod
     def of_magic_name(cls, magic_name: str) -> CellLanguage | None:
--- a/src/databricks/labs/ucx/source_code/path_lookup.py
+++ b/src/databricks/labs/ucx/source_code/path_lookup.py
@@ -87,9 +87,6 @@ class PathLookup:
             and any(subfolder.name.lower() == "egg-info" for subfolder in path.iterdir())
         )
 
-    def has_path(self, path: Path):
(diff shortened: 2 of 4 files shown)


src/databricks/labs/ucx/source_code/notebooks/cells.py
```diff
@@ -289,9 +289,11 @@ def requires_isolated_pi(self) -> str:
...
     @classmethod
     def of_language(cls, language: Language) -> CellLanguage:
-        # TODO: Should this not raise a ValueError if the language is not found?
-        #  It also  causes a GeneratorExit exception to be raised. Maybe an explicit loop is better.
-        return next((cl for cl in CellLanguage if cl.language == language))
+        for cell_language in CellLanguage:
+            if cell_language.language == language:
+                return cell_language
+        msg = f"Unsupported cell language: {language}"
+        raise ValueError(msg)
```
asnare, line 296: This would raise `StopIteration` if the language wasn't supported. The TODO has now been resolved: it raises `ValueError` instead.

src/databricks/labs/ucx/source_code/path_lookup.py
```diff
@@ -87,9 +87,6 @@ def _is_egg_folder(path: Path) -> bool:
             and any(subfolder.name.lower() == "egg-info" for subfolder in path.iterdir())
         )
 
-    def has_path(self, path: Path):
-        return next(p for p in self._sys_paths if path == p) is not None
-
```
asnare, line 92: This code assumed that `next()` would return `None` if no path matched, but it would instead raise `StopIteration`.

This method is not used anywhere, so I removed it.

tests/unit/hive_metastore/test_table_migrate.py
```diff
@@ -1028,7 +1028,7 @@ def test_migrate_views_should_be_properly_sequenced(ws):
     table_keys = [task.args[0].src.key for task in tasks]
     assert table_keys.index("hive_metastore.db1_src.v1_src") > table_keys.index("hive_metastore.db1_src.v3_src")
     assert table_keys.index("hive_metastore.db1_src.v3_src") > table_keys.index("hive_metastore.db1_src.v2_src")
-    assert next((key for key in table_keys if key == "hive_metastore.db1_src.t1_src"), None) is None
+    assert not any(key for key in table_keys if key == "hive_metastore.db1_src.t1_src")
```
asnare, line 1031: There was no problem with correctness here, it's merely a switch from:
```python
next(…, None) is None
```
to
```python
not any(…)
```

nfx (approved):
lgtm

PR #2912, opened 2024-10-10
Verify migration progress prerequisites during UCX catalog creation

### LINKED ISSUE #2816: [FEATURE] Skip the `migration-progress` run when the `assessment` job did not run yet
### Summary

Skip the `migration-process` run when the `assessment` job did not run yet.

Extra
- [x] Warn that the assessment needs to run during creation of the UCX catalog, if it has not yet already.
- [ ] ~~Remove running "assessment" from the integration `test_running_real_migration_progress_job` to reduce flakiness~~

### Implementation

At the start of the `migration-process` workflow:
- Check if the `assessment` job has run
- If not, stop the workflow
- If so, continue the workflow

JCZuurmond:
## Changes
Verify migration progress prerequisites during UCX catalog creation:
- UC metastore exists.
- UCX catalog exists.
- A job run corresponding to the "assessment" job:
  - Finished successfully.
  - OR if pending or running, we will wait up to 1 hour for the assessment run to finish. If did still not finish successfully, we fail.

Otherwise, we consider the prerequisites to be NOT matched.

### Linked issues

Resolves #2816

### Functionality

- [x] modified CLI command `create-ucx-catalog`

### Tests

- [x] manually tested
- [x] added unit tests


src/databricks/labs/ucx/progress/install.py
```diff
@@ -52,3 +55,50 @@ def run(self) -> None:
         self._schema_deployer.deploy_table("workflow_runs", WorkflowRun)
         self._schema_deployer.deploy_table("historical", Historical)
         logger.info("Installation completed successfully!")
+
+
+class VerifyProgressTracking:
```
JCZuurmond, line 60: @nfx : What do you think of these `Verify...` classes? They are small classes that mainly function as a wrapper around one method

JCZuurmond (commented):
See comment

src/databricks/labs/ucx/progress/install.py
nfx: ideally we'd merge those

nfx (approved):
lgtm

src/databricks/labs/ucx/progress/install.py
nfx: > VerifyHasMetastore and VerifyHasCatalog

let's start with merging those once we have cycles for it

PR #2927, opened 2024-10-10
Fixed an issue with source table deleted after migration

### LINKED ISSUE #2901: [BUG]: migrate_tables fails if source table is deleted after its been migrated

### Current Behavior

Faced this issue while re-running migrate-tables workflow. The destination table was already created in UC and during subsequent runs, the workflow failed (even when the table-mapping did not not have the table or view in 

### Expected Behavior

Even if this error comes up, log a warning and continue the workflow.

### Steps To Reproduce

Run table migration
Delete source table
Re-run table migration

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

```shell
NotFound: [TABLE_OR_VIEW_NOT_FOUND] The table or view cannot be found. Verify the spelling and correctness of the schema and catalog.
If you did not qualify the name with a schema, verify the current_schema() output, or qualify the name with the correct schema and catalog.
```

FastLee:
closes #2901

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/table_migration_status.py
+++ b/src/databricks/labs/ucx/hive_metastore/table_migration_status.py
@@ -108,15 +108,18 @@ class TableMigrationStatusRefresher(CrawlerBase[TableMigrationStatus]):
         return seen_tables
 
     def is_migrated(self, schema: str, table: str) -> bool:
-        results = self._backend.fetch(
-            f"SHOW TBLPROPERTIES {escape_sql_identifier(schema + '.' + table)} ('upgraded_to')"
-        )
-        for result in results:
-            if "does not have property" in result.value:
-                continue
-            logger.info(f"{schema}.{table} is set as migrated")
-            return True
-        logger.info(f"{schema}.{table} is set as not migrated")
+        try:
+            results = self._backend.fetch(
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/hive_metastore/table_migration_status.py
```diff
@@ -108,15 +108,19 @@ def get_seen_tables(self) -> dict[str, str]:
...
+            )
+            for result in results:
+                if "does not have property" in result.value:
+                    continue
+                logger.info(f"{schema}.{table} is set as migrated")
+                return True
+            logger.info(f"{schema}.{table} is set as not migrated")
+        except NotFound:
+            # If the source table doesn't exist, it will not be shown as migrated
+            logger.info(f"{schema}.{table} set as a source does no longer exist")
```
nfx, line 123:
```suggestion
            logger.warning(f"failed-to-migrate: {schema}.{table} set as a source does no longer exist")
            return True
```
1) log warning, so it gets to dashboard
2) if we don't return true, there'll be an attempt to migrate table, which will also fail, because there's nothing to migrate from

nfx (changes requested):

nfx (approved):
lgtm

PR #2928, opened 2024-10-10
Fix issue with migrating MANAGED hive_metastore table to UC

### LINKED ISSUE #2838: [BUG]: Migrate tables errors out when migrating managed table with external storage

### Current Behavior

When running a migrate table job for a table which is managed with an external location, the code fails with the error Description: [UPGRADE_NOT_SUPPORTED.NOT_EXTERNAL] Table is not eligible for an upgrade from Hive Metastore to Unity Catalog. Reason: Not an external table. SQLSTATE: 0AKUC

### Expected Behavior

For managed table with external storage location the sync command should run with additional clause as "AS EXTERNAL"

### Steps To Reproduce

_No response_

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

HariGS-DB:
## Changes
HMS MANAGED tables when deleted also delete their underlying data.
If an HMS-managed table is migrated to UC as EXTERNAL, dropping the HMS table will delete the underlying data file and render the UC table unusable, leading to a non-recoverable data loss.
Changing the MANAGED table to EXTERNAL may have consequences on regulatory data cleanup, as deleting the EXTERNAL table no longer deletes the underlying table. It would cause leakage of data when tables are dropped.
As with the case of duplicating the data, if new data is added to either HMS or UC, the other table goes out of sync requiring re-migration

Resolves #2838 

### Functionality

### Tests

- [ ] added unit tests
- [ ] added integration tests


src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -52,6 +55,7 @@ def __init__(
         self._seen_tables: dict[str, str] = {}
         self._migrate_grants = migrate_grants
         self._external_locations = external_locations
+        self._spark = SparkSession.builder.getOrCreate()
```
nfx, line 58: please isolate spark session creation only to the method that uses it. See this PR https://github.com/databrickslabs/ucx/pull/2947/files

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/table_migrate.py
HariGS-DB: i have moved this to a seperate property called _spark_session
but I am now calling this in migrate_tables so that the spark session instantiate when the value passed is "CONVERT_TO_EXTERNAL". also doing it at migrate_tables allows to create the session just once. calling it in child method would invoke this for each table migration.

src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -207,8 +223,58 @@ def _sql_migrate_view(self, src_view: ViewToMigrate) -> str:
...
+                old_table.schemaPreservesCase(),
+                old_table.ignoredProperties(),
+                old_table.viewOriginalText(),
+            )
+            self._catalog.alterTable(new_table)
+            logger.info(f"Converted {src_table.name} to External Table type.")
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            logger.warning(f"Error converting HMS table {src_table.name} to external: {e}", exc_info=True)
+            return None
+        return None
```
nfx, line 277:
```suggestion
            return False
        return True
```
and check the result

src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -123,7 +127,19 @@ def _migrate_views(self):
...
+    @cached_property
+    def _spark_session(self):
+        # pylint: disable-next=import-error,import-outside-toplevel
+        from pyspark.sql.session import SparkSession  # type: ignore[import-not-found]
+
+        return SparkSession.builder.getOrCreate()
+
     def _migrate_managed_table(self, managed_table_external_storage: str, src_table: TableToMigrate):
+        if managed_table_external_storage == 'CONVERT_TO_EXTERNAL':
+            self._convert_hms_table_to_external(src_table.src)
```
nfx, line 139: and if `_convert_hms_table_to_external` fails?

nfx (changes requested):

nfx (approved):
lgtm, though we're missing an integration test

PR #2929, opened 2024-10-10
Log warnings when mounts are discovered on incorrect cluster type

### LINKED ISSUE #2498: [BUG]: table migration job fails with DBUtilsCore.mounts() is not whitelisted

### Current Behavior

- Table migration job running in DBR 15.4 and shared cluster fails with table migration job fails with DBUtilsCore.mounts() is not whitelisted error

- Table migration job running in DBR 15.4 and single user cluster fails with Table migration job running in DBR 15.4 and shared cluster fails with

- Table migration job running in DBR 14.3 and shared cluster fails with no error in log4j

### Expected Behavior

Job succeeds with no manual intervention

### Steps To Reproduce

Execute table migration with CLI
after job fails remove cluster policy and run with different cluster configs: 14.3 shared and 15.4 single user

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

works on no-isolation clusters. table migration workflow should rely on `ucx.mounts` table and not make a call to `dbutils.fs.mounts()` from table migration cluster
image

FastLee:
closes #2498

### DIFF
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -650,7 +650,7 @@ def migrate_tables(
         try:
             workspace_context.verify_progress_tracking.verify()
         except RuntimeWarning:
-            logger.info(
+            logger.warning(
                 "We couldn't detect a successful run of the assessment workflow."
                 "The assessment workflow is a prerequisite for the migrate-tables workflow."
                 "It can be run by using the `ensure-assessment-run` command."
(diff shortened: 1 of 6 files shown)


nfx (changes requested):

nfx (approved):
hacky, but okay

src/databricks/labs/ucx/cli.py
```diff
@@ -646,6 +646,17 @@ def migrate_tables(
         workspace_contexts = _get_workspace_contexts(w, a, run_as_collection)
     for workspace_context in workspace_contexts:
         deployed_workflows = workspace_context.deployed_workflows
+
+        try:
+            workspace_context.verify_progress_tracking.verify()
```
JCZuurmond, line 651: @FastLee : A rename would wake sense as this is not the place to "verify progress tracking". Either scope it to check if the assessment has ran (like the warning says) or rename the class to something broader than "verify progress tracking"

PR #2920, opened 2024-10-10
Fixed absolute path normalisation in source code analysis

ericvergnaud:
## Changes
Workspace API does not support relative subpaths such as "/a/b/../c".
This PR fixes the issue by resolving workspace paths before calling the API.

### Linked issues
Resolves #2882 
Requires #https://github.com/databrickslabs/blueprint/pull/156
Requires #https://github.com/databrickslabs/blueprint/pull/157

### Functionality
None

### Tests
- [x] added integration tests

Issue #2882, opened 2024-10-09
[BUG]: assess_workflow failing with "Parsing Python code failed"


antoniopinheirofilho:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

assess_workflow fails with several "Parsing Python code failed" errors

### Expected Behavior

assess_workflow should not fail. Instead, it should flag the code/files that it can't parse

i think we need to pass absolute paths to API: `/Repos/NBA_PROD/nba-data-load-scripts/real-time/../../nba-common-scripts/nba_common_functions` -> `/Repos/NBA_PROD/nba-common-scripts/nba_common_functions`

### Cloud

Azure

### Operating System

macOS

### Version

latest via Databricks CLI

ericvergnaud:
@antoniopinheirofilho would you be able to share the failing notebook ? (presumably `/Repos/NBA_PREPROD/nba-data-load-scripts/i360-table-refresh-monitoring/../../nba-common-scripts/nba_common_functions`)

nfx:
@ericvergnaud i think we need to pass absolute paths to API: `/Repos/NBA_PROD/nba-data-load-scripts/real-time/../../nba-common-scripts/nba_common_functions` -> `/Repos/NBA_PROD/nba-common-scripts/nba_common_functions`

antoniopinheirofilho:
@ericvergnaud it appears that the path you mentioned doesn't exist.

@nfx here is the content of the "assess_workflow.log" file:

18:45:42 DEBUG [databricks.labs.ucx.framework.crawlers] {MainThread} [hive_metastore.ucx.directfs_in_paths] found 55 new records for directfs_in_paths
18:45:46 DEBUG [databricks.labs.ucx.framework.crawlers] {MainThread} [hive_metastore.ucx.used_tables_in_paths] found 1755 new records for used_tables_in_paths
18:45:50 ERROR [databricks.labs.ucx] {MainThread} Execute `databricks workspace export //Applications/ucx/logs/assessment/run-854607653748637-0/assess_workflows.log` locally to troubleshoot with more details. Detected 478 failures: AstroidSyntaxError: Parsing Python code failed:
cannot use starred expression here (<unknown>, line 76), AstroidSyntaxError: Parsing Python code failed:
invalid syntax (<unknown>, line 1), AstroidSyntaxError: Parsing Python code failed:
unexpected indent (<unknown>, line 5), AstroidSyntaxError: Parsing Python code failed:
[...]

PR #2950, opened 2024-10-14
Dependency update: ensure we install with at least version 0.9.1 of `databricks-labs-blueprint`

asnare:
## Changes

In #2920 we introduced a change to ensure that notebook paths are normalised, resolving #2882. This change depended on an upstream fix (databrickslabs/blueprint#157) included in the 0.9.1 release of the dependency. This PR ensures that we run against that release or later.


nfx (approved):
lgtm

PR #2953, opened 2024-10-14
Update the workspace content cache used during linting

asnare:
## Changes

This PR began as an effort to retrofit the `WorkspaceCache` (and associated implementation) with type hints to help ensure correctness. This proved to be quite difficult due to the dynamic nature of the cache and binary vs. text modes. As such, this PR updates the `WorkspaceCache` in the following ways:

 - We now only cache the binary content of workspace files. This simplifies the implementation substantially and avoids correctness issues with respect to various options that can be specified to control decoding as binary content. (Decoding is well optimised compared to upstream/downstream processing, so this is a reasonable tradeoff.)
 - The cache now resolves paths, ensuring that different ways of referring to the same content are handled as a single cache entry.
 - We now raise an error immediately if a non-absolute path is requested: workspace paths are always absolute, so this will catch errors earlier.
 - For the paths where we check for a BOM-marker prior to reading as text, if the opened stream for the path is seekable then we only open it once instead of twice.

### Tests

- added (and updated existing) unit tests
- existing integration tests


src/databricks/labs/ucx/mixins/cached_workspace_path.py
```diff
@@ -1,139 +1,138 @@
...
-    def __init__(self, ws: WorkspaceClient, max_entries=2048):
+    def __init__(self, ws: WorkspaceClient, max_entries: int = 2048) -> None:
         self._ws = ws
         self._cache = _PathLruCache(max_entries)
 
-    def get_path(self, path: str):
+    def get_workspace_path(self, path: str) -> WorkspacePath:
+        if not path.startswith("/"):
+            msg = f"Invalid workspace path; must be absolute and start with a slash ('/'): {path}"
+            raise ValueError(msg)
```
nfx, line 137: please catch this exception in all places in production code that call this method and return `MaybeGraph(None, [DependencyProblem(...)])`

src/databricks/labs/ucx/mixins/cached_workspace_path.py
```diff
@@ -1,139 +1,138 @@
...
-        return StringIO(self._content.decode("utf-8"))
+    def __init__(self, max_entries: int) -> None:
+        # Ordered from oldest to newest.
+        self._datas = OrderedDict()
+        self._max_entries = max_entries
 
+    @classmethod
+    def _normalize(cls, path: _CachedPath) -> PurePosixPath:
+        # Note: must not return the same instance that was passed in, to avoid circular references (and memory leaks).
+        return PurePosixPath(path.resolve())
```
nfx, line 35: and what happens if `.resolve()` fails?

src/databricks/labs/ucx/mixins/cached_workspace_path.py
```diff
@@ -1,139 +1,138 @@
...
-                msg = f"Cannot initialise without object path: {object_info}"
-                raise ValueError(msg)
-            child = _CachedPath(self._cache, self._ws, path)
-            yield child.with_object_info(object_info)
-
-    def open(
+            yield self._from_object_info_with_cache(self._cache, self._ws, object_info)
+
+    @classmethod
+    def _detect_encoding_bom(cls, data: io.BytesIO) -> str | None:
```
nfx, line 86: dupe of `databricks.labs.ucx.source_code.base.guess_encoding`.

nfx (changes requested):

src/databricks/labs/ucx/mixins/cached_workspace_path.py
asnare: I've updated all the call paths that (indirectly) call this method, and convert the exception raised here into `DependencyProblem`.

To avoid accidentally catching other triggers for `ValueError`, I updated it to a specific subclass.

src/databricks/labs/ucx/mixins/cached_workspace_path.py
asnare: That's a good question. It shouldn't happen: `WorkspacePath.resolve()` only fails when the path is not absolute (because `Workspace.cwd()` doesn't exist). However we now guard against non-absolute paths on the way in.

That said, now that you point it out I'm on the fence about keeping it. Removing it means we might end up double-caching the same thing, but I'm not even sure if that's possible for `WorkspacePath`.

Will remove…

src/databricks/labs/ucx/mixins/cached_workspace_path.py
asnare: Refactored, and updated it at the same time to avoid double-opening files.

src/databricks/labs/ucx/mixins/cached_workspace_path.py
asnare: … and removed.

nfx (approved):
lgtm. make sure the integration tests are passing

asnare:
> lgtm. make sure the integration tests are passing

Indeed; they've flagged a legitimate problem and I'm fixing it now.

PR #2954, opened 2024-10-14
Make `ExternalLocations` explicitly depend on `TablesCrawler` and `MountsCrawler`

nfx:
This PR streamlines resolution of tables and mounts into external locations. previously dependencies were not direct and explicit


nfx (commented):
...

PR #2955, opened 2024-10-14
Let `create-catalogs-schemas` reuse `MigrateGrants` so that it applies group renaming

### LINKED ISSUE #2932: [BUG]: `create-catalog-schemas` might fail when transferring ownership of the catalog

### Current Behavior

`create-catalog-schemas` might fail when transferring ownership of the catalog if the transferal of the ownership removes the privileges for the user running the command to apply pending grants.

For example,
- The table mapping contains entries mapping: `hive_metastore.database.table` to `catalog.database.table`
- User B has legacy `OWN` permissions on `hive_metastore.database` 
- User C has legacy `USAGE` permissions on `hive_metastore.database` 
- Above permissions are ordered like this list in the UCX inventory grants table.
- User A runs the `create-catalog-schemas` command, in UC
   1. The catalog `catalog` is created with ownership for user A
   1. The database `database` is created with ownership for user A
   1. The ownership of database `database` is transferred from user A to user B
   1. User C gets `USAGE` permission database `database` (executed by user A) - this might fail(?) as user A is still owner of the catalog but not of the database
   1. The ownership of catalog `catalog` is transferred from user A to user B
   2. User C **fails to** get `USAGE` permissions on catalog `catalog` as user A does not have the permissions to grant these privileges to user C (anymore).

Notes: 
- If the ownership is transferred (correctly) it might still become a problem when migrating tables as the permissions to create UC tables might be missing.
- If the ownership is **not** transferred this might become a data security breach as the user performing the migration receives ownership of data objects it maybe did not had access to before.

Linked issue:
#2890

### Expected Behavior

- The `create-catalog-schemas` command should not fail
- The end state should reflect the right ownership, while not blocking table migration

### Steps To Reproduce

See above

### Cloud

Azure

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

### LINKED ISSUE #2934: [BUG]: `create-catalog-schemas` does not apply possible group renaming when applying grants

### Current Behavior

When we migrate grants during `create-catalog-schemas` command: https://github.com/databrickslabs/ucx/blob/38bd88b3d26b9e944bf6ce04868bddb54145bb00/src/databricks/labs/ucx/hive_metastore/catalog_schema.py#L61-L62

We do not take group renaming into account as in `MigrateGrants` 

https://github.com/databrickslabs/ucx/blob/38bd88b3d26b9e944bf6ce04868bddb54145bb00/src/databricks/labs/ucx/hive_metastore/grants.py#L778

Meaning the command might fail when groups are renamed when moving workspace level groups to account level.

### Expected Behavior

The group renaming should be applied when migrating grants in the `create-catalogs-schemas` (or anywhere for that matter). I suggest to reuse the `MigrateGrants` in the `create-catalog-schemas` logic.

### Steps To Reproduce

N.A.

### Cloud

Azure

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

JCZuurmond:
## Changes
Let `create-catalogs-schemas` reuse `MigrateGrants` so that it applies group renaming and eliminate duplicate code.

### Linked issues

Resolves #2934
Resolves #2932

### Functionality

- [x] modified existing command: `databricks labs ucx create-catalogs-schemas`
- [x] modified existing workflow: `migrate-tables` (*functionally no change)

### Tests

- [ ] manually tested
- [x] added unit tests
- [x] added integration tests


src/databricks/labs/ucx/hive_metastore/catalog_schema.py
```diff
@@ -44,162 +40,81 @@ def create_ucx_catalog(self, prompts: Prompts, *, properties: dict[str, str] | N
...
-        for grant in grants:
-            acl_migrate_sql = grant.uc_grant_sql()
-            if acl_migrate_sql is None:
-                logger.warning(
-                    f"Skipping legacy grant that is not supported in UC: {grant.action_type} on {grant.this_type_and_key()}"
-                )
+        catalogs_existing, schemas_existing = self._catalogs_schemas_from_unity_catalog()
+        catalogs, schemas = self._catalogs_schemas_from_table_mapping()
+        for src_schema, dst_catalog in catalogs:
+            if dst_catalog in catalogs_existing:
```
JCZuurmond, line 49: @nfx and @FastLee : This introduces the question on "do we want to skip applying ACLs when the catalog exists?" 

Before this PR:
- ACLs might be applied to an existing catalog if a schema will be created within that catalog
- ACLs will be skipped for an existing schema.

FastLee: I would skip.
Let's just alert.

JCZuurmond: Skip what exactly? The ones for the catalog if the catalog exists, the ones for the schema if the schema exits, the ones for the catalog if the schema exits? All ACLs if any schema or catalog exists?

JCZuurmond: Currently implemented that **all** legacy grants are migrated to UC and created [an issue](https://github.com/databrickslabs/ucx/issues/2964) to make this part optional.

src/databricks/labs/ucx/hive_metastore/objects.py
```diff
@@ -0,0 +1,36 @@
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class Catalog:
```
nfx, line 5: ~this data structure seems to be used only in one place. why do we need to create whole new modules for those? also, it really is `tuple[str,str] = ("CATALOG", name)`~

might be a good idea to co-locate next to `Table` in `src/databricks/labs/ucx/hive_metastore/tables.py`. make sure to implement only the required interface methods.

src/databricks/labs/ucx/hive_metastore/grants.py
```diff
@@ -737,20 +752,22 @@ def __init__(
         self._group_manager = group_manager
         self._grant_loaders = grant_loaders
 
-    def apply(self, src: Table, uc_table_key: str) -> bool:
+    def apply(self, src: Catalog | Schema | Table, dst: Catalog | Schema | Table) -> bool:
```
nfx, line 755: it looks like you need a duck-typed protocol (a.k.a. interface-on-read, coming from Go):

```
class WithFullName(typing.Protocol):
    def full_name(self) -> str: ...
```

src/databricks/labs/ucx/hive_metastore/grants.py
```diff
@@ -765,16 +782,14 @@ def _grants(self) -> list[Grant]:
                 grants.append(grant)
         return grants
 
-    def _match_grants(self, table: Table) -> list[Grant]:
+    def _match_grants(self, src: Catalog | Schema | Table) -> list[Grant]:
```
nfx, line 785: same protocol here. https://typing.readthedocs.io/en/latest/spec/protocol.html#protocols

src/databricks/labs/ucx/hive_metastore/objects.py
```diff
@@ -0,0 +1,36 @@
+from dataclasses import dataclass
```
JCZuurmond, line 1: @nfx: Have a look at this module.

Notes:
- Needed to be a separate module to avoid circular imports
- Maybe we want to introduce a parent object [securable object](https://docs.databricks.com/en/sql/language-manual/sql-ref-privileges.html#securable-objects) (later), to have a common interface for securable objects
- Note some inconsistencies across this code base and including the sdk:
  - `SchemaInfo(catalog_name="catalog")` vs `Schema(catalog="catalog")`. I choose the later to be consistent with the already existing `Table` class
  - Should we have a `Schema` or `Database`?
  - When do we use `full_name` and when `key`?
  - More?

src/databricks/labs/ucx/hive_metastore/mapping.py
```diff
@@ -58,6 +58,10 @@ def from_src_dst(cls, src_table: TableInfo, dst_schema: SchemaInfo) -> "Rule":
     def match(self, table: TableIdentifier) -> bool:
         return table.catalog == "hive_metastore" and self.src_schema == table.schema and self.src_table == table.table
 
+    @property
+    def as_uc_table(self) -> Table:
+        return Table(self.catalog_name, self.dst_schema, self.dst_table)
+
     @property
     def as_uc_table_key(self) -> str:
```
JCZuurmond, line 66: Not for this PR, but would try to avoid using this  key and use the `Table` data class instead

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -53,8 +53,8 @@ class Table:  # pylint: disable=too-many-public-methods
     catalog: str
     database: str
     name: str
-    object_type: str
-    table_format: str
+    object_type: str | None = None
```
JCZuurmond, line 56: Note change on this data type to support `Rule.as_uc_table`

tests/integration/hive_metastore/test_catalog_schema.py
```diff
@@ -95,37 +150,35 @@ def test_create_catalog_schema_with_principal_acl_aws(
...
 def test_create_catalog_schema_with_legacy_acls(
-    ws, make_user, make_catalog, make_schema, make_mounted_location, runtime_ctx, sql_backend
-):
-    src_schema = make_schema(catalog_name="hive_metastore")
-    src_external_table = runtime_ctx.make_table(
-        catalog_name=src_schema.catalog_name,
-        schema_name=src_schema.name,
-        external_csv=make_mounted_location,
-    )
-    dst_catalog = make_catalog()
```
JCZuurmond, line 106: @FastLee : I removed this to test create a new catalog. Please review if the test is correct after my changes

tests/integration/hive_metastore/test_catalog_schema.py
```diff
@@ -3,21 +3,37 @@
...
+
+
+@retried(on=[NotFound], timeout=timedelta(seconds=20))
+def get_schema(ws: WorkspaceClient, full_name: str) -> SchemaInfo:
+    return ws.schemas.get(full_name)
+
+
+@retried(on=[NotFound], timeout=timedelta(seconds=20))
+def get_schema_permissions_list(ws: WorkspaceClient, full_name: str) -> PermissionsList:
+    return ws.grants.get(SecurableType.SCHEMA, full_name)
```
nfx, line 34: move this logic down to `CatalogSchema` class, so that you handle eventual consistency in production code, not in the test code. remember - CREATE-READ-READ is the pattern used in Terraform to handle these situations.

e.g. this makes the wrong thing more robust :)

tests/unit/hive_metastore/test_catalog_schema.py
```diff
@@ -10,126 +10,168 @@
...
 from databricks.labs.ucx.hive_metastore.mapping import TableMapping
-
-
-def prepare_test(ws, backend: MockBackend | None = None) -> CatalogSchema:
-    ws.catalogs.list.return_value = [CatalogInfo(name="catalog1")]
+from databricks.labs.ucx.workspace_access.groups import GroupManager
+
+
+def prepare_test(ws, backend: MockBackend | None = None) -> CatalogSchema:  # pylint: disable=too-complex
+    """Prepare tests with the following setup:
```
JCZuurmond, line 19: Did a clean up of the `prepare_tests` to verify it is testing what I expect it does

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/grants.py
JCZuurmond, line 755: This type change allows to match on a `Catalog`, `Schema` and `Table` not just table key.

JCZuurmond (commented):
See comments

src/databricks/labs/ucx/hive_metastore/objects.py
JCZuurmond: > Needed to be a separate module to avoid circular imports

Not a problem anymore after using a Protocol, https://github.com/databrickslabs/ucx/pull/2955#discussion_r1801182440

src/databricks/labs/ucx/hive_metastore/grants.py
```diff
@@ -109,6 +110,15 @@ def object_key(self) -> str:
...
         return key.lower()
 
+    @property
+    def order(self) -> int:
+        """Order of the grants to be upheld when applying."""
+        match self.action_type:
+            case "OWN":
+                return 1
+            case _:
+                return 0
```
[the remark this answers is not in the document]
JCZuurmond: Because I like to be fancy and show off 💁‍♂️ 

It is copied from: https://github.com/databrickslabs/ucx/blob/d3d08f9c3de178258150b0ea3fcd93cb4aedb836/src/databricks/labs/ucx/installer/workflows.py#L766-L776. 

Though, with just two values `0` and `1` it maybe does not matter

tests/integration/hive_metastore/test_catalog_schema.py
JCZuurmond: See the changes to the `CatalogSchema` class. 

Note that I went for a `(READ-)CREATE-READ` pattern:
- `READ` before `CREATE` to check if resource already exists.
- One `READ` after create because all settings can be set during the `CREATE` which is not always the case for Terraform resources. 

Also note that I kept `get_schema_permissions_list` as this uses a different API for `CREATE` (with SQL) and `READ` (with SDK/RestAPI). For now, I do not think it is worth it to implement a mapping between these two APIs to introduce the `CREATE-READ` pattern.

src/databricks/labs/ucx/hive_metastore/objects.py
JCZuurmond: Added the above as code comments

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -99,6 +99,7 @@ def is_hive(self) -> bool:
 
     @property
     def key(self) -> str:
+        # TODO: figure out what the difference is between key and full_name
```
JCZuurmond, line 102: @nfx, @william-conti and @ericvergnaud : (Tagging you as you had the most recent changes with this code and the `full_name` attribute.) I would like the code to be more consistent with Databricks naming conventions so that it is easier for (new) developers to understand the intent. 

As I see it:
- `key` should be `table_reference`, see [this](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-qry-select-table-reference.html). Or maybe we call it `sql_table_reference` to make clear it is intended to be used in Databricks SQL queries.
   - Synonym for `full_name` in most cases. 
   - Possible a reference to a location, as shown in the code.
- `full_name` aligns with Databricks SDK

nfx: we can change `key` to `full_name`, but in a separate PR

src/databricks/labs/ucx/hive_metastore/catalog_schema.py
```diff
@@ -229,18 +205,96 @@ def _validate_location(self, location: str) -> bool:
...
+            self._ws.catalogs.create(catalog.name, comment="Created by UCX", properties=properties)
         else:
             self._ws.catalogs.create(
-                catalog,
+                catalog.name,
                 storage_root=catalog_storage,
                 comment="Created by UCX",
                 properties=properties,
             )
+        catalog_created = self._get_catalog(catalog, timeout=dt.timedelta(seconds=30))
```
nfx, line 255:
```suggestion
        catalog_created = self._get_catalog(catalog, timeout=self._timeout)
```
inject this timeout via constructor, set it to 30 seconds in the context and None in unit tests. same for other calls

src/databricks/labs/ucx/hive_metastore/tables.py
JCZuurmond: https://github.com/databrickslabs/ucx/issues/2979

nfx (changes requested):
one last change and we're good to merge

src/databricks/labs/ucx/hive_metastore/tables.py
nfx: this will cause table migration issues. can we get without this change?

nfx (changes requested):

nfx (approved):
lgtm

PR #2956, opened 2024-10-14
Addressing Ownership Conflict when creating catalog/schemas

### LINKED ISSUE #2932: [BUG]: `create-catalog-schemas` might fail when transferring ownership of the catalog

FastLee:
Closes #2932

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/catalog_schema.py
+++ b/src/databricks/labs/ucx/hive_metastore/catalog_schema.py
@@ -75,8 +75,9 @@ class CatalogSchema:
                 continue
             logger.debug(f"Migrating acls on {grant.this_type_and_key()} using SQL query: {acl_migrate_sql}")
             self._backend.execute(acl_migrate_sql)
-        own_grants = [grant for grant in grants if grant.action_type == "OWN"]
-        for grant in own_grants:
+        for grant in grants:
+            if grant.action_type != "OWN":
+                continue
             own_acl_migrate_sql = grant.uc_grant_sql()
             logger.debug(f"Migrating ownership on {grant.this_type_and_key()} using SQL query: {own_acl_migrate_sql}")
             if own_acl_migrate_sql is None:
--- a/src/databricks/labs/ucx/hive_metastore/catalog_schema.py
+++ b/src/databricks/labs/ucx/hive_metastore/catalog_schema.py
@@ -63,7 +63,10 @@ class CatalogSchema:
 
     def _apply_from_legacy_table_acls(self) -> None:
         grants = self._get_catalog_schema_hive_grants()
+        # Scanning for non own grants first
         for grant in grants:
+            if grant.action_type == "OWN":
+                continue
             acl_migrate_sql = grant.uc_grant_sql()
             if acl_migrate_sql is None:
                 logger.warning(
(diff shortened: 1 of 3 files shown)


src/databricks/labs/ucx/hive_metastore/catalog_schema.py
```diff
@@ -72,6 +75,16 @@ def _apply_from_legacy_table_acls(self) -> None:
                 continue
             logger.debug(f"Migrating acls on {grant.this_type_and_key()} using SQL query: {acl_migrate_sql}")
             self._backend.execute(acl_migrate_sql)
+        own_grants = [grant for grant in grants if grant.action_type == "OWN"]
+        for grant in own_grants:
```
nfx, line 79:
```suggestion
        for grant in grants:
            if grant.action_type != "OWN":
                continue
```
can be simplified

src/databricks/labs/ucx/hive_metastore/catalog_schema.py
```diff
@@ -72,6 +75,16 @@ def _apply_from_legacy_table_acls(self) -> None:
...
+            own_acl_migrate_sql = grant.uc_grant_sql()
+            logger.debug(f"Migrating ownership on {grant.this_type_and_key()} using SQL query: {own_acl_migrate_sql}")
```
nfx, line 81: this debug statement does not make sense logically here, move it after the if-continue

tests/integration/conftest.py
```diff
@@ -1244,6 +1244,28 @@ def prepared_principal_acl(runtime_ctx, env_or_skip, make_mounted_location, make
     )
 
 
+@pytest.fixture
+def prepared_legacy_acl(runtime_ctx, make_mounted_location, make_catalog, make_schema):
```
nfx, line 1248: why do you need to introduce new fixture if it's used only once? it hinders readability

FastLee: I'm anticipating more tests with legacy ACLs. But will remove for now.

src/databricks/labs/ucx/hive_metastore/catalog_schema.py
FastLee: Of course, the if continue was added in the review.

src/databricks/labs/ucx/hive_metastore/catalog_schema.py
FastLee: Done.

nfx (approved):
lgtm

PR #2965, opened 2024-10-15
Use `LocationTrie` to infer a list of UC external locations

nfx:
This PR removes duplicate code for determining overlapping storage prefixes. this also lays down some groundwork for federated SQL connections

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/locations.py
+++ b/src/databricks/labs/ucx/hive_metastore/locations.py
@@ -42,9 +42,9 @@ class LocationTrie:
     """
 
     key: str = ""
-    parent: OptionalLocationTrie = None
+    parent: OptionalLocationTrie = dataclasses.field(repr=False, default=None)
     children: dict[str, "LocationTrie"] = dataclasses.field(default_factory=dict)
-    tables: list[Table] = dataclasses.field(default_factory=list)
+    tables: list[Table] = dataclasses.field(repr=False, default_factory=list)
 
     @cached_property
     def _path(self) -> list[str]:
@@ -57,9 +57,14 @@ class LocationTrie:
         return list(reversed(parts))[1:]
 
     @property
-    def location(self):
-        scheme, netloc, *path = self._path
-        return f"{scheme}://{netloc}/{'/'.join(path)}"
+    def location(self) -> str | None:
+        if not self.is_valid():
+            return None
+        try:
+            scheme, netloc, *path = self._path
+            return f"{scheme}://{netloc}/{'/'.join(path)}"
+        except ValueError:
+            return None
 
     @staticmethod
     def _parse_location(location: str | None) -> list[str]:
@@ -91,10 +96,14 @@ class LocationTrie:
 
     def is_valid(self) -> bool:
         """A valid location has a scheme and netloc; the path is optional."""
-        if len(self._path) < 3:
+        if len(self._path) < 2:
             return False
         scheme, netloc, *_ = self._path
         return scheme in _EXTERNAL_FILE_LOCATION_SCHEMES and len(netloc) > 0
+
+    def all_tables(self) -> Iterable[Table]:
+        for node in self:
+            yield from node.tables
 
     def has_children(self):
         return len(self.children) > 0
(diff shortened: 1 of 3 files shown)


src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -125,64 +164,59 @@ def __init__(
...
-                and not location.startswith("jdbc")
-            ):
-                self._dbfs_locations(external_locations, location, min_slash)
-            if location.startswith("jdbc"):
-                self._add_jdbc_location(external_locations, location, table)
-        return external_locations
+            queue.extend(curr.children.values())
+        return sorted(external_locations, key=lambda _: _.location)
+
+    def _resolve_location(self, table: Table) -> Table:
```
JCZuurmond, line 191: I think it is confusing that this `resolve_location` returns a `Table` with the resolved location on it. Where `resolve_mount` returns a `str` representing the resolved `mount`

JCZuurmond: Should this become part of the `Table` instead?

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -125,64 +164,59 @@ def __init__(
...
-            if location.startswith("dbfs:/mnt"):
-                location = self.resolve_mount(location)
-            if not location:
+            trie.insert(table)
+        queue = list(trie.children.values())
+        external_locations = []
+        while queue:
+            curr = queue.pop()
+            num_children = len(curr.children)  # 0 - take parent
+            if curr.location and (num_children > 1 or num_children == 0):
```
JCZuurmond, line 181: Isn't the following the same?
```suggestion
            if num_children != 1:
```

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -125,64 +164,59 @@ def __init__(
...
-                dupe = True
-            loc += 1
-        if not dupe:
-            external_locations.append(ExternalLocation(os.path.dirname(location) + "/", 1))
-
-    def _add_jdbc_location(self, external_locations, location, table):
-        dupe = False
+    def _resolve_jdbc(self, table: Table) -> str | None:
+        location = table.location
+        if not location or not table.storage_properties or not location.startswith('jdbc:'):
```
JCZuurmond, line 218: `location.startswith('jdbc:')` -> `table.is_jdbx()`?

tests/unit/hive_metastore/test_locations.py
```diff
@@ -212,18 +213,18 @@ def test_external_locations():
...
-    assert result_set[4].location == "jdbc:mysql://somemysql.us-east-1.rds.amazonaws.com:3306/test_db"
-    assert result_set[5].location == "jdbc:providerknown://somedb.us-east-1.rds.amazonaws.com:1234/test_db"
-    assert result_set[6].location == "jdbc:providerunknown://somedb.us-east-1.rds.amazonaws.com:1234/test_db"
+    assert crawler.snapshot() == [
+        ExternalLocation('gcs://test_location2/a/b', 1),
+        ExternalLocation(
+            'jdbc:databricks://dbc-test1-aa11.cloud.databricks.com;httpPath=/sql/1.0/warehouses/65b52fb5bd86a7be', 1
+        ),
+        ExternalLocation('jdbc:mysql://somemysql.us-east-1.rds.amazonaws.com:3306/test_db', 1),
+        ExternalLocation('jdbc:providerknown://somedb.us-east-1.rds.amazonaws.com:1234/test_db', 2),
```
JCZuurmond, line 222: There are three tables above that match this location, though, two are a collision. Is that intended? If so, could you add a comment that explains that.

src/databricks/labs/ucx/hive_metastore/locations.py
JCZuurmond: Also explain why we can simply extend when there is one child

src/databricks/labs/ucx/hive_metastore/locations.py
```diff
@@ -91,11 +119,22 @@ def find(self, table: Table) -> OptionalLocationTrie:
...
             return False
         scheme, netloc, *_ = self._path
+        if scheme.startswith('jdbc:') and len(netloc) > 0:
+            return True
         return scheme in _EXTERNAL_FILE_LOCATION_SCHEMES and len(netloc) > 0
 
+    def is_jdbc(self) -> bool:
+        if not self.is_valid():
+            return False
+        return self._path[0].startswith('jdbc:')
```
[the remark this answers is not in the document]
nfx: JDBC is a special kind of location

PR #2982, opened 2024-10-16
Return `False` when failed to migrate table

JCZuurmond:
Return `False` when failed to migrate.

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/table_migrate.py
+++ b/src/databricks/labs/ucx/hive_metastore/table_migrate.py
@@ -145,7 +145,7 @@ class TablesMigrator:
         if managed_table_external_storage == 'CLONE':
             return self._migrate_table_create_ctas(src_table.src, src_table.rule)
         logger.warning(f"failed-to-migrate: unknown managed_table_external_storage: {managed_table_external_storage}")
-        return True
+        return False
 
     def _safe_migrate_table(
         self,
--- a/src/databricks/labs/ucx/hive_metastore/table_migration_status.py
+++ b/src/databricks/labs/ucx/hive_metastore/table_migration_status.py
@@ -119,7 +119,8 @@ class TableMigrationStatusRefresher(CrawlerBase[TableMigrationStatus]):
                 return True
             logger.info(f"{schema}.{table} is set as not migrated")
         except NotFound:
-            # If the source table doesn't exist, it will not be shown as migrated
+            # If the source table doesn't exist anymore, we mark it as migrated to avoid trying to migrate it again and
+            # for views that have the table as dependency
             logger.warning(f"failed-to-migrate: {schema}.{table} set as a source does no longer exist")
             return True
         return False


JCZuurmond:
@FastLee : See the following code that you introduced in [PR](https://github.com/databrickslabs/ucx/pull/2927): 

https://github.com/databrickslabs/ucx/blob/22091746a993e8b65e681fe876f2b0b17fda78cd/src/databricks/labs/ucx/hive_metastore/table_migration_status.py#L121-L124

The comment contradicts the return statement, should it be `return False` instead?

JCZuurmond:
@nfx and @FastLee : In migrate grants we have: https://github.com/databrickslabs/ucx/blob/22091746a993e8b65e681fe876f2b0b17fda78cd/src/databricks/labs/ucx/hive_metastore/grants.py#L740-L754

I understand we skip some unmappable grants (e.g. READ_METADATA):

https://github.com/databrickslabs/ucx/blob/22091746a993e8b65e681fe876f2b0b17fda78cd/src/databricks/labs/ucx/hive_metastore/grants.py#L742-L748

However, should we consider the migration successful when a mappable grant can not be executed?

https://github.com/databrickslabs/ucx/blob/22091746a993e8b65e681fe876f2b0b17fda78cd/src/databricks/labs/ucx/hive_metastore/grants.py#L750-L753

I would argue the migration is unsuccessful, thus `return False`

FastLee:
> @nfx and @FastLee : In migrate grants we have:
> 
> https://github.com/databrickslabs/ucx/blob/22091746a993e8b65e681fe876f2b0b17fda78cd/src/databricks/labs/ucx/hive_metastore/grants.py#L740-L754
> 
> I understand we skip some unmappable grants (e.g. READ_METADATA):
> 
> https://github.com/databrickslabs/ucx/blob/22091746a993e8b65e681fe876f2b0b17fda78cd/src/databricks/labs/ucx/hive_metastore/grants.py#L742-L748
> 
> However, should we consider the migration successful when a mappable grant can not be executed?
> 
> https://github.com/databrickslabs/ucx/blob/22091746a993e8b65e681fe876f2b0b17fda78cd/src/databricks/labs/ucx/hive_metastore/grants.py#L750-L753
> 
> I would argue the migration is unsuccessful, thus `return False`

I would argue we have to keep it as True. We may want to consider ACL migration status.

FastLee (approved):
LGTM

FastLee:
> @FastLee : See the following code that you introduced in [PR](https://github.com/databrickslabs/ucx/pull/2927):
> 
> https://github.com/databrickslabs/ucx/blob/22091746a993e8b65e681fe876f2b0b17fda78cd/src/databricks/labs/ucx/hive_metastore/table_migration_status.py#L121-L124
> 
> The comment contradicts the return statement, should it be `return False` instead?

This is a case where the old table is removed. The migration status should be successful. Otherwise we will not be able to migrate views.

JCZuurmond:
> This is a case where the old table is removed. The migration status should be successful. Otherwise we will not be able to migrate views.

Oke, then I will update the comment as that is not what is says

PR #2984, opened 2024-10-16
Warn when table has column without no name during table migration

### LINKED ISSUE #2891: [BUG]: Cannot migrate table to UC when the table has a column without a name

### Current Behavior

UCX cannot migrate table to UC when the table has a column without a name

### Expected Behavior

TBD. Not supporting tables with columns that have no names is a UC restriction. It is allowed in the hive metastore, but the UI will not show it.

Options:
1. Migrate the tables with a CTAS renaming the empty column to a dummy value, like `__UNKNOWN_COLUMN_NAME__`.
2. More?

### Steps To Reproduce

Try migrating the following table:

```sql
CREATE TABLE AS hive_metastore.schema.table (`` id);
```

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

See the logs reported in #2544

JCZuurmond:
## Changes
Warn when table has column without no name during table migration. 

The code goes with a "do first, apologize later" approach to avoid an additional call to retrieve the column schema.

### Linked issues
Resolves #2891

### Functionality

- [x] modified existing workflow: `tables-migrate`

### Tests

- [x] added unit tests

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/table_migrate.py
+++ b/src/databricks/labs/ucx/hive_metastore/table_migrate.py
@@ -149,9 +149,9 @@ class TablesMigrator:
                 r'At columns.\d+: name "" is not a valid name`'
             )
             if re.match(pattern, str(e)):
-                logger.error(f"Cannot migrate table with empty column name: {src_table.src.key}", exc_info=e)
+                logger.warning(f"Cannot migrate table with empty column name: {src_table.src.key}", exc_info=e)
             else:
-                logger.error(f"Cannot migrate table: {src_table.src.key}", exc_info=e)
+                logger.warning(f"Cannot migrate table: {src_table.src.key}", exc_info=e)
             return False
 
     def _migrate_table(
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -131,15 +131,35 @@ def _migrate_managed_table(self, managed_table_external_storage: str, src_table:
...
+        try:
+            return self._migrate_table(src_table, managed_table_external_storage, hiveserde_in_place_migrate)
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            # Catching a Spark AnalysisException here, for which we do not have the dependency to catch explicitly
+            pattern = (  # See https://github.com/databrickslabs/ucx/issues/2891
+                r"INVALID_PARAMETER_VALUE: Invalid input: RPC CreateTable Field managedcatalog.ColumnInfo.name: "
+                r'At columns.\d+: name "" is not a valid name`'
+            )
+            if re.match(pattern, str(e)):
+                logger.error(f"Cannot migrate table with empty column name: {src_table.src.key}", exc_info=e)
```
nfx, line 152:
```suggestion
                logger.warning(f"Cannot migrate table with empty column name: {src_table.src.key}", exc_info=e)
```
otherwise job fails

JCZuurmond: I thought an error made sense in this case, but the other `failed-to-migrate` are also warning. I changed the code

nfx (approved):
lgtm

Issue #1415, opened 2024-04-16
[FEATURE]: Integration HMS lineage into DependencyGraph


rwforest:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Table mapping does not solve everything, chances are there's still error after migration. Since HMS lineage is there, UCX should target merging with dependencyGraph

### Proposed Solution

Merge HMS lineage with dependencyGraph. While it is dependent on the version of the DBR, it should start with the highest runtime and then backfill anything that's not captured using other means. Static lineage parsing or Spark listener.

nfx:
@rwforest The request is not clear. Please elaborate.

rwforest:
@nfx When doing table replacement, source to target table is only the first pass. How do we guarantee that anything will run successfully? That's the goal of the migration. So the goal here is, for example, given 1000 notebooks, I need to be able to understand which chain of commands are the most critical. In terms of dependencyGraph we can check the incoming and outgoing edges. If we know a notebook is an orphan, then we don't attempt to fix it. Right now even if you fix all the cells, it is only the beginning of all the errors that will surface eventually.

JCZuurmond:
@rwforest : Could you clarify the following:
1. What is the HMS dependency graph?
2. Why is this dependent on the Databricks runtime?

Generally speaking, I would like capability to plan code migrations using UCX based on a dependency graph of the (to-be-migrated) tables linted from the code

rwforest:
@JCZuurmond there's a feature in Databricks that never made it to public I think it was to build run time dataset dependency graph, and I was told by @FastLee that it is dependent on the DBR version. I believe the linted code is static lineage but I got some notebooks that are heavily parameterized. 

And I agree for the planning part. Is there a roadmap on some planning UI? I can't imagine how someone would plan code migration using csv.

JCZuurmond:
A planning UI is not on the roadmap. We are considering to include this issue in our upcoming planning, but no decision yet

PR #2999, opened 2024-10-17
More ownership classes

ericvergnaud:
## Changes
Existing Ownership classes deal with inventory objects.
We sometimes also need to get ownership from workspace objects.
This PR implements these for `Job` and `ClusterDetails`

### Linked issues
Progresses #1415 

### Functionality
None

### Tests
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/assessment/clusters.py
+++ b/src/databricks/labs/ucx/assessment/clusters.py
@@ -192,6 +192,16 @@ class ClusterInfoOwnership(Ownership[ClusterInfo]):
         return record.creator
 
 
+class ClusterDetailsOwnership(Ownership[ClusterDetails]):
+    """Determine ownership of clusters in the workspace.
+
+    This is the cluster creator (if known).
+    """
+
+    def _maybe_direct_owner(self, record: ClusterDetails) -> str | None:
+        return record.creator_user_name
+
+
 @dataclass
 class PolicyInfo:
     policy_id: str
--- a/src/databricks/labs/ucx/assessment/clusters.py
+++ b/src/databricks/labs/ucx/assessment/clusters.py
@@ -182,7 +182,7 @@ class ClustersCrawler(CrawlerBase[ClusterInfo], CheckClusterMixin):
             yield ClusterInfo(*row)
 
 
-class ClusterOwnership(Ownership[ClusterInfo]):
+class ClusterInfoOwnership(Ownership[ClusterInfo]):
     """Determine ownership of clusters in the inventory.
 
     This is the cluster creator (if known).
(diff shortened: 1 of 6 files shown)


tests/integration/assessment/test_clusters.py
```diff
@@ -76,9 +77,12 @@ def test_cluster_ownership(ws, runtime_ctx, make_cluster, make_user, inventory_s
 
     # Verify ownership is as expected.
     administrator_locator = runtime_ctx.administrator_locator
-    ownership = ClusterOwnership(administrator_locator)
-    assert ownership.owner_of(my_cluster_record) == ws.current_user.me().user_name
-    assert ownership.owner_of(their_cluster_record) == another_user.user_name
+    info_ownership = ClusterInfoOwnership(administrator_locator)
+    assert info_ownership.owner_of(my_cluster_record) == ws.current_user.me().user_name
```
asnare, line 81: `ws.current_user.me()` is a REST call, so it's probably polite to factor that out as a local variable rather than call it twice.

tests/unit/assessment/test_clusters.py
```diff
@@ -185,27 +186,48 @@ def test_unsupported_clusters():
...
+    assert owner == "bob"
+    admin_locator.get_workspace_administrator.assert_not_called()
+
+
+def test_cluster_details_owner_creator_unknown() -> None:
+    admin_locator = create_autospec(AdministratorLocator)
+    admin_locator.get_workspace_administrator.return_value = "an_admin"
+
+    ownership = ClusterDetailsOwnership(admin_locator)
+    owner = ownership.owner_of(ClusterDetails(cluster_id="1"))
```
asnare, line 225: I know the default for `creator_name` is `None`, but I think I prefer to explicitly provide it because it's an important detail for what we're testing.

tests/unit/assessment/test_jobs.py
```diff
@@ -135,22 +135,43 @@ def test_job_run_crawler(jobruns_ids, cluster_ids, run_ids, failures):
...
+    assert owner == "bob"
+    admin_locator.get_workspace_administrator.assert_not_called()
+
+
+def test_job_owner_creator_unknown() -> None:
+    admin_locator = create_autospec(AdministratorLocator)
+    admin_locator.get_workspace_administrator.return_value = "an_admin"
+
+    ownership = JobOwnership(admin_locator)
+    owner = ownership.owner_of(Job(job_id=1))
```
asnare, line 174: Likewise here with `creator_user_name=None`.

asnare (approved):
LGTM; a few minor quibbles but more questions of preference than anything blocking.

src/databricks/labs/ucx/assessment/clusters.py
```diff
@@ -192,6 +192,16 @@ def _maybe_direct_owner(self, record: ClusterInfo) -> str | None:
         return record.creator
 
 
+class ClusterDetailsOwnership(Ownership[ClusterDetails]):
```
nfx, line 195: where is it coming from? we don't have a concept of `cluster details` in our object model - we just have a cluster.

nfx (changes requested):

PR #3004, opened 2024-10-17
Remove a dead (ownership-related) class

asnare:
## Changes

This PR trivially removes a dead class; this was detritus left over from an intermediate version of the ownership mechanism that limited `Record` to `dataclass` classes. However this limitation was removed prior to the original code being merged (and the mechanism can be used with any class).


nfx (approved):
lgtm

PR #2980, opened 2024-10-16
Implement migration sequencing (phase 1)

ericvergnaud:
## Changes
Users need the ability to sequence migration tasks
This PR implements sequencing for workflow tasks, jobs and clusters.
It will be followed by PRs for other workspace items 

### Linked issues
Progresses #1415 

### Functionality
None

### Tests
- [x] added unit tests

PR #3008, opened 2024-10-17
Add `MigrationSequencer` for jobs

ericvergnaud:
## Changes
Add a `MigrationSequencer` class to sequence the migration steps for jobs. 

The PR includes the following resources in its sequence:
- Jobs
- Job tasks
- Job tasks dependencies
- Job clusters
- Cluster

Other elements part of the sequence are added later

### Linked issues
Progresses #1415 
Supersedes #2980 

### Tests
- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/assessment/sequencing.py
+++ b/src/databricks/labs/ucx/assessment/sequencing.py
@@ -13,7 +13,7 @@ from databricks.sdk.service import jobs
 from databricks.labs.ucx.assessment.clusters import ClusterOwnership, ClusterInfo
 from databricks.labs.ucx.assessment.jobs import JobOwnership, JobInfo
 from databricks.labs.ucx.framework.owners import AdministratorLocator
-from databricks.labs.ucx.source_code.graph import DependencyGraph, DependencyProblem
+from databricks.labs.ucx.source_code.graph import DependencyProblem
 
 
 @dataclass
--- a/src/databricks/labs/ucx/assessment/sequencing.py
+++ b/src/databricks/labs/ucx/assessment/sequencing.py
@@ -214,6 +214,7 @@ class MigrationSequencer:
             object_owner=parent.object_owner,  # No task owner so use parent job owner
         )
         self._nodes[task_node.key] = task_node
+        # `task.new_cluster` is not handled because it is part of the task and not a separate node
         if task.existing_cluster_id:
             maybe_cluster_node = self._register_cluster(task.existing_cluster_id)
             problems.extend(maybe_cluster_node.problems)
@@ -226,7 +227,6 @@ class MigrationSequencer:
             else:
                 problem = DependencyProblem('cluster-not-found', f"Could not find cluster: {task.job_cluster_key}")
                 problems.append(problem)
-        # TODO: Handle `task.new_cluster`
         return MaybeMigrationNode(task_node, problems)
 
     def _register_job_cluster(self, cluster: jobs.JobCluster, job: jobs.Job) -> MaybeMigrationNode:
--- a/src/databricks/labs/ucx/assessment/sequencing.py
+++ b/src/databricks/labs/ucx/assessment/sequencing.py
@@ -157,6 +157,16 @@ class MigrationSequencer:
         self._outgoing: dict[MigrationNodeKey, set[MigrationNode]] = defaultdict(set)
 
     def register_job(self, job: jobs.Job) -> MaybeMigrationNode:
+        """Register a job.
+
+        Args:
+            job (jobs.Job) : The job to register.
+
+        Returns:
+            MaybeMigrationNode : A maybe migration node, which has the migration node if no problems occurred during
+                registering. Otherwise, the maybe migration node contains the dependency problems occurring during
+                registering the job.
+        """
         problems = []
         job_node = self._nodes.get(("JOB", str(job.job_id)), None)
         if job_node:
--- a/src/databricks/labs/ucx/assessment/sequencing.py
+++ b/src/databricks/labs/ucx/assessment/sequencing.py
@@ -193,6 +193,17 @@ class MigrationSequencer:
                 problems.extend(maybe_task_node.problems)
                 if maybe_task_node.node:
                     self._outgoing_references[job_node.key].add(maybe_task_node.node)
+            # Only after registering all tasks, we can resolve the task dependencies
+            for task in job.settings.tasks or []:
+                task_key = ("TASK", f"{job.job_id}/{task.task_key}")
+                for task_dependency in task.depends_on or []:
+                    task_dependency_key = ("TASK", f"{job.job_id}/{task_dependency.task_key}")
+                    maybe_task_dependency = self._nodes.get(task_dependency_key)
+                    if maybe_task_dependency:
+                        self._outgoing_references[task_key].add(maybe_task_dependency)
+                    else:
+                        problem = DependencyProblem('task-dependency-not-found', f"Could not find task: {task_dependency}")
+                        problems.append(problem)
         return MaybeMigrationNode(job_node, problems)
 
     def _register_workflow_task(self, task: jobs.Task, parent: MigrationNode) -> MaybeMigrationNode:
--- a/src/databricks/labs/ucx/assessment/sequencing.py
+++ b/src/databricks/labs/ucx/assessment/sequencing.py
@@ -157,9 +157,10 @@ class MigrationSequencer:
         self._outgoing: dict[MigrationNodeKey, set[MigrationNode]] = defaultdict(set)
 
     def register_job(self, job: jobs.Job) -> MaybeMigrationNode:
+        problems = []
         job_node = self._nodes.get(("JOB", str(job.job_id)), None)
         if job_node:
-            return MaybeMigrationNode(job_node, [])
+            return MaybeMigrationNode(job_node, problems)
         job_name = job.settings.name if job.settings and job.settings.name else str(job.job_id)
         job_node = MigrationNode(
             node_id=next(self._counter),
@@ -176,9 +177,10 @@ class MigrationSequencer:
                     self._outgoing[job_node.key].add(maybe_cluster_node.node)
             for task in job.settings.tasks or []:
                 maybe_task_node = self._register_workflow_task(task, job)
+                problems.extend(maybe_task_node.problems)
                 if maybe_task_node.node:
                     self._outgoing[job_node.key] = maybe_task_node.node
-        return MaybeMigrationNode(job_node, [])
+        return MaybeMigrationNode(job_node, problems)
 
     def _register_workflow_task(self, task: jobs.Task, parent: MigrationNode) -> MaybeMigrationNode:
         """Register a workflow task.
@@ -189,10 +191,11 @@ class MigrationSequencer:
             parent : MigrationNode
                 The migration node for the parent job
         """
+        problems = []
         task_id = f"{parent.key}/{task.task_key}"
         task_node = self._nodes.get(("TASK", task_id), None)
         if task_node:
-            return MaybeMigrationNode(task_node, [])
+            return MaybeMigrationNode(task_node, problems)
         task_node = MigrationNode(
             node_id=next(self._counter),
             object_type="TASK",
@@ -203,10 +206,11 @@ class MigrationSequencer:
         self._nodes[task_node.key] = task_node
         if task.existing_cluster_id:
             maybe_cluster_node = self._register_cluster(task.existing_cluster_id)
+            problems.extend(maybe_cluster_node.problems)
             if maybe_cluster_node.node:
                 self._outgoing[task_node.key].add(maybe_cluster_node.node)
         # TODO: register `job_cluster_key
-        return MaybeMigrationNode(task_node, [])
+        return MaybeMigrationNode(task_node, problems)
 
     def _register_job_cluster(self, cluster: jobs.JobCluster) -> MaybeMigrationNode:
         if cluster.new_cluster:
--- a/src/databricks/labs/ucx/assessment/sequencing.py
+++ b/src/databricks/labs/ucx/assessment/sequencing.py
@@ -238,6 +238,7 @@ class MigrationSequencer:
             # Job clusters are ephemeral and exist during a job run for a specific job only
             object_owner=JobOwnership(self._admin_locator).owner_of(JobInfo.from_job(job)),
         )
+        self._nodes[cluster_node.key] = cluster_node
         return MaybeMigrationNode(cluster_node, [])
 
     def _register_cluster(self, cluster_id: str) -> MaybeMigrationNode:
--- a/src/databricks/labs/ucx/assessment/sequencing.py
+++ b/src/databricks/labs/ucx/assessment/sequencing.py
@@ -270,21 +270,20 @@ class MigrationSequencer:
           leaf during processing)
         - We handle cyclic dependencies (implemented in PR #3009)
         """
-        # pre-compute incoming keys for best performance of self._required_step_ids
-        incoming = self._invert_outgoing_to_incoming()
-        queue = self._create_node_queue(incoming)
-        seen = set[MigrationNode]()
-        node = queue.get()
         ordered_steps: list[MigrationStep] = []
+        incoming = self._invert_outgoing_to_incoming()  # For updating the priority of steps that depend on other steps
+        seen = set[MigrationNode]()
+        queue = self._create_node_queue(self._outgoing)
+        node = queue.get()
         while node is not None:
-            step = node.as_step(len(ordered_steps), sorted(n.node_id for n in incoming[node.key]))
+            step = node.as_step(len(ordered_steps), sorted(n.node_id for n in self._outgoing[node.key]))
             ordered_steps.append(step)
             seen.add(node)
             # Update the queue priority as if the migration step was completed
-            for dependency in self._outgoing[node.key]:
+            for dependency in incoming[node.key]:
                 if dependency in seen:
                     continue
-                priority = len(incoming[dependency.key] - seen)
+                priority = len(self._outgoing[dependency.key] - seen)
                 queue.put(priority, dependency)
             node = queue.get()
         return ordered_steps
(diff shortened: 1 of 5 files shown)


src/databricks/labs/ucx/sequencing/sequencing.py
```diff
@@ -0,0 +1,167 @@
...
+            object_name=object_name,
+            object_owner=ClusterOwnership(self._admin_locator).owner_of(ClusterInfo.from_cluster_details(details)),
+        )
+        self._nodes[cluster_node.key] = cluster_node
+        # TODO register warehouses and policies
+        return cluster_node
+
+    def generate_steps(self) -> Iterable[MigrationStep]:
+        # algo adapted from Kahn topological sort. The main differences is that
+        # we want the same step number for all nodes with same dependency depth
```
nfx, line 136: this requirement is not strictly necessary. also, having a queue here is a more common form and more maintainable by others.

src/databricks/labs/ucx/sequencing/sequencing.py
```diff
@@ -0,0 +1,167 @@
...
+
+@dataclass
+class MigrationStep:
+    step_id: int
+    step_number: int
+    object_type: str
+    object_id: str
+    object_name: str
+    object_owner: str
+    required_step_ids: list[int]
```
nfx, line 24: do we need this `required_step_ids` for vizualizations?

src/databricks/labs/ucx/sequencing/sequencing.py
```diff
@@ -0,0 +1,167 @@
...
+
+class MigrationSequencer:
+
+    def __init__(self, ws: WorkspaceClient, admin_locator: AdministratorLocator):
+        self._ws = ws
+        self._admin_locator = admin_locator
+        self._last_node_id = 0
+        self._nodes: dict[tuple[str, str], MigrationNode] = {}
+        self._incoming: dict[tuple[str, str], set[tuple[str, str]]] = defaultdict(set)
+        self._outgoing: dict[tuple[str, str], set[tuple[str, str]]] = defaultdict(set)
```
nfx, line 59: why do we need two adjacency sets? why can't we do it with a single one?

tests/unit/sequencing/test_sequencing.py
```diff
@@ -0,0 +1,34 @@
...
+    sequencer.register_workflow_task(task, job, graph)
+    steps = list(sequencer.generate_steps())
+    step = steps[-1]
+    assert step.step_id
+    assert step.object_type == "CLUSTER"
+    assert step.object_id == "cluster-123"
+    assert step.object_name == "my-cluster"
+    assert step.object_owner == "John Doe"
+    assert step.step_number == 3
+    assert len(step.required_step_ids) == 2
```
nfx, line 34:
```suggestion
    assert steps == [
      MigrationStep(...),
      MigrationStep(...),
    ]
```
assert on concrete records and concrete sequence. absence of this pattern led to numerous bugs already.

nfx (changes requested):

src/databricks/labs/ucx/sequencing/sequencing.py
ericvergnaud: Yes for the user to understand why a given step cannot be immediately migrated

src/databricks/labs/ucx/sequencing/sequencing.py
ericvergnaud: it's not strictly needed but it's much easier to debug, notably during population of incoming counts

src/databricks/labs/ucx/sequencing/sequencing.py
```diff
@@ -0,0 +1,167 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
ericvergnaud: done

src/databricks/labs/ucx/sequencing/sequencing.py
```diff
@@ -0,0 +1,167 @@
...
+        if job_node:
+            return job_node
+        self._last_node_id += 1
+        job_name = job.settings.name if job.settings and job.settings.name else str(job.job_id)
+        job_node = MigrationNode(
+            node_id=self._last_node_id,
+            object_type="JOB",
+            object_id=str(job.job_id),
+            object_name=job_name,
+            object_owner=JobOwnership(self._admin_locator).owner_of(JobInfo.from_job(job)),
```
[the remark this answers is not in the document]
ericvergnaud: Since @asnare is revisiting the design of xxxOwnership classes, isn't it better to wait for that to be done ? (Some of these classes may disappear altogether...)

tests/unit/sequencing/test_sequencing.py
ericvergnaud: done

src/databricks/labs/ucx/sequencing/sequencing.py
nfx: step number is still relative for all owners. we can compute it on-read in a widget via [ROW_NUMBER() window function](https://docs.databricks.com/en/sql/language-manual/functions/row_number.html) e.g. `SELECT *, ROW_NUMBER() OVER (PARTITION BY owner ORDER BY global_sequence_number) AS owner_step_number`. Still, the step number is not globally deterministic and can differ depending on the job task list order. that order is not deterministic on the API side. 

i don't want to tradeoff clarity here.

TLDR: algo implementation here has to be more straightforward for maintainability reasons and the sequence number we store in a database has to be global (e.g. `global_sequence_number`), because we can compute `owner_sequence_number` on read via window function.

please iterate on this quicker.

src/databricks/labs/ucx/sequencing/sequencing.py
nfx: let's remove it, as it may confuse the rest of maintainers.

tests/unit/sequencing/test_sequencing.py
nfx: how many steps do we have? this assert has to be on a list, not on a `step[-1]`. what happens if we yield more steps than necessary?

ericvergnaud: 'how many steps' is not what's being tested here. We're checking that the ultimate step is indeed the cluster.

src/databricks/labs/ucx/sequencing/sequencing.py
ericvergnaud: done (btw `_outgoing` is necessary, it's `_incoming` that's not). Re the latter, I now keep a transient computed list for performance.

src/databricks/labs/ucx/assessment/sequencing.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        # pre-compute incoming keys for best performance of self._required_step_ids
+        incoming_keys = self._collect_incoming_keys()
+        incoming_counts = self.compute_incoming_counts(incoming_keys)
+        step_number = 1
+        sorted_steps: list[MigrationStep] = []
+        while len(incoming_counts) > 0:
+            leaf_keys = list(self._get_leaf_keys(incoming_counts))
+            for leaf_key in leaf_keys:
+                del incoming_counts[leaf_key]
```
nfx, line 145: please rewrite with a more known queue implementation

nfx (changes requested):

src/databricks/labs/ucx/sequencing/sequencing.py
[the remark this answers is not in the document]
ericvergnaud: Done in PR #3076

src/databricks/labs/ucx/assessment/sequencing.py
JCZuurmond: done, needed to use the example from heapqueue to update priorities, see link in class docstring

tests/unit/assessment/test_sequencing.py
```diff
@@ -0,0 +1,74 @@
...
+from databricks.labs.ucx.framework.owners import AdministratorLocator, AdministratorFinder
+from databricks.labs.ucx.source_code.base import CurrentSessionState
+from databricks.labs.ucx.source_code.graph import DependencyGraph
+from databricks.labs.ucx.source_code.jobs import WorkflowTask
+
+
+def test_sequence_steps_from_job_task_with_cluster(ws, simple_dependency_resolver, mock_path_lookup) -> None:
+    """Sequence a job with a task referencing a cluster.
+
+    Sequence:  # TODO: @JCZuurmond: Would expect cluster first.
```
JCZuurmond, line 16: @nfx and @ericvergnaud : What order do we expect?

If I would go about it, I would:
1. Cluster
2. Task
3. Job

Not sure though, depends a bit on the changes the cluster requires

src/databricks/labs/ucx/assessment/sequencing.py
```diff
@@ -0,0 +1,268 @@
...
+        - We handle cyclic dependencies (implemented in PR #3009)
+        """
+        # pre-compute incoming keys for best performance of self._required_step_ids
+        incoming = self._invert_outgoing_to_incoming()
+        queue = self._create_node_queue(incoming)
+        seen = set[MigrationNode]()
+        node = queue.get()
+        ordered_steps: list[MigrationStep] = []
+        while node is not None:
+            step = node.as_step(len(ordered_steps), sorted(n.node_id for n in incoming[node.key]))
```
JCZuurmond, line 242: Currently, the step number is redundant as it is the position in the steps list

src/databricks/labs/ucx/assessment/sequencing.py
```diff
@@ -0,0 +1,268 @@
...
+            object_type="CLUSTER",
+            object_id=cluster_id,
+            object_name=object_name,
+            object_owner=ClusterOwnership(self._admin_locator).owner_of(ClusterInfo.from_cluster_details(details)),
+        )
+        self._nodes[cluster_node.key] = cluster_node
+        # TODO register warehouses and policies
+        return cluster_node
+
+    def generate_steps(self) -> Iterable[MigrationStep]:
```
JCZuurmond, line 226: Add integration test

JCZuurmond: - Unhappy
- Happy wo cycle
- Cycle
-

JCZuurmond (changes requested):
In separate PR:
- Release gradually to the assessment workflow, maybe use a feature flag from the configuration which is disable by default
- Explain above in the migration sequencing dashboard
- Add to migration process as well

JCZuurmond:
Extending test coverage, moving to in progress for now

src/databricks/labs/ucx/assessment/sequencing.py
```diff
@@ -0,0 +1,261 @@
...
+        self._last_node_id = 0
+        self._nodes: dict[MigrationNodeKey, MigrationNode] = {}
+        self._outgoing: dict[MigrationNodeKey, set[MigrationNode]] = defaultdict(set)
+
+    def register_workflow_task(self, task: jobs.Task, job: jobs.Job, _graph: DependencyGraph) -> MigrationNode:
+        task_id = f"{job.job_id}/{task.task_key}"
+        task_node = self._nodes.get(("TASK", task_id), None)
+        if task_node:
+            return task_node
+        job_node = self.register_workflow_job(job)
```
JCZuurmond, line 153: @ericvergnaud : Why do we go from the task to the job instead of the other way around?

src/databricks/labs/ucx/assessment/sequencing.py
```diff
@@ -0,0 +1,313 @@
...
+        self._admin_locator = admin_locator
+        self._counter = itertools.count()
+        self._nodes: dict[MigrationNodeKey, MigrationNode] = {}
+
+        # Outgoing references contains edges in the graph pointing from a node to a set of nodes that the node
+        # references. These references follow the API references, e.g. a job contains tasks in the
+        # `jobs.Job.settings.tasks`, thus a job has an outgoing reference to each of those tasks.
+        self._outgoing_references: dict[MigrationNodeKey, set[MigrationNode]] = defaultdict(set)
+
+    def register_job(self, job: jobs.Job) -> MaybeMigrationNode:
```
JCZuurmond, line 163: @nfx : Please comment on the signature. Do you think this is the right one? An other option would be to add an external facing `register_jobs()` that loops over all jobs it can find and then calls this method (which we can make protected). 

Or we could rename this to a `JobMigrationSequencer` with a `register` and `generate_steps` public method. Later we can add a `DashboardMigrationSequencer` and abstract away a parent `Sequencer` class enforcing this API

@ericvergnaud : You added the dependency graph here. Why did you do that? It was not used (yet)

JCZuurmond (commented):
@nfx: I highlighted the most important changes in the comments below

tests/integration/assessment/test_sequencing.py
```diff
@@ -0,0 +1,51 @@
...
+    task = jobs.Task(task_key="test-task", existing_cluster_id="non-existing-id")
+    settings = jobs.JobSettings(name="test-job", tasks=[task])
+    job = jobs.Job(job_id=1234, settings=settings)
+
+    maybe_job_node = runtime_ctx.migration_sequencer.register_job(job)
+    assert maybe_job_node.failed
+
+    steps = runtime_ctx.migration_sequencer.generate_steps()
+    step_object_types = [step.object_type for step in steps]
+    assert step_object_types == ["CLUSTER", "TASK", "JOB"]  # TODO: What do we expect?
```
JCZuurmond, line 51: Same question on what todo when a cluster is not resolved

src/databricks/labs/ucx/assessment/sequencing.py
```diff
@@ -0,0 +1,349 @@
...
+                problem = DependencyProblem('cluster-not-found', f"Could not find cluster: {task.job_cluster_key}")
+                problems.append(problem)
+        return MaybeMigrationNode(task_node, problems)
+
+    def _register_job_cluster(self, cluster: jobs.JobCluster, parent: MigrationNode) -> MaybeMigrationNode:
+        """Register a job cluster.
+
+        A job cluster is defined within a job and therefore is found when defined on the job by definition.
+        """
+        # Different jobs can have job clusters with the same key, therefore job id is prepended to ensure uniqueness
```
JCZuurmond, line 260: Note the differentiation between job clusters and "regular" clusters by prefixing the job id, similarly to tasks. The object_type is "CLUSTER" for both

src/databricks/labs/ucx/assessment/sequencing.py
JCZuurmond: Done, except cycle. Added todo for cycle to: https://github.com/databrickslabs/ucx/pull/3009. Cycle cannot be implemented here as notebook or files are not yet added to the sequence. See other PR

src/databricks/labs/ucx/assessment/sequencing.py
JCZuurmond: Thought about it a bit more, maybe we need to make this method protected access as well and introduce a `build_dependency_graph` instead. I am thinking that it makes sense to get this closer to the other linter code.

Open questions I have:
- Should this method expect a job or job_id. The later is similar to the cluster: get the job given the id, if not found return problem, otherwise parse the job like the code below.
- Should we distinguish between "resolve" and "register", where we do both in the same method now. If so, should we do an early return of the resolve problems, if any.
- Should the `job_migration_node` get the problems from the `tasks` and `clusters`? I think it makes more sense to isolate the problems to the node it **directly** applies too and surface all the problems when building the dependency graph.

src/databricks/labs/ucx/assessment/sequencing.py
```diff
@@ -0,0 +1,352 @@
...
+
+
+class MigrationSequencer:
+    """Sequence the migration dependencies in order to execute the migration.
+
+    Similar to the other graph logic, we first build the graph by registering dependencies, then we analyse the graph.
+    Analysing the graph in this case means: computing the migration sequence in `meth:generate_steps`.
+    """
+
+    def __init__(self, ws: WorkspaceClient, admin_locator: AdministratorLocator):
```
nfx, line 151:
```suggestion
    def __init__(self, ws: WorkspaceClient, admininstrator_locator: AdministratorLocator):
```
properties have to be named consistently - makes it easier to refactor.

src/databricks/labs/ucx/assessment/sequencing.py
nfx: `register_jobs()` is preferred.

> Should this method expect a job or job_id. The later is similar to the cluster: get the job given the id, if not found return problem, otherwise parse the job like the code below.

`jobs.Job` - we get tasks from it.

> Should we distinguish between "resolve" and "register", where we do both in the same method now. If so, should we do an early return of the resolve problems, if any.

imho, we need just `register()`

> Should the job_migration_node get the problems from the tasks and clusters? I think it makes more sense to isolate the problems to the node it directly applies too and surface all the problems when building the dependency graph.

tasks and job cluster problems have to rollup to a job. less objects for users to think of. though, if you think that having separate nodes for tasks in the graph is easier to implement - go ahead with `("TASK", f"{job.job_id}/{task.task_key}")`

tests/integration/assessment/test_sequencing.py
```diff
@@ -0,0 +1,51 @@
...
+
+def test_migration_sequencing_job_with_task_referencing_non_existing_cluster(runtime_ctx) -> None:
+    """Sequence a job with a task referencing existing cluster"""
+    # Cannot make an actual job referencing a non-existing cluster
+    task = jobs.Task(task_key="test-task", existing_cluster_id="non-existing-id")
+    settings = jobs.JobSettings(name="test-job", tasks=[task])
+    job = jobs.Job(job_id=1234, settings=settings)
+
+    maybe_job_node = runtime_ctx.migration_sequencer.register_job(job)
+    assert maybe_job_node.failed
```
nfx, line 47: what is the failure you're expecting here? assert explicitly.

tests/integration/assessment/test_sequencing.py
nfx: non-existing cluster should not be added as a step and there should be a problem added to job/task level that the cluster is removed.

JCZuurmond (approved):
https://github.com/databrickslabs/ucx/pull/3008#discussion_r1824449214

- `register_jobs` -> also reflected in the return type `list[MaybeMigrationNode
- kept the tasks as separate nodes

nfx (approved):
lgtm

PR #3016, opened 2024-10-18
Add dashboard for tracking migration progress

### LINKED ISSUE #2596: [FEATURE] Visualize migration process in dashboard
### Summary

Visualize `migration-process` results in dashboard, similarly to the other dashboards:

- Big widget with % work remaining
- Table with:
   - Owner
   - Number of objects
   - Work remaining in number and %

JCZuurmond:
## Changes
Add dashboard for tracking migration process, screenshot below as comments.

Open changes:
- [x] Update install to replace UCX catalog in queries, call the value to be replace `multiworkspaces`
- [x] Add counter for tables and views migrated
- [x] Use object serializer from #2743
- [x] Filter for latest scan only
- [x] Add overview of the (unique) failure messages
- [x] Reword "readiness"

### Linked issues

Resolves #2596
References https://github.com/databrickslabs/lsql/issues/306
Requires: #3083 

### Functionality

- [x] added a new dashboard: migration-progress

### Tests

- [x] manually tested
- [x] added integration tests


JCZuurmond:
See screenshots:

Screenshot 2024-10-18 at 14 25 31
Screenshot 2024-10-18 at 14 25 38

src/databricks/labs/ucx/progress/install.py
```diff
@@ -14,7 +14,7 @@
 logger = logging.getLogger(__name__)
 
 
-@dataclass(frozen=True, kw_only=True)
```
JCZuurmond, line 17: @asnare : Does this matter to you?

asnare: It doesn't matter in the technical sense.

I think it mattered more when we had lots of adjacent fields with the same types: keyword arguments allow verification of calls by trivial inspection that the correct value is being passed as the correct argument.

src/databricks/labs/ucx/queries/progress/02_1_migration_status_by_owner_bar_graph.sql
```diff
@@ -0,0 +1,22 @@
...
+    "encodings": {
+        "x":{"fieldName": "owner", "scale": {"type": "categorical"}, "displayName": "owner"},
+        "y":{"fieldName": "count", "scale": {"type": "quantitative"}, "displayName": "count"}
+    }
+}}'
+*/
+WITH migration_statuses AS (
+    SELECT *
+    FROM inventory.historical
+    WHERE object_type = "migration_status"
```
asnare, line 15: This needs to:
a) Join against the `workflow_runs` table;
b) Only select rows that belong to the latest run for each workspace.

src/databricks/labs/ucx/queries/progress/02_1_migration_status_by_owner_bar_graph.sql
```diff
@@ -0,0 +1,22 @@
...
+)
+
+SELECT
+    owner,
+    COUNT_IF(SIZE(failures) > 0) AS count
+FROM migration_statuses
+GROUP BY owner
```
asnare, line 22: Do we want to show owners that have a zero count?

If we don't, this can be more efficient by `WHERE size(failures) > 0` and then just using `count(1) as count`.

src/databricks/labs/ucx/queries/progress/02_2_migration_status_by_owner_overview.sql
```diff
@@ -0,0 +1,15 @@
+/* --title 'Tables and views migration completion (%)' --description 'Per owner' */
+WITH migration_statuses AS (
+    SELECT *
+    FROM inventory.historical
+    WHERE object_type = "migration_status"
```
asnare, line 5: Need to limit to just the latest snapshot for each workspace.

JCZuurmond:
Waits on https://github.com/databrickslabs/ucx/pull/2743/ for its serializer of objects

src/databricks/labs/ucx/queries/progress/01_0_percentage_migration_readiness.sql
```diff
@@ -0,0 +1,5 @@
+/* --title 'Overall readiness (%)' --width 2 */
+SELECT
+    100 * COUNT_IF(size(failures) = 0) / COUNT(*) AS percentage
+FROM multiworkspace.historical
+WHERE object_type != 'migration_status'
```
asnare, line 5: I'm a bit on the fence on this one; part of me wonders whether we should list the known object-types (to deal with another version introducing something like `migration_status` that should be excluded).

Linting issues are also coming shortly: these are per-problem records instead of per-resource.

JCZuurmond:
> The intent of the current widgets is to show the current migration status only; as such all the queries where we select from `multiworkspace.historical` need to join against `multiworkspace.workflow_runs` and only select rows from the most-recent run in each workspace.
> 
> It might make sense to define a view that does this in one place.

Did this, could you review?

src/databricks/labs/ucx/queries/progress/01_0_percentage_migration_readiness.sql
JCZuurmond: I made the check inclusive instead of exclusive

src/databricks/labs/ucx/queries/views/latest_historical_per_workspace.sql
```diff
@@ -0,0 +1,17 @@
+WITH last_workflow_run AS (
+    SELECT
+      workspace_id,
+      MAX(STRUCT(finished_at, workflow_run_attempt, started_at, workflow_run_id)) AS max_struct
+    FROM $inventory.workflow_runs  -- $inventory is a hardcoded name for replacing target schema in a view definition
+    WHERE workflow_name = 'migration-progress-experimental'
+    GROUP BY workspace_id
+)
+
+SELECT historical.*
```
asnare, line 10: It's best to explicitly list the columns instead of using `.*` in this sort of context in production queries. It's tedious, I know, but it makes the queries more forward/backward compatible with respect to schema changes: otherwise new columns break aggregations, changed column order breaks unions, etc. Plus it catches dropped columns early.)

src/databricks/labs/ucx/queries/views/latest_historical_per_workspace.sql
```diff
@@ -0,0 +1,17 @@
+WITH last_workflow_run AS (
+    SELECT
+      workspace_id,
+      MAX(STRUCT(finished_at, workflow_run_attempt, started_at, workflow_run_id)) AS max_struct
+    FROM $inventory.workflow_runs  -- $inventory is a hardcoded name for replacing target schema in a view definition
+    WHERE workflow_name = 'migration-progress-experimental'
+    GROUP BY workspace_id
```
asnare, line 7: This will locate the latest workflow run for each workspace, but it won't return the _set of latest runs_ for each workspace.

Workflow runs will not only be recorded by the migration-progress job, but also `migrate-tables` (which updates some of the inventory). As things stand, when `migrate-tables` does a run that only updates the grants and migration-status inventory, this query will only return _that_ run-id and downstream queries will therefore ignore all the other record-types that weren't refreshed by that job.

Circling back, for each workspace it needs to return the latest run id for each job in the workspace.

asnare (changes requested):

JCZuurmond:
https://github.com/user-attachments/assets/875a2b9f-e936-4e54-8299-38de0d9f7672

src/databricks/labs/ucx/queries/views/latest_historical_per_workspace.sql
JCZuurmond: The where covers this right?     WHERE workflow_name = 'migration-progress-experimental'

asnare: Ah, you're right… of-course that will be fine.

That said I thought we want to query the latest known data whereas here we will be ignoring some to an extent.

JCZuurmond: yeah, for now we only update the UCX catalog tables in the migration progress. If we change that, we can update the query at that time.

src/databricks/labs/ucx/queries/views/latest_historical_per_workspace.sql
JCZuurmond: It's a view being used by dashboards for which specific rows are filtered using a join, I think `.*` is fine here

src/databricks/labs/ucx/queries/progress/main/01_1_percentage_udf_migration_readiness.sql
```diff
@@ -0,0 +1,5 @@
+/* --title 'UDF migration readiness (%)' */
+SELECT
+    100 * COUNT_IF(size(failures) = 0) / COUNT(*) AS percentage
```
nfx, line 3:
```suggestion
    100 * COUNT_IF(SIZE(failures) = 0) / COUNT(*) AS percentage
```

nit: `SIZE()` is a function, hence should be uppercase

src/databricks/labs/ucx/queries/progress/main/01_0_migration_readiness.md
```diff
@@ -0,0 +1,5 @@
+# 1. Migration Readiness
```
nfx, line 1:
```suggestion
# 1. Migration Progress
```
shouldn't we re-word this with "Progress" instead? "100% progress" here meaning "no more work necessary"

src/databricks/labs/ucx/queries/progress/main/02_4_migration_status_by_owner_overview.sql
```diff
@@ -0,0 +1,15 @@
+/* --title 'Overview' --description 'Tables and views migration' --width 5 */
+WITH migration_statuses AS (
+    SELECT *
+    FROM ucx_catalog.multiworkspace.latest_historical_per_workspace
+    WHERE object_type = "migration_status"
+)
+
+SELECT
+    owner,
+    100 * COUNT_IF(SIZE(failures) = 0) / SUM(COUNT(*)) OVER (partition by owner) AS percentage,
```
nfx, line 10:
```suggestion
    100 * COUNT_IF(SIZE(failures) = 0) / SUM(COUNT(*)) OVER (PARTITION BY owner) AS percentage,
```
nit: `PARTITION BY` are keywords

src/databricks/labs/ucx/queries/views/latest_historical_per_workspace.sql
```diff
@@ -0,0 +1,25 @@
...
+    historical.workspace_id,
+    historical.job_run_id,
+    historical.object_type,
+    historical.object_id,
+    historical.data,
+    historical.failures,
+    historical.owner,
+    historical.ucx_version
+FROM
+    $inventory.historical AS historical  -- $inventory is a hardcoded name for replacing target schema in a view definition
```
nfx, line 20: do we have an issue for fixing this?

src/databricks/labs/ucx/progress/install.py
```diff
@@ -47,13 +47,17 @@ class ProgressTrackingInstallation:
...
+        from databricks.labs import ucx  # pylint: disable=import-outside-toplevel
+
+        self._schema_deployer = SchemaDeployer(sql_backend, self._SCHEMA, mod=ucx, catalog=ucx_catalog)
 
     def run(self) -> None:
         self._schema_deployer.deploy_schema()
         self._schema_deployer.deploy_table("workflow_runs", WorkflowRun)
         self._schema_deployer.deploy_table("historical", Historical)
+        self._schema_deployer.deploy_view(
+            "latest_historical_per_workspace", "queries/views/latest_historical_per_workspace.sql"
```
nfx, line 59:
```suggestion
            "objects", "queries/views/latest_historical_per_workspace.sql"
```
could we rename this view to `objects` or `objects_snapshot`. shorter name would be better here.

tests/integration/queries/test_migration_progress.py
```diff
@@ -0,0 +1,353 @@
...
+
+
+@pytest.fixture
+def tables():
+    records = []
+    for table in (
+        Table("hive_metastore", schema, table, "MANAGED", "delta")
+        for schema in ("schema1", "schema2")
+        for table in ("table1", "table2", "table3", "table4", "table5")
+    ):
```
nfx, line 59: i wonder why this does not trigger https://github.com/databrickslabs/pylint-plugin?tab=readme-ov-file#r8923-rewrite-as-for-loop

tests/integration/queries/test_migration_progress.py
```diff
@@ -0,0 +1,353 @@
...
+        records.append(record)
+    return records
+
+
+@pytest.fixture
+def jobs():
+    records = []
+    for job in (
+        JobInfo("1", success=1, failures=""),
+        JobInfo("2", success=0, failures="No isolation shared clusters not supported in UC"),
```
nfx, line 148:
```suggestion
        JobInfo("2", success=0, failures=`["No isolation shared clusters not supported in UC"]`),
```
`failures` is a JSON-serialized array of strings

tests/integration/queries/test_migration_progress.py
```diff
@@ -0,0 +1,353 @@
...
+    installation_ctx.sql_backend.save_table(
+        f"{ucx_catalog}.multiworkspace.workflow_runs", workflow_runs, WorkflowRun, mode="overwrite"
+    )
+    workspace_id = installation_ctx.workspace_client.get_workspace_id()
+    historicals = []
+    for job_run_id in range(1, 3):  # No changes between migration progress run
+        for table_name, id_, instance, failures, owner in historical_objects:
+            # TODO: Use historical encoder from https://github.com/databrickslabs/ucx/pull/2743/
+            data = {
+                field.name: str(getattr(instance, field.name))
```
nfx, line 258: https://github.com/databrickslabs/pylint-plugin?tab=readme-ov-file#r8923-rewrite-as-for-loop

tests/integration/queries/test_migration_progress.py
```diff
@@ -0,0 +1,353 @@
...
+) -> str:
+    """Populate the UCX catalog given the objects from the fixtures.
+
+    For optimization purposes, this fixture could be "module" (or "session") scoped. However, dependant fixtures are
+    "function" scoped, thus one should first evaluate if those can be changed.
+    """
+    installation_ctx.progress_tracking_installation.run()
+    ucx_catalog = installation_ctx.ucx_catalog
+    installation_ctx.sql_backend.save_table(
+        f"{ucx_catalog}.multiworkspace.workflow_runs", workflow_runs, WorkflowRun, mode="overwrite"
```
nfx, line 250:
```suggestion
        f"{ucx_catalog}.multiworkspace.workflow_runs", workflow_runs, WorkflowRun, mode="overwrite",
```

nit: `make fmt`

src/databricks/labs/ucx/queries/progress/main/01_4_percentage_cluster_migration_readiness.sql
```diff
@@ -0,0 +1,5 @@
+/* --title 'Cluster migration readiness (%)' */
+SELECT
+    100 * COUNT_IF(size(failures) = 0) / COUNT(*) AS percentage
```
nfx, line 3:
```suggestion
    ROUND(100 * COUNT_IF(size(failures) = 0) / COUNT(*), 2) AS percentage
```

`73.91304347826087` -> `73.91`. probably `CEIL()` would be even better here.

nfx (changes requested):
minor comments about view renames

tests/integration/queries/test_migration_progress.py
JCZuurmond: Kept it to replace it with the encoder from Andrew's PR

tests/integration/queries/test_migration_progress.py
```diff
@@ -0,0 +1,353 @@
...
+        table_migration_status = TableMigrationStatus(table.catalog, table.database, table.name)
+        failures = ["not migrated"]
+        owner = "Andrew" if table.name == "table1" else "Cor"
+        if table.database == "schema1":  # Simulate one schema being migrated
+            if table.name == "table2":  # An owner with only migrated objects
+                owner = "Eric"
+            table_migration_status.dst_catalog = "catalog1"
+            table_migration_status.dst_schema = table.database
+            table_migration_status.dst_table = table.name
+            failures = []
```
JCZuurmond, line 78: @asnare and @nfx : I am incorporating the historical encoder from https://github.com/databrickslabs/ucx/pull/2743, which brings up the question on what the failures are for `TableMigrationStatus`. 

I understood that the intend of the failures is to say:
- if there are failures, then the object is *not* migrated
- The object is migrated otherwise

This means I need to add additional logic in the `Encoder` to do this for `TableMigrationStatus` similar to what I do here: if `dst_` values exists, then no failures

nfx: failures for migration state should be "table x.y is not yet migrated to a.b.c"

JCZuurmond:
Moving this back to draft to resolve encoding `failures` first, then see if the dashboard shows what we need it to

src/databricks/labs/ucx/queries/views/latest_historical_per_workspace.sql
JCZuurmond: https://github.com/databrickslabs/lsql/issues/311

tests/integration/queries/test_migration_progress.py
JCZuurmond: covered by https://github.com/databrickslabs/ucx/pull/3083

JCZuurmond:
Waiting for https://github.com/databrickslabs/ucx/pull/3083, then will make a new video of the dashboard

JCZuurmond:
Uploading Screen Recording 2024-10-28 at 18.13.52.mov…

JCZuurmond:
No functional changes, some fixes only:


https://github.com/user-attachments/assets/e908df4f-75a0-4924-80d9-fed6790ce874

nfx (approved):
lgtm

PR #3027, opened 2024-10-21
Preemptively fix unknown errors of Python AST parsing coming from `astroid` and `ast` libraries

nfx:


src/databricks/labs/ucx/source_code/python/python_ast.py
```diff
@@ -77,9 +76,7 @@ def maybe_parse(cls, code: str) -> MaybeTree:
             root = parse(code)
             tree = Tree(root)
             return MaybeTree(tree, None)
-        except AstroidSyntaxError as e:
-            return cls._definitely_failure('syntax-error', code, e)
-        except SystemError as e:
+        except Exception as e:  # pylint: disable=broad-exception-caught
```
JCZuurmond, line 79: @nfx : Would have like to see a comment that motivates why we have a broad exception

PR #3034, opened 2024-10-21
Ensure that pipeline assessment doesn't fail if a pipeline is deleted…

asnare:
## Changes

This PR updates the pipelines (DLT) crawler so that crawling doesn't fail if a pipeline is deleted while we are iterating over them all. Instead of failing a warning is now logged and the crawling continues.

### Functionality

- modified existing workflows: `assessment`, `migration-progress-experimental`

### Tests

- added unit tests
- existing integration tests


nfx (approved):
lgtm

PR #3039, opened 2024-10-22
Migration progress: include DFSA records in the history log

asnare:
## Changes

This PR follows on from #2743 by extending the set of updates that we capture to include updated `DirectFsAccess` snapshots for dashboards and jobs.

### Linked issues

Follows #2743.

### Functionality

### Tests

- added unit tests
- existing integration tests


asnare:
> ❌ test_running_real_migration_progress_job: AssertionError: Workflow failed: migration-progress-experimental (25m43.878s)

This is currently failing due to a bug in the crawlers that means the snapshots cannot be loaded when the Spark-based runtime is being used; fixed in #3046.

nfx (approved):
lgtm

asnare:
Following a discussion, we've decided not to include DFSA records in their current form in the history table. Each DFSA record corresponds to a problem with another resource (eg. notebook, jobs). As such the intent is to aggregate these records and include them in the list of failures on the resource-specific record.

PR #3056, opened 2024-10-24
Fix import member issue

ericvergnaud:
## Changes
Our current implementation does not properly collect imports when source code uses indirect import patterns such as:
```
from importlib import import_module
module = import_module("some-module")

```
This PR fixes that

### Linked issues
None

### Functionality
None

### Tests
- [x] added unit tests


nfx (approved):
lgtm

PR #3068, opened 2024-10-24
Tech debt: let `HistoryEncoding` work with string-based type annotations in addition to the normal type-based ones

asnare:
## Changes

This PR cherry-picks some changes from #3039 that updated the `HistoryEncoder` to work correctly with databases that are declared with `__future__.__annotations__` in effect. 

When this annotation is in effect, python converts all type-hints during import/declaration into strings and then performs deferred resolution at a later stage. (This is why forward references work.) Unfortunately the dataclass mechanism captures field types prior to deferred resolution. This PR ensures that our type checking works anyway.

### Linked issues

Cherry-picks from #3039.

### Tests

- updated unit tests

### DIFF
--- a/src/databricks/labs/ucx/progress/history.py
+++ b/src/databricks/labs/ucx/progress/history.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 import dataclasses
 import datetime as dt
+import typing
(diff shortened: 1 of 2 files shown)


nfx (approved):
lgtm

PR #3072, opened 2024-10-24
Implementation an Ownership factory

ericvergnaud:
## Changes
Every asset has an owner and our implementation requires that the owner be retrieved using a specialized `Ownership` instance. 
There is also a requirement that such `Ownership` be retrieved from the `GlobalContext`, which in complex scenarios may result in a proliferation of constructor parameters.
This PR implements an `Ownership` factory that makes it easy to retrieve the appropriate `xxxOwnership` from the `GlobalContext`, simply using the `Record` type being considered.

### Linked issues
Progresses #1415

### Functionality
None

### Tests
- [x] added unit tests

### DIFF
--- a/src/databricks/labs/ucx/contexts/application.py
+++ b/src/databricks/labs/ucx/contexts/application.py
@@ -28,7 +28,11 @@ from databricks.labs.ucx.assessment.azure import AzureServicePrincipalCrawler
 from databricks.labs.ucx.assessment.export import AssessmentExporter
 from databricks.labs.ucx.aws.credentials import CredentialManager
 from databricks.labs.ucx.config import WorkspaceConfig
-from databricks.labs.ucx.framework.owners import AdministratorLocator, WorkspacePathOwnership
+from databricks.labs.ucx.framework.owners import (
+    AdministratorLocator,
+    WorkspacePathOwnership,
+    Ownership,
+)
 from databricks.labs.ucx.hive_metastore import ExternalLocations, MountsCrawler, TablesCrawler
 from databricks.labs.ucx.hive_metastore.catalog_schema import CatalogSchema
 from databricks.labs.ucx.hive_metastore.grants import (
@@ -558,6 +562,19 @@ class GlobalContext(abc.ABC):
     @cached_property
     def administrator_locator(self) -> AdministratorLocator:
         return AdministratorLocator(self.workspace_client)
+
+    @cached_property
+    def ownership_factory(self) -> Callable[[type], Ownership]:
+        # ensure registration of Ownerships
+        names_with_ownership = [name for name in dir(GlobalContext) if "ownership" in name]
+        for name in names_with_ownership:
+            if name == "ownership_factory":
+                continue
+            prop = getattr(GlobalContext, name)
+            if not isinstance(prop, cached_property):
+                continue
+            _ = getattr(self, name)
+        return Ownership.for_record_type
 
 
 class CliContext(GlobalContext, abc.ABC):
--- a/src/databricks/labs/ucx/framework/owners.py
+++ b/src/databricks/labs/ucx/framework/owners.py
@@ -1,9 +1,10 @@
+from __future__ import annotations
 import logging
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable, Sequence
 from datetime import timedelta
 from functools import cached_property
-from typing import Generic, TypeVar, final
+from typing import Generic, TypeVar, final, cast
 
 from databricks.labs.blueprint.paths import WorkspacePath
 from databricks.sdk import WorkspaceClient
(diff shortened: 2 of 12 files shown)


src/databricks/labs/ucx/framework/owners.py
```diff
@@ -169,8 +171,15 @@ def get_workspace_administrator(self) -> str:
...
     """Determine an owner for a given type of object."""
 
-    def __init__(self, administrator_locator: AdministratorLocator) -> None:
+    _factories: dict[type, Ownership] = {}
+
+    @classmethod
+    def for_record_type(cls, record_type: type) -> Ownership[Record]:
+        return cast(Ownership[Record], cls._factories[record_type])
+
+    def __init__(self, administrator_locator: AdministratorLocator, record_type: type) -> None:
```
nfx, line 180: this control flow confuses `pylint` and it will fail to detect hidden future bugs.wouldn't it be more clear if we add `is_applicable` abstract method

```
def is_applicable(self, record: Any) -> bool:
  return isinstance(record, Table)
```

this will:

1. avoid changing every constructor at the cost of adding new method
2. allow for one ownership to work with more than one record class
3. allow for an explicit injectable facade:

```python
class AnyOwnership(Ownership[Any]):
  def __init__(self, ownerships: list[Ownership]): ...

  def owner_of(self, record: Any) -> str:
    for o for self._ownerships: if o.is_applicable(record): return o.owner_of(record)
    return self._administrator_locator....
```

src/databricks/labs/ucx/contexts/application.py
```diff
@@ -571,6 +576,19 @@ def migration_recon(self) -> MigrationRecon:
     def administrator_locator(self) -> AdministratorLocator:
         return AdministratorLocator(self.workspace_client)
 
+    @cached_property
+    def ownership_factory(self) -> Callable[[type], Ownership]:
+        # ensure registration of Ownerships
+        names_with_ownership = [name for name in dir(GlobalContext) if "ownership" in name]
```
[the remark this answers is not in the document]
ericvergnaud: done

src/databricks/labs/ucx/framework/owners.py
ericvergnaud: done

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -169,8 +171,21 @@ def get_workspace_administrator(self) -> str:
 class Ownership(ABC, Generic[Record]):
     """Determine an owner for a given type of object."""
 
+    _ownerships: set[Ownership] = set()
```
asnare, line 174: I don't think the type annotation here is quite correct: it needs to be `ClassVar[…]`? (Without this I believe it is considered by the type tooling to be an instance property.)

src/databricks/labs/ucx/contexts/application.py
```diff
@@ -571,6 +577,20 @@ def migration_recon(self) -> MigrationRecon:
...
+        _ = [
+            self.directfs_access_ownership,
+            self.grant_ownership,
+            self.legacy_query_ownership,
+            self.table_migration_ownership,
+            self.table_ownership,
+            self.udf_ownership,
+            self.workspace_path_ownership,
+        ]
+        return Ownership.for_record
```
asnare, line 592: As things stand during tests we re-create the application context for each test that needs one, and this is going to lead to the registry accumulating instances for each test that hits this property.

For this reason (and others) in the past we've preferred to avoid global singletons (which the class-hosted registry is), and move the cache to be owned and managed outside of it so that the lifecycle of the cache is tied to the application context.

asnare (commented):
I've added some comments.

One thing I wonder: if there a context where the ownership instance that we need is not static in nature and tied to the type?

In not, I think we could replace things with a dictionary of `type: ownership_instance` in the application context. (Right now we wouldn't even need to deal with specialisation.)

I don't know about the context or any discussions that lead to this PR, so this is just based on the context provided here alone. (To be clear, I do understand the desire to avoid the proliferation of properties we have on the application context.)

src/databricks/labs/ucx/framework/owners.py
```diff
@@ -169,8 +171,21 @@ def get_workspace_administrator(self) -> str:
...
+            if ownership.is_applicable_to(record):
+                return ownership
+        raise ValueError(f"Ownership not implemented or not registered for {type(record).__name__}")
+
     def __init__(self, administrator_locator: AdministratorLocator) -> None:
         self._administrator_locator = administrator_locator
+        self._ownerships.add(self)
+
+    @abstractmethod
+    def is_applicable_to(self, record: Any) -> bool: ...
```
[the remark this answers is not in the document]
asnare: Sorry, I missed [this comment](https://github.com/databrickslabs/ucx/pull/3072#discussion_r1824480883) indicating that we want to allow instances to support multiple types.

That said, I don't fully follow the reasoning: type-based lookups can also support multiple instance types (`A | B`) and encourage a single lookup rather than the (current) situation that encourages callers to implicitly linear probe the registry within a loop. I think I'm missing something here, and hope that @nfx can help me understand what he has in mind.

PR #3075, opened 2024-10-24
Use table ownership heuristics for migrating ACLs

nfx:
This PR prepends `OWN` grants when migrating ACLs for tables. For detailed logic on ownership heuristics, see #3066


nfx:
fails integration tests

nfx:
need local scheduler to improve on debuggability

```
import logging

from databricks.labs.ucx.contexts.workflow_task import RuntimeContext
from databricks.labs.ucx.runtime import Workflows

class InProcessDeployedWorkflows:
    def __init__(self, ctx: RuntimeContext):
        self._workflows = {workflow.name: workflow for workflow in Workflows.definitions()}
        self._ctx = ctx

    def run_workflow(self, step: str, **_):
        workflow = self._workflows[step]
        incoming = {task.name: 0 for task in workflow.tasks()}
        queue = []
        for task in workflow.tasks():
            task.workflow = workflow.name
            incoming[task.name] += len(task.depends_on)
        for task in workflow.tasks():
            if incoming[task.name] == 0:
                queue.append(task)
        while queue:
            task = queue.pop(0)
            fn = getattr(workflow, task.name)
            fn(self._ctx)
            for dep in task.depends_on:
                incoming[dep] -= 1
                if incoming[dep] == 0:
                    queue.append(dep)

    def relay_logs(self, step: str):
        pass  # noop

```

src/databricks/labs/ucx/contexts/application.py
```diff
@@ -316,9 +320,15 @@ def acl_migrator(self):
             self.config.inventory_database,
         )
 
+    @cached_property
+    def table_ownership_grant_loader(self) -> TableOwnershipGrantLoader:
+        return TableOwnershipGrantLoader(self.tables_crawler, self.table_ownership)
+
     @cached_property
     def migrate_grants(self) -> MigrateGrants:
+        # owner grants have to come first
```
JCZuurmond, line 329: Why?

src/databricks/labs/ucx/contexts/application.py
```diff
@@ -316,9 +320,15 @@ def acl_migrator(self):
...
         grant_loaders: list[Callable[[], Iterable[Grant]]] = [
+            self.table_ownership_grant_loader.load,
             self.grants_crawler.snapshot,
```
JCZuurmond, line 332: Should the OWN grants in this crawler be ignored?

JCZuurmond (changes requested):

src/databricks/labs/ucx/contexts/application.py
nfx: because of `match_grants` returning the first matched grant

JCZuurmond: Please add that to the comment

PR #3080, opened 2024-10-25
Use LegacyQueryPath dataclass as record for LegacyQueryOwnership

ericvergnaud:
## Changes
`LegacyQueryOwnership` takes a `str` as record type which obfuscates the meaning of that str and creates risk of collision in the ownership factory
This PR fixes that by introducing a `LegacyQueryPath` data class as a wrapper around the `str`

### Linked issues
None

### Functionality
None

### Tests
- [x] ran unit tests


nfx (approved):
lgtm

PR #3087, opened 2024-10-25
Ignore failed inference codes when presenting results to Databricks Runtime

nfx:
This PR disables the following advice codes from being propagated in our notebook editor:

- 'cannot-autofix-table-reference',
- 'default-format-changed-in-dbr8',
- 'dependency-not-found',
- 'not-supported',
- 'notebook-run-cannot-compute-value',
- 'sql-parse-error',
- 'sys-path-cannot-compute-value',
- 'unsupported-magic-line',

### DIFF
--- a/src/databricks/labs/ucx/source_code/lsp_plugin.py
+++ b/src/databricks/labs/ucx/source_code/lsp_plugin.py
@@ -15,6 +15,18 @@ from databricks.labs.ucx.source_code.lsp import Diagnostic
 logger = logging.getLogger(__name__)
 
 
+DEBUG_MESSAGE_CODES = {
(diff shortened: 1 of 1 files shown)


src/databricks/labs/ucx/source_code/lsp_plugin.py
```diff
@@ -15,6 +15,18 @@
 logger = logging.getLogger(__name__)
 
 
+DEBUG_MESSAGE_CODES = {
```
JCZuurmond, line 18: @nfx : Please add a comment about why we ignore this

JCZuurmond (changes requested):
Changes lgtm, requested clarification

PR #3200, opened 2024-11-04
Speed up `update_migration_status` jobs by eliminating lots of redundant SQL queries

### LINKED ISSUE #2397: [TECH DEBT] Update the signature of `_is_migrated` function to only get table names as input
On one hand, I would like this method to have the following signature: `_is_migrated(self, table: Table) -> bool`. To me that makes more sense, it is "better" OO programming and it is probably more future proof (what if we want to include the catalog at some point).

On the other hand, I do not think you have to change it because: i) the `index.is_migrated` has the same signature, so it is better to be consitent now, log this as a (small) tech debt issue and do refactoring in a separate PR. (We prefer to separate refactoring from feature implementation.) and ii) it is a hidden method, therefore we can refactor the signature without worrying about changing the public API

_Originally posted by @JCZuurmond in https://github.com/databrickslabs/ucx/pull/2205#discussion_r1706541771_

### LINKED ISSUE #2730: [TECH DEBT]: Inefficient use of `TableMigrationStatusRefresher.index()`
### Problem statement

The `TableMigrationStatusRefresher.index()` method is used to load the migration state of all tables and views in the inventory. The fast-path is that this loads the migration state from the inventory table (`select * from migration_status`).

We have a lot of code-paths that invoke this function to obtain the index, especially in the migration-related code, not realising that it's relatively expensive. One particular case is: `TablesMigrator.get_remaining_tables()` which loads this index (and discards it) within a loop over every table in the inventory.

### Proposed Solution

At the very least `TablesMigrator.get_remaining_tables()` should be refactored to only build the index once.

I haven't some a full survey of usage, but there are probably many other places where refactoring is needed to avoid redundant loading of the index.

asnare:
## Changes

This PR updates the migration workflows where they run an `update_migration_status` task at the end. This task checks every table in the inventory, and previously it would reload (via SQL) the index of migrated tables for every table that it checked. This PR eliminates this redundant reloading by using the refreshed index (loaded once) for all the checking.

Some incidental dead code is also removed.

### Linked issues

Resolves #2730.
Resolves #2397.

### Functionality

   - `migrate-tables`
   - `migrate-external-hiveserde-tables-in-place-experimental`
   - `migrate-external-tables-ctas`
   - `scan-tables-in-mounts-experimental`
   - `migrate-tables-in-mounts-experimental`

### Tests

- existing unit tests
- existing integration tests


nfx (approved):
lgtm

nfx (approved):
lgtm

PR #3212, opened 2024-11-05
Cater for empty python cells

ericvergnaud:
## Changes
Notebok cells such as the following will crash the dependency builder.
This PR fixes the issue.
```
# COMMAND ----------

#only run this step if you are running below Databricks runtime 13.2 GPU; if using 13.2 and above this is not needed.
%sh
wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/cuda-ubuntu2004.pin
sudo mv cuda-ubuntu2004.pin /etc/apt/preferences.d/cuda-repository-pin-600
sudo apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/3bf863cc.pub
sudo add-apt-repository -y "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/ /"
sudo apt-get update

```

### Linked issues
None

### Functionality
None

### Tests
- [x] ran solacc on failed repo

PR #3216, opened 2024-11-05
Improve solacc

ericvergnaud:
## Changes
Addresses various issues that affect `solacc` CI job

### Linked issues
includes #3212 
includes #3217 
includes #3218 

### Functionality
None

### Tests
- [x] manually ran solacc job

SHOULD BE EMPTY ONCE ABOVE PRs ARE MERGED


ericvergnaud:
The failing `solacc` CI job seems to be caused by some infrastructure timeout, where GH loses contact with the job. A possible solution is to split the work into chunks processed by dedicated threads. To be explored.

PR #3236, opened 2024-11-10
Proposal for dbfs file access api with mocked test suite

danzafar:
## Changes

| :exclamation:  This PR does not provide any new user-facing functionality nor refactor existing code, it is simply intended to be a sanity check on approach before the comprehensive refactor of `scan-tables-in-mounts`.   |
|--------------------------------------------------------------------------------------------|

### Why this PR?
Refactoring `scan-tables-in-mounts` to enhance performance will need to leverage parallelized crawling of `dbfs:/mnt/` locations. @nfx suggested using backend Hadoop libraries through `py4j` for this through the `SparkSession`. Since this is a fairly big change, I wanted to make sure we are aligned on the backend file lister before I proceed to modify any _existing_ files.

### Enable `dbfs` file listing
This PR proposes the `DbfsFiles` class which uses the `SparkSession`'s java backend with `py4j` to leverage the following Hadoop libraries for efficient `dbfs:/` (and `mount`) file access:

- `org.apache.hadoop.fs.FileSystem`
- `org.apache.hadoop.fs.Path`

As of now, the only useful method is `list_dir`, but if the approach is confirmed I plan to add crawling capabilities leveraging `databricks.labs.blueprint.parallel`.

### Test strategy
A testing strategy was painstakingly created to:
- mock the jvm functionality needed
- clean up the currently messy test suite pattern for existing `scan-tables-in-mounts`. I created a simple trie-based mock file system which can be leveraged for `scan-tables-in-mounts` functionality after the refactor. This mock file system also has it's own unit tests.

| :warning: WARNING           |
|:----------------------------|
| If you don't like my mock file system, it will probably hurt my feelings so please have a good reason!   |

### Functionality

- [ ] added new CLI command
- [ ] added a new workflow
- [ ] added a new table

### Tests

- [ ] manually tested
- [x] added unit tests
- [ ] added integration tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,8 @@ dependencies = ["databricks-sdk~=0.30",
                 "databricks-labs-blueprint>=0.9.1,<0.10",
                 "PyYAML>=6.0.0,<7.0.0",
                 "sqlglot>=25.5.0,<25.30",
-                "astroid>=3.3.1"]
+                "astroid>=3.3.1",
+                "py4j==0.10.9.7"]
 
 [project.optional-dependencies]
 pylsp = [
--- a/src/databricks/labs/ucx/hive_metastore/dbfs_files.py
+++ b/src/databricks/labs/ucx/hive_metastore/dbfs_files.py
@@ -1,0 +1,85 @@
+import logging
+from dataclasses import dataclass
+from functools import cached_property
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class DbfsFileInfo:
+    path: str
+    name: str
+    is_dir: bool
+    modification_time: int
+
+
+class DbfsFiles:
+
+    def __init__(self, jvm_interface=None):
+        # pylint: disable=import-error,import-outside-toplevel
+        if jvm_interface:
+            self._spark = jvm_interface
+        else:
+            try:
+                from pyspark.sql.session import SparkSession  # type: ignore[import-not-found]
+
+                self._spark = SparkSession.builder.getOrCreate()
+            except Exception as err:
+                logger.error(f"Unable to get SparkSession: {err}")
+                raise err
+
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -49,7 +49,8 @@ dependencies = ["databricks-sdk~=0.30",
                 "databricks-labs-blueprint>=0.9.1,<0.10",
                 "PyYAML>=6.0.0,<7.0.0",
                 "sqlglot>=25.5.0,<25.30",
-                "astroid>=3.3.1"]
+                "astroid>=3.3.1",
+                "py4j==0.10.9.7"]
```
nfx, line 53: we cannot add `py4j` as a dependency to UCX, as it'll have to be packaged with transitive dependencies to run CLI.

src/databricks/labs/ucx/hive_metastore/dbfs_files.py
```diff
@@ -0,0 +1,85 @@
...
+        # pylint: disable=import-outside-toplevel
+        from py4j.java_gateway import java_import  # type: ignore[import]
+
+        java_import(jvm, import_path)
+
+    @cached_property
+    def _jvm(self):
+        try:
+            _jvm = self._spark._jvm
+            self._java_import(_jvm, "org.apache.hadoop.fs.FileSystem")
```
nfx, line 49:
```suggestion
            FileSystem = jvm.org.apache.hadoop.fs.FileSystem
```
use the same technique as here, where you don't have to depend directly on py4j:
https://github.com/databrickslabs/ucx/blame/main/src/databricks/labs/ucx/hive_metastore/locations.py#L397-L401

nfx (changes requested):
don't directly depend on `py4j`, see [techniques](https://github.com/databrickslabs/ucx/blame/main/src/databricks/labs/ucx/hive_metastore/locations.py#L397-L401). Otherwise the approach is good.

PR #3239, opened 2024-11-11
Update table-migration workflows to also capture updated migration progress into the history log

asnare:
## Changes

The table-migration workflows already contained tasks at the end that log information about tables that still need to be migrated. The primary purpose of this PR is to update these workflows so they also capture updated progress information into the history log.

Other changes include:

 - Updating the documentation for which workflows update which tables.
 - ~Updating the (singleton) encoder for table-history so that initialisation doesn't trigger an implicit refresh of the `TableMigrationStatus` data. Instead this is controlled at the workflow level, as intended.~ Moved to #3270.

### Linked issues

~Conflicts with #3200 (will need rebasing).~ (Resolved.)

### Functionality

- updated documentation

  - `migrate-tables`
  - `migrate-external-hiveserde-tables-in-place-experimental`
  - `migrate-external-tables-ctas`
  - `scan-tables-in-mounts-experimental`
  - `migrate-tables-in-mounts-experimental`

### Tests

- manually tested
- updated and new unit tests
- updated and new integration tests


nfx (changes requested):

tests/integration/hive_metastore/test_workflows.py
```diff
@@ -27,39 +28,60 @@ def test_table_migration_job_refreshes_migration_status(
...
+        f"SELECT 1 FROM {installation_ctx.ucx_catalog}.multiworkspace.workflow_runs\n"
+        f"WHERE workspace_id = {installation_ctx.workspace_id}\n"
+        f"  AND workflow_run_id = {run_id}\n"
+        f"LIMIT 1\n"
+    )
+    assert any(installation_ctx.sql_backend.fetch(query)), f"No workflow run captured: {query}"
+
+    # Ensure that the history file has table records written to it that correspond to this run.
+    query = (
+        f"SELECT 1 from {installation_ctx.ucx_catalog}.multiworkspace.historical\n"
```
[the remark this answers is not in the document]
asnare: That's a good question; I've fixed this.

src/databricks/labs/ucx/hive_metastore/workflows.py
```diff
@@ -57,10 +57,50 @@ def migrate_views(self, ctx: RuntimeContext):
...
+    @job_task(
+        depends_on=[
+            convert_managed_table,
+            migrate_external_tables_sync,
+            migrate_dbfs_root_delta_tables,
+            migrate_dbfs_root_non_delta_tables,
+            migrate_views,
+            setup_tacl,
+        ],
+        job_cluster="tacl",
```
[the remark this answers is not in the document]
asnare: I'll try this.

tests/unit/hive_metastore/test_table_migrate.py
```diff
@@ -1375,63 +1375,70 @@ def test_revert_migrated_tables_failed(caplog, mock_pyspark):
...
-            object_type="EXTERNAL",
-            table_format="DELTA",
-            catalog="hive_metastore",
-            database="schema1",
-            name="table1",
-            location="s3://some_location/table1",
-            upgraded_to="ucx_default.db1_dst.dst_table1",
+    tables = (
+        (
+            Table(
```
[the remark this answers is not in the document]
asnare: The was part of factoring out the sequence of tables returned by the `snapshot()` mock (and make it immutable) so that the migration-index mock return the correct values. I'm not sure why the extra parentheses were in there, but I've removed them. Other than those it was only whitespace.

src/databricks/labs/ucx/progress/tables.py
```diff
@@ -38,17 +42,23 @@ def __init__(
...
+        self._migration_status_refresher = migration_status_refresher
 
-    def _encode_record_as_historical(self, record: Table) -> Historical:
+    SC: ClassVar = TableMigrationIndex
+
+    @contextmanager
+    def _snapshot_context(self) -> Generator[SC, None, None]:
+        yield TableMigrationIndex(self._migration_status_refresher.snapshot())
+
+    def _encode_record_as_historical(self, record: Table, snapshot_context: SC) -> Historical:
```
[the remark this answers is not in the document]
nfx: > Prior to these changes, this was being injected via the constructor and could lead to mismatches because it needs to correspond to the Table snapshot that is being encoded. 

why `force_refresh=True` is not enough there?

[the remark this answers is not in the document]
nfx: > Drop `@final` on the base-class method, and simply specialise it in TableProgressEncoder to handle both loading the index and supplying it as a parameter for encoding.

this is the preferred way, as it the least amount of change.

> Update the signature of ProgressEncoder so that the caller is responsible for providing both the Table snapshot to encode, as well as the accompanying TableMigrationIndex that it should use for enrichment. This is also a public API change. However I don't really see it as an improvement, and it makes some things internally more complicated to deal with.

it doesn't seem to be the least amount of change. Also, other encoders don't need tables.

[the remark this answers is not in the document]
asnare: Okay, I've ditched the context and switched to specialisation just within `TableProgressEncoder`.

Regarding `force_refresh=True`, the issue is not so much about whether it's enough: it's about where it occurs and which component is responsible for it. My understanding based on [this](https://github.com/databrickslabs/ucx/pull/2743#discussion_r1783326822) is that workflows are where force-refreshing is supposed to happen explicitly, and not implicitly within other components. The changes here are aimed at moving closer to that.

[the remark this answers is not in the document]
asnare: I've extracted this part of the PR to #3270 to: a) reduce the size and scope of this PR; b) it's clearer what's going on.

asnare:
Most of the integration test failures are due to #3272 (fixed in #3273).

asnare:
Integration tests currently failing due to a dependency issue on `main`; resolved by #3312.

nfx (approved):
lgtm

src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -56,14 +60,11 @@ def __init__(
         self._migrate_grants = migrate_grants
         self._external_locations = external_locations
 
-    def get_remaining_tables(self) -> list[Table]:
-        migration_index = self.index(force_refresh=True)
-        table_rows = []
+    def check_remaining_tables(self, migration_status: Iterable[TableMigrationStatus]) -> None:
```
JCZuurmond, line 63: Plural for an iterable

```suggestion
    def check_remaining_tables(self, migration_statuses: Iterable[TableMigrationStatus]) -> None:
```

src/databricks/labs/ucx/hive_metastore/workflows.py
```diff
@@ -57,10 +59,50 @@ def migrate_views(self, ctx: RuntimeContext):
         """
         ctx.tables_migrator.migrate_tables(what=What.VIEW)
 
-    @job_task(job_cluster="user_isolation", depends_on=[migrate_views])
-    def update_migration_status(self, ctx: RuntimeContext):
-        """Refresh the migration status to present it in the dashboard."""
-        ctx.tables_migrator.get_remaining_tables()
+    @job_task(job_cluster="user_isolation")
+    def verify_prerequisites(self, ctx: RuntimeContext) -> None:
```
JCZuurmond, line 63:
```suggestion
    def verify_progress_tracking_prerequisites(self, ctx: RuntimeContext) -> None:
```

src/databricks/labs/ucx/hive_metastore/workflows.py
```diff
@@ -57,10 +59,50 @@ def migrate_views(self, ctx: RuntimeContext):
...
+            migrate_views,
+            verify_prerequisites,
+        ],
+    )
+    def update_table_inventory(self, ctx: RuntimeContext) -> None:
+        """Refresh the tables inventory, prior to updating the migration status of all the tables."""
+        # The table inventory cannot be (quickly) crawled from the table_migration cluster, and the main cluster is not
+        # UC-enabled, so we cannot both snapshot and update the history log from the same location.
+        # Step 1 of 3: Just refresh the tables inventory.
+        ctx.tables_crawler.snapshot(force_refresh=True)
```
JCZuurmond, line 82: Does the inventory table for tables need updating before updating the migration status? Why?

src/databricks/labs/ucx/hive_metastore/table_migrate.py
JCZuurmond: And maybe rename it to `warn_about_remaining_tables`. I expect a `True` / `False` gate-like logic when the name is `check_`, though `check_` is also accurate to an extend

src/databricks/labs/ucx/hive_metastore/workflows.py
```diff
@@ -57,10 +59,50 @@ def migrate_views(self, ctx: RuntimeContext):
...
+
+    @job_task(depends_on=[verify_prerequisites, update_table_inventory], job_cluster="user_isolation")
+    def update_migration_status(self, ctx: RuntimeContext) -> None:
+        """Scan the tables (and views) in the inventory and record whether each has been migrated or not."""
+        # Step 2 of 3: Refresh the migration status of all the tables (updated in the previous step on the main cluster.)
+        updated_migration_progress = ctx.migration_status_refresher.snapshot(force_refresh=True)
+        ctx.tables_migrator.check_remaining_tables(updated_migration_progress)
```
JCZuurmond, line 89: See note about renaming this method,  in this context it is less clear to me what happens -> Will this fail the workflow if the checks fail?

src/databricks/labs/ucx/hive_metastore/workflows.py
```diff
@@ -57,10 +59,50 @@ def migrate_views(self, ctx: RuntimeContext):
...
+
+    @job_task(depends_on=[verify_prerequisites, update_migration_status], job_cluster="user_isolation")
+    def update_tables_history_log(self, ctx: RuntimeContext) -> None:
+        """Update the history log with the latest tables inventory and migration status."""
+        # Step 3 of 3: Assuming (due to depends-on) the inventory and migration status were refreshed, capture into the
+        # history log.
+        # TODO: Avoid triggering implicit refresh here if either the table or migration-status inventory is empty.
+        history_log = ctx.tables_progress
```
JCZuurmond, line 97: Redundant local variable IMO

```suggestion
```

src/databricks/labs/ucx/hive_metastore/workflows.py
```diff
@@ -86,10 +128,41 @@ def migrate_views(self, ctx: RuntimeContext):
         """
         ctx.tables_migrator.migrate_tables(what=What.VIEW)
 
-    @job_task(job_cluster="user_isolation", depends_on=[migrate_views])
-    def update_migration_status(self, ctx: RuntimeContext):
-        """Refresh the migration status to present it in the dashboard."""
-        ctx.tables_migrator.get_remaining_tables()
+    @job_task(job_cluster="user_isolation")
```
JCZuurmond, line 131: Similar comments goes for each table migration workflow

tests/integration/hive_metastore/test_workflows.py
```diff
@@ -11,11 +12,11 @@
         ("regular", "migrate-tables"),
         ("hiveserde", "migrate-external-hiveserde-tables-in-place-experimental"),
         ("hiveserde", "migrate-external-tables-ctas"),
+        # TODO: Some workflows are missing here, and also need to be included in the tests.
```
JCZuurmond, line 15: Add them

JCZuurmond (changes requested):
Added some comments

src/databricks/labs/ucx/hive_metastore/table_migrate.py
asnare: > And maybe rename it to `warn_about_remaining_tables`. I expect a `True` / `False` gate-like logic when the name is `check_`, though `check_` is also accurate to an extend

I've gone with the `warn_about_[…]` suggest: it's closer to what it does.

src/databricks/labs/ucx/hive_metastore/workflows.py
asnare: I've renamed this; indeed it would not fail because this was just about logging the names of tables that still need to be migrated.

(This was existing behaviour that I've left intact.)

src/databricks/labs/ucx/hive_metastore/table_migrate.py
JCZuurmond: Interesting, I use statuses: https://grammarhow.com/plural-of-status/. Still statuses is more clear to be plural than status, as status can be singular too

src/databricks/labs/ucx/hive_metastore/workflows.py
asnare: In terms of progress the `tables` and `migration_status` inventories are tightly coupled: refreshing the `migration_status` information involves iterating over the table inventory, the set of tables for which progress information is logged is based on the table inventory, and logging the that information involves enriching with with the `migration_status` information. To ensure the least amount of surprise, they are therefore both refreshed.

This is also the approach we take in the `migration-progress` workflow.

tests/integration/hive_metastore/test_workflows.py
asnare: The omitted workflows can't be trivially added to this list, and this was something I noted while working on the PR rather than a problem being introduced.

I've created #3422 to cover the integration tests for workflows that aren't being tested.

JCZuurmond (approved):

src/databricks/labs/ucx/hive_metastore/workflows.py
```diff
@@ -57,10 +59,53 @@ def migrate_views(self, ctx: RuntimeContext):
...
+    def verify_progress_tracking_prerequisites(self, ctx: RuntimeContext) -> None:
+        """Verify the prerequisites for running this job on the table migration cluster are fulfilled."""
+        ctx.verify_progress_tracking.verify(timeout=dt.timedelta(hours=1))
```
JCZuurmond, line 65: This forces the UCX catalog to be created **before** table migration, while it was not a pre-requiste before

JCZuurmond:
@asnare : When you are back, rebase with main and check if the CI still fails, the linked PR make the tests more robust

PR #3244, opened 2024-11-11
Tech Debt: Fixed issue with Incorrect unit test practice

### LINKED ISSUE #3126: [TODO] this is bad practice, we should not be mocking load() methon on a MockInstallation class
See: https://github.com/databrickslabs/ucx/blob/b97a8d54c42cbf335c7d60f68c0ed0713a414c99/tests/unit/aws/test_access.py#L519-L524

HariGS-DB:
## Changes
unit test case test_get_uc_compatible_roles is mocking the mock_installation.load to simulate save_uc_compatible_roles call through load_uc_compatible_roles. This change removes the mocking code and instead calls save_uc_compatible_roles directly which is what the code is testing.

Resolves #3126 

### Tests

- [ ] added unit tests

### DIFF
--- a/tests/unit/aws/test_access.py
+++ b/tests/unit/aws/test_access.py
@@ -1,13 +1,13 @@
 import json
 import logging
-from unittest.mock import MagicMock, call, create_autospec
+from unittest.mock import call, create_autospec
 
 import pytest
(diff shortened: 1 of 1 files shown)


nfx (approved):
lgtm

PR #3269, opened 2024-11-13
Update `migration-progress-experimental` workflow to crawl tables from the `main` cluster

### LINKED ISSUE #3268: [BUG]: Workflow `migration-progress-experimental` runs tables on the wrong job cluster

### Current Behavior

The `migration-progress-experimental` workflow refreshes the tables inventory from the `tacl` cluster. Although the task appears to complete it really doesn't do anything because the crawler can't use the Py4j bridge on that job cluster.

### Expected Behavior

Table crawling should happen from the `main` cluster, not `tacl`.

asnare:
## Changes

This PR updates the `migration-progress-experimental` workflow so that it uses the `main` cluster instead of the `tacl` one to crawl for tables.

Crawling from the `tacl` cluster fails: the Py4j bridge isn't available (which the crawler relies on).

### Linked issues

Resolves #3268 

### Functionality

### Tests

- manually tested
- existing integration tests

### DIFF
--- a/src/databricks/labs/ucx/progress/workflows.py
+++ b/src/databricks/labs/ucx/progress/workflows.py
@@ -31,12 +31,7 @@ class MigrationProgress(Workflow):
         """
         ctx.verify_progress_tracking.verify(timeout=dt.timedelta(hours=1))
 
-    @job_task(job_cluster="tacl")
-    def setup_tacl(self, ctx: RuntimeContext):
-        """(Optimization) Allow the TACL job cluster to be started while we're verifying the prerequisites for
-        refreshing everything."""
-
-    @job_task(depends_on=[verify_prerequisites, setup_tacl], job_cluster="tacl")
+    @job_task(depends_on=[verify_prerequisites])
     def crawl_tables(self, ctx: RuntimeContext) -> None:
         """Iterates over all tables in the Hive Metastore of the current workspace and persists their metadata, such
         as _database name_, _table name_, _table type_, _table location_, etc., in the table named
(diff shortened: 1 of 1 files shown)


asnare:
Note: `build / ci (3.10)` is failing due to the coverage check. However the coverage on `main` currently is less than the required threshold: 89.96%. (This PR actually improves it by a very small amount.)

src/databricks/labs/ucx/progress/workflows.py
```diff
@@ -31,34 +31,31 @@ def verify_prerequisites(self, ctx: RuntimeContext) -> None:
...
         depends_on=[verify_prerequisites, crawl_tables, refresh_table_migration_status], job_cluster="table_migration"
     )
     def update_tables_history_log(self, ctx: RuntimeContext) -> None:
         """Update the history log with the latest tables inventory snapshot."""
-        # The table migration cluster is not legacy-ACL enabled, so we can't crawl from here.
-        # Step 2 of 2: Assuming (due to depends-on) the inventory was refreshed, capture into the history log.
-        # WARNING: this will fail if the inventory is empty, because it will then try to perform a crawl.
+        # Step 3 of 3: Assuming (due to depends-on) the inventory and migration status were refreshed, capture into the
+        # history log.
+        # WARNING: this will fail if the tables inventory is empty, because it will then try to perform a crawl.
```
nfx, line 58: technically, we can catch the exception and trigger the wait for an assessment workflow within this job. technically, this i a TODO - can you rewrite it as such, so it becomes a tech debt issue?

nfx (approved):
fix `ci` on py3.10

src/databricks/labs/ucx/progress/workflows.py
asnare: Rephrasing this as a TODO is a good idea, which I've done.

Note that currently there is no exception to catch: the fast crawler swallows the exceptions that it encounters if the Py4j bridge is unavailable, and we end up with an empty crawl instead of a failed one. (I've just created #3271 to address that: it should indeed fail noisily.)

asnare:
Integration test failures seem to be due to #3273.

PR #3270, opened 2024-11-13
Refactor refreshing of migration-status information for tables, eliminate another redundant refresh.

asnare:
## Changes

This PR updates the way table records are enriched with migration-status information during encoding for the history log. Changes include:

 - Ensuring refresh of the migration-status information is explicit and under control of the workflow, an intent expressed [here](https://github.com/databrickslabs/ucx/pull/2743#discussion_r1783326822) when the current design was laid out.
 - A redundant refresh of the migration-status information is eliminated. (This took place in the `update_tables_history_log` task, even though it had just occurred in the `refresh_table_migration_status` task immediately prior to it.)  
 - Some additional (unit) test coverage of the `migration-progress-experimental` workflow.

### Linked issues

Splits #3239.

### Functionality

### Tests

- added/updated unit tests
- existing integration tests


src/databricks/labs/ucx/contexts/workflow_task.py
```diff
@@ -189,7 +189,7 @@ def policies_progress(self) -> ProgressEncoder[PolicyInfo]:
         )
 
     @cached_property
-    def grants_progress(self) -> GrantProgressEncoder:
+    def grants_progress(self) -> ProgressEncoder[Grant]:
```
JCZuurmond, line 192: Why change this type hint?

src/databricks/labs/ucx/progress/tables.py
```diff
@@ -38,17 +45,24 @@ def __init__(
             schema,
             table,
         )
-        self._table_migration_index = table_migration_index
+        self._migration_status_refresher = migration_status_refresher
+
+    def append_inventory_snapshot(self, snapshot: Iterable[Table]) -> None:
+        migration_index = TableMigrationIndex(self._migration_status_refresher.snapshot())
```
JCZuurmond, line 51: We commonly use a `cached_property` for this "caching" of overwriting the `append_inventory_snapshot1

tests/unit/progress/test_workflows.py
```diff
@@ -47,22 +48,31 @@ def test_migration_progress_runtime_refresh(run_workflow, task, crawler, history
...
+        "migration_status_refresher": mock_migration_status_refresher,
         "tables_progress": mock_history_log,
         "named_parameters": {"parent_run_id": 53},
     }
 
-    # The first part of a 2-step update: the crawl without updating the history log.
+    # The first part of a 3-step update: the table crawl without updating the history log.
     run_workflow(MigrationProgress.crawl_tables, **context_replacements)
     mock_tables_crawler.snapshot.assert_called_once_with(force_refresh=True)
+    mock_tables_crawler.snapshot.reset_mock()
```
JCZuurmond, line 63: Did not know this, nice!

Do you prefer a reset over separate tests?

JCZuurmond (changes requested):
See pointers

src/databricks/labs/ucx/contexts/workflow_task.py
asnare: The interface this exposes (via the type) is the encoder interface; the fact that we have a specialisation in place is an implementation detail. (It also discourages additional public methods from appearing on the specialisation.)

src/databricks/labs/ucx/progress/tables.py
asnare: That would be wrong, because the snapshot has to correspond to the snapshot that is passed in. The caller is responsible for controlling when refreshes occur, and caching this internally would defeat the control of the caller being able to control that.

tests/unit/progress/test_workflows.py
asnare: No, I would normally say we should _never_ reset mocks.

However in this test we are checking a sequence of operations that need to work together, and resetting allows the test to verify the interactions are happening at the correct point (rather than accumulating).

This _could_ be split into a sequence of tests (one for each task), but grouping them here in this way emphasises that the tasks are intended to run in a particular sequence.

asnare: I just realised that the test did nothing to verify the task ordering, so I've added that in.

With that in mind, what would you prefer: like this, or split into separate unit tests? I'm fine with either approach…

asnare:
Integration tests currently failing due to a dependency issue on `main`; resolved by #3312.

tests/unit/progress/test_workflows.py
JCZuurmond: Generally I prefer separate unit tests over big unit tests

JCZuurmond (changes requested):
Generally LGTM. See comment on splitting tests, resolve it and request another review

JCZuurmond (approved):
LGTM

PR #3318, opened 2024-11-16
Added noisy failures when cluster encounters Py4J error

### LINKED ISSUE #3271: [BUG]: `FasterTableScanCrawler` returns an empty inventory instead of failing if the Py4j bridge isn't available

### Current Behavior

The `FasterTableScanCrawler` is used to crawl tables on the runtime contexts, and it uses the Py4j bridge to reach Spark and crawl using its catalog.

Unfortunately there are 2 problems right now:
1. Not all runtime contexts have the Py4j bridge available.
2. When the Py4j bridge is unavailable, crawling logs warnings (not errors) but does not fail. Instead it returns an empty snapshot.

The latter has masked downstream bugs where crawling took place on the wrong runtime cluster, but instead of this producing a noisy failure instead an empty inventory is the result. 

### Expected Behavior

The `FasterTableScanCrawler` should fail properly with an exception if the Py4j bridge isn't available: this indicates it's being used in an unsuitable (runtime) context.

pritishpai:
## Changes
Added noisy failures when cluster encounters Py4J error instead of swallowing the error with a warning.

### Linked issues
Resolves #3271 

### Functionality

- [ ] handle error correctly

### Tests
- [x] manually tested
- [x] verified on labs environment (screenshot attached)


pritishpai:
image

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -544,13 +544,29 @@ def _all_databases(self) -> list[str]:
                 return list(self._iterator(self._external_catalog.listDatabases()))
             return self._include_database
         except Exception as err:  # pylint: disable=broad-exception-caught
+            if "py4j.security.Py4JSecurityException" in str(err):
+                logger.error(
+                    "Failed to list databases due to Py4JSecurityException. "
+                    "Ensure that the cluster is configured correctly."
```
nfx, line 550: `Ensure that the cluster is configured correctly.` - how? you should not assume that all users know the answer. option here is to update UCX. or reinstall it as the last resort.

nfx (changes requested):

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -544,13 +544,29 @@ def _all_databases(self) -> list[str]:
                 return list(self._iterator(self._external_catalog.listDatabases()))
             return self._include_database
         except Exception as err:  # pylint: disable=broad-exception-caught
+            if "py4j.security.Py4JSecurityException" in str(err):
+                logger.error(
+                    "Failed to list databases due to Py4JSecurityException. "
+                    "Update or reinstall UCX to resolve this issue."
+                    f"Error details: {err}",
```
JCZuurmond, line 551: Not needed when adding `exc_info=True`

 ```suggestion
```

File: src/databricks/labs/ucx/hive_metastore/tables.py [hunk 3]
```diff
@@ -544,13 +544,29 @@ def _all_databases(self) -> list[str]:
...
 
     def _list_tables(self, database: str) -> list[str]:
         try:
             return list(self._iterator(self._external_catalog.listTables(database)))
         except Exception as err:  # pylint: disable=broad-exception-caught
+            if "py4j.security.Py4JSecurityException" in str(err):
+                logger.error(
+                    "Failed to list databases due to Py4JSecurityException. "
+                    "Update or reinstall UCX to resolve this issue."
+                    f"Error details: {err}",
```
JCZuurmond, line 566: Same here

```suggestion
```

JCZuurmond (changes requested):
Minor changes, otherwise good to go

src/databricks/labs/ucx/hive_metastore/tables.py
```diff
@@ -544,13 +544,27 @@ def _all_databases(self) -> list[str]:
                 return list(self._iterator(self._external_catalog.listDatabases()))
             return self._include_database
         except Exception as err:  # pylint: disable=broad-exception-caught
+            if "py4j.security.Py4JSecurityException" in str(err):
+                logger.error(
+                    "Failed to list databases due to Py4JSecurityException. "
+                    "Update or reinstall UCX to resolve this issue.",
+                    exc_info=True,
```
JCZuurmond, line 551: Fyi, I prefer, as it is more explicit about which error info is logged

```suggestion
                    exc_info=err,
```

PR #3375, opened 2024-11-22
Track `DirectFsAccess` on `JobsProgressEncoder`

### LINKED ISSUE #3059: [FEATURE]: History log encoder for jobs

### Problem statement

For creating a history log from a job, we need an encoder that captures the `failures`.

### Proposed Solution

Create a history log encoder for jobs to encode the `failures`.

### Additional Context

_No response_

JCZuurmond:
## Changes

Track `DirectFsAccess` on `JobsProgressEncoder`

### Linked issues

Resolves #3059

### Functionality

- [x] modified existing workflow: `migration-progress-experimental`

### Tests

- [x] added unit tests
- [x] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/progress/jobs.py
+++ b/src/databricks/labs/ucx/progress/jobs.py
@@ -17,7 +17,7 @@ class JobsProgressEncoder(ProgressEncoder[JobInfo]):
         self,
         sql_backend: SqlBackend,
         ownership: JobOwnership,
-        direct_fs_access_crawler: DirectFsAccessCrawler,
+        direct_fs_access_crawlers: list[DirectFsAccessCrawler],
         inventory_database: str,
         run_id: int,
         workspace_id: int,
@@ -35,7 +35,7 @@ class JobsProgressEncoder(ProgressEncoder[JobInfo]):
             schema,
             table,
         )
-        self._direct_fs_access_crawler = direct_fs_access_crawler
+        self._direct_fs_access_crawlers = direct_fs_access_crawlers
         self._inventory_database = inventory_database
 
     @cached_property
@@ -54,18 +54,19 @@ class JobsProgressEncoder(ProgressEncoder[JobInfo]):
     @cached_property
     def _direct_fs_accesses(self) -> dict[str, list[str]]:
         index = collections.defaultdict(list)
-        for direct_fs_access in self._direct_fs_access_crawler.snapshot():
-            # The workflow and task source lineage are added by the WorkflowLinter
-            if len(direct_fs_access.source_lineage) < 2:
-                continue
-            if direct_fs_access.source_lineage[0].object_type != "WORKFLOW":
-                continue
-            if direct_fs_access.source_lineage[1].object_type != "TASK":
-                continue
-            job_id = direct_fs_access.source_lineage[0].object_id
-            task_key = direct_fs_access.source_lineage[1].object_id  # <job id>/<task key>
-            failure = f"Direct file system access by '{task_key}' in '{direct_fs_access.source_id} to '{direct_fs_access.path}'"
-            index[job_id].append(failure)
+        for crawler in self._direct_fs_access_crawlers:
+            for direct_fs_access in crawler.snapshot():
+                # The workflow and task source lineage are added by the WorkflowLinter
+                if len(direct_fs_access.source_lineage) < 2:
+                    continue
+                if direct_fs_access.source_lineage[0].object_type != "WORKFLOW":
+                    continue
+                if direct_fs_access.source_lineage[1].object_type != "TASK":
+                    continue
+                job_id = direct_fs_access.source_lineage[0].object_id
+                task_key = direct_fs_access.source_lineage[1].object_id  # <job id>/<task key>
+                failure = f"Direct file system access by '{task_key}' in '{direct_fs_access.source_id} to '{direct_fs_access.path}'"
+                index[job_id].append(failure)
         return index
 
     def _encode_record_as_historical(self, record: JobInfo) -> Historical:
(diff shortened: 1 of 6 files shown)


src/databricks/labs/ucx/source_code/directfs_access.py
```diff
@@ -50,7 +50,7 @@ def dump_all(self, dfsas: Sequence[DirectFsAccess]) -> None:
         Providing a multi-entity crawler is out-of-scope of this PR
         """
         try:
-            # TODO until we historize data, we append all DFSAs
+            # TODO: @JCZuurmond until we historize data, we append all DFSAs
```
JCZuurmond, line 53: @asnare : Do you recall what we should do here? As we now are historizing the DFSAs, still we need to get the snapshot from the inventory

src/databricks/labs/ucx/progress/jobs.py
```diff
@@ -7,21 +7,22 @@
 from databricks.labs.ucx.assessment.jobs import JobInfo, JobOwnership
 from databricks.labs.ucx.progress.history import ProgressEncoder
 from databricks.labs.ucx.progress.install import Historical
+from databricks.labs.ucx.source_code.directfs_access import DirectFsAccessCrawler
 from databricks.labs.ucx.source_code.jobs import JobProblem
 
 
 class JobsProgressEncoder(ProgressEncoder[JobInfo]):
+    """Encoder class:Grant to class:History."""
```
asnare, line 15: Copypasta?

src/databricks/labs/ucx/progress/jobs.py
```diff
@@ -48,7 +50,36 @@ def _job_problems(self) -> dict[int, list[str]]:
...
+                if direct_fs_access.source_lineage[1].object_type != "TASK":
+                    continue
+                job_id = direct_fs_access.source_lineage[0].object_id
+                task_key = direct_fs_access.source_lineage[1].object_id  # <job id>/<task key>
+                # Follow same failure message structure as the JobProblems above and DirectFsAccessPyLinter deprecation
+                code = "direct-filesystem-access"
+                message = f"The use of direct filesystem references is deprecated: {direct_fs_access.path}"
+                failure = f"{code}: {task_key} task: {direct_fs_access.source_id}: {message}"
+                index[job_id].append(failure)
+        return index
```
asnare, line 72: This shouldn't be part of a cached property because there's no way to reset it (or signal a refresh) and the encoder itself is a singleton.

Instead, look at how we do this in `TableEncoder`:

 - Override `.append_inventory_snapshot()`, so it loads the enrichment data.
 - Implement a new per-record encoder, for example `._encode_table_as_historical(record, extra, things, needed, for, enrichment)`
 - The new per-record encoder invokes `super().encode_record_as_historical(record)` and then performs enrichment.

This achieves the goals of only loading the data once (per snapshot encode) while not inadvertently keeping (potentially stale) data cached.

src/databricks/labs/ucx/progress/jobs.py
```diff
@@ -48,7 +50,36 @@ def _job_problems(self) -> dict[int, list[str]]:
...
+        Failures are detected by the WorkflowLinter:
+        - Job problems
+        - Direct filesystem access by code used in job
+        """
         historical = super()._encode_record_as_historical(record)
-        failures = self._job_problems.get(int(record.job_id), [])
+        failures = []
+        failures.extend(self._job_problems.get(int(record.job_id), []))
+        failures.extend(self._direct_fs_accesses.get(record.job_id, []))
         return replace(historical, failures=historical.failures + failures)
```
asnare, line 85: See above; the existing `_encode_record_as_historical()` shouldn't need to be overridden; instead an adjacent method should be implemented.

src/databricks/labs/ucx/source_code/directfs_access.py
asnare: The intent here is that a crawl should override the existing data. In this case we were instead appending so that history would be kept, pending implementation of the history log.

So this can be updated to overwrite instead of append, with the major proviso that we check to make sure the data is being written once for a crawl. (For a while a single "crawl" was writing its data across several operations, which won't work as intended if the mode for such operations is changed to `overwrite`.)

asnare (changes requested):
Goal and purpose is correct, the implementation needs to be updated. (I haven't received the tests yet.)

src/databricks/labs/ucx/progress/jobs.py
JCZuurmond: I saw that and it is better yes, but decided not to rewrite it here to separate feature from refactor

JCZuurmond: https://github.com/databrickslabs/ucx/issues/3377

src/databricks/labs/ucx/progress/jobs.py
JCZuurmond: https://github.com/databrickslabs/ucx/issues/3377

src/databricks/labs/ucx/source_code/directfs_access.py
JCZuurmond: I decide to not update the mode to "overwrite" yet as we do not historise the DFSA from QueryLinter (yet). 

I did confirm there is one single "crawl" per run - of `WorkflowLinter` and `QueryLinter` separately (different DFSAs).

src/databricks/labs/ucx/progress/jobs.py
```diff
@@ -48,7 +50,36 @@ def _job_problems(self) -> dict[int, list[str]]:
             index[job_problem.job_id].append(failure)
         return index
 
+    @cached_property
+    def _direct_fs_accesses(self) -> dict[str, list[str]]:
```
FastLee, line 54: Why should we keep it as a cached property?

JCZuurmond: To compute the property only once

FastLee (approved):
LGTM

PR #3379, opened 2024-11-25
Tech debt: avoid caching with some test context properties

asnare:
## Changes

This PR modifies some properties on the integration test context so that they don't cache their return values; these properties are based on internal mutable state and the current caching situation makes it difficult to know if they are yielding the current value or an earlier one that might no longer be correct.

### Tests

- existing integration tests


JCZuurmond (approved):
LGTM

PR #3395, opened 2024-11-29
Remove redundant internal methods from create_account_group

### LINKED ISSUE #3170: [TODO] remove this method and rely on _include_workspace_ids
See: https://github.com/databrickslabs/ucx/blob/b97a8d54c42cbf335c7d60f68c0ed0713a414c99/src/databricks/labs/ucx/account/workspaces.py#L142-L147

HariGS-DB:
## Changes
changed the signature of create_account_groups and removed the workspace_id parameter and instead retrieve it from accountworkspace._workspaces() which contains the account context workspace_id named parameter

Resolves #3170 

### Functionality

### Tests

- [ ] added unit tests


src/databricks/labs/ucx/account/workspaces.py
```diff
@@ -76,9 +76,11 @@ def sync_workspace_info(self, workspaces: list[Workspace] | None = None):
             except (PermissionDenied, NotFound, ValueError):
                 logger.warning(f"Failed to save workspace info for {ws.config.host}")
 
-    def create_account_level_groups(self, prompts: Prompts, workspace_ids: list[int] | None = None):
+    def create_account_level_groups(self, prompts: Prompts):
         acc_groups = self._get_account_groups()
-        workspace_ids = self._get_valid_workspaces_ids(workspace_ids)
+        workspace_ids = [workspace.workspace_id for workspace in self._workspaces()]
+        if not workspace_ids:
+            raise ValueError("No workspace ids provided in the configuration found in the account")
```
JCZuurmond, line 83: I see the message has not changed, however, I think it is inaccurate. There is not such thing as a configuration in the **account**. Could you update it to be accurate?

JCZuurmond (approved):
Changes LGTM, added a note about something that was already there

Issue #2854, opened 2024-10-07
[FEATURE]: Include Lakeview Dashboards as part of Assessment/Migration


allan-cao-databricks:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

The Assessment Workflow currently does not crawl permissions tied to Lakeview Dashboards (only legacy dashboards) as part of the `crawl_grants` task. Would like to request for them to be added as both part of the Assessment Workflow, but also the ACLs migration process.

### Proposed Solution

- Include Lakeview Dashboards as part of `workspace_listing`/`crawl_grants`
- Populate results of permissions under `ucx.permissions`
- Propagate privileges down to group migration workflow

nfx:
there are 3 things to be done:

1. assessment
2. migrate permissions
3. migrate dashboards

PR #3397, opened 2024-12-02
Add dashboard crawlers

### LINKED ISSUE #3366: [FEATURE] Redash Dashboards Crawler
### Description

Create a crawler for Redash dashboards so that we can track them during the migration progress.

### Considerations

- Use the crawler for dashboard migration

### Related

- #3367

### LINKED ISSUE #3367: [FEATURE] Lakeview Dashboards Crawler
### Description

Create a crawler for Lakeview dashboards so that we can track them during the migration progress.

### Considerations

- Use the crawler for dashboard and permission migration

### Related

- #3366
- #2854

JCZuurmond:
## Changes
Add dashboard crawlers and use them:
- In the assessment workflow to crawl and persist the dashboards
- In the Redash migration to migrate or revert the crawled dashboards
- In the QueryLinter to lint the queries of the crawled dashboards

### Linked issues

Resolves #3366
Resolves #3367
Progresses #2854

### Functionality

- [ ] modified existing command: `databricks labs ucx {migrate-dbsql-dashboards|revert-dbsql-dashboards}`
- [ ] modified existing workflow: `assessment`

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


pyproject.toml
```diff
@@ -600,7 +600,8 @@ disable = [
     "consider-using-any-or-all",
     "too-many-positional-arguments",
     "unnecessary-default-type-args",
-    "logging-not-lazy"
+    "logging-not-lazy",
+    "too-many-public-methods",  # TODO: Remove by someone who can bypass CI cheat linter check
```
JCZuurmond, line 604: Had to add this as we are missing someone who can bypass the CI. Too many public methods on the Assessment workflow with the crawl dashboards task

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,359 @@
...
+        return cls(
+            id=dataset.name,
+            name=dataset.display_name or cls.name,
+            parent=parent or cls.parent,
+            query=dataset.query,
+        )
+
+
+@dataclass
+class RedashDashboard:
```
JCZuurmond, line 67: I might merge this with the LakeviewDashboard object, only difference are the tags

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+    """The identifier of the workspace folder containing the object."""
+
+    query: str = ""
+    """The text of the query to be run."""
+
+    catalog: str = ""
+    """The name of the catalog to execute this query in."""
+
+    schema: str = ""
+    """The name of the schema to execute this query in."""
```
asnare, line 48: I don't think we should have default values here: it's an error if they have not been set, and assigning them a (nonsense) default masks errors because we end up persisting nonsensical data.

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+        if query.options:
+            catalog = query.options.catalog
+            schema = query.options.schema
+        return cls(
+            id=query.id,
+            name=query.name or cls.name,
+            parent=query.parent or cls.parent,
+            query=query.query or cls.query,
+            catalog=catalog or cls.catalog,
+            schema=schema or cls.schema,
```
asnare, line 67: These shouldn't fall back on class-defined defaults. If the values may not be available then our model should define them as optional. If they're mandatory, then it's an error and something else is wrong.

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+            tags=query.tags or [],
+        )
+
+    @classmethod
+    def from_lakeview_dataset(cls, dataset: Dataset, *, parent: str | None = None) -> Query:
+        """Create query from a :class:Dataset"""
+        return cls(
+            id=dataset.name,
+            name=dataset.display_name or cls.name,
+            parent=parent or cls.parent,
```
asnare, line 77: The same comment applies here: if these values may not be available then we should make the fields optional rather than filling them with class-defined sentinels or some other default value.

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+
+    id: str
+    """The ID for this dashboard."""
+
+    name: str = "UNKNOWN"
+    """The title of the dashboard that appears in list views and at the top of the dashboard page."""
+
+    parent: str = "ORPHAN"
+    """The identifier of the workspace folder containing the object."""
+
```
asnare, line 94: As with `Query` these should be optional and be left empty if a value is not available.

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+                continue
+            if widget.visualization.query is None:
+                continue
+            if widget.visualization.query.id is None:
+                continue
+            query_ids.append(widget.visualization.query.id)
+        return cls(
+            id=dashboard.id or cls.id,
+            name=dashboard.name or cls.name,
+            parent=dashboard.parent or cls.parent,
```
asnare, line 115: The same comment as elsewhere applies.

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+
+    @classmethod
+    def from_sdk_lakeview_dashboard(cls, dashboard: SdkLakeviewDashboard) -> Dashboard:
+        assert dashboard.dashboard_id
+        lsql_dashboard = _convert_sdk_to_lsql_lakeview_dashboard(dashboard)
+        query_ids = [dataset.name for dataset in lsql_dashboard.datasets]
+        return cls(
+            id=dashboard.dashboard_id,
+            name=dashboard.display_name or cls.name,
+            parent=dashboard.parent_path or cls.parent,
```
asnare, line 128: The same comment as elsewhere applies.

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+        schema: str,
+        *,
+        include_dashboard_ids: list[str] | None = None,
+        include_query_ids: list[str] | None = None,
+        debug_listing_upper_limit: int | None = None,
+    ):
+        super().__init__(sql_backend, "hive_metastore", schema, "redash_dashboards", Dashboard)
+        self._ws = ws
+        self._include_dashboard_ids = include_dashboard_ids or []
+        self._include_query_ids = include_query_ids or []
```
asnare, line 149: I think this precludes us from configuring it with an empty list (`[]`), the obvious way to prevent it from crawling any dashboards/queries.

It would be cleaner to leave `None` as meaning "crawl everything".

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+        # to a small number of items in debug mode for the assessment workflow just to complete.
+        while self._debug_listing_upper_limit is None or self._debug_listing_upper_limit > len(dashboards):
+            try:
+                dashboards.append(next(dashboards_iterator))
+            except StopIteration:
+                break
+            except DatabricksError as e:
+                logger.warning("Cannot list next Redash dashboards page", exc_info=e)
+                break
+        return dashboards
```
asnare, line 180: I think this can be simplified somewhat:
 - SDK calls to `.list()` normally allow a maximum number of results.
 - `itertools.islice()` can be used to ignore everything beyond the first $$n$$ results.

That would lead to something like:
```python
try:
    if self._debug_listing_upper_limit is None:
        dashboards_iterator = self._ws.dashboards.list()
    else:
        dashboards_iterator = itertools.islice(self._ws.dashboards.list(page_size=self._debug_listing_upper_limit), self._debug_listing_upper_limit)  # might also need page=0 or page=1 or something.
    # Force SDK errors to be materialised here rather than deferred.
    return list(dashboards_iterator)
except DatabricksError as e:
    logger.warning("Cannot list Redash dashboards", exc_info=e)
    return []
```
This also eliminates handling the same error twice.

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+        else:
+            queries_iterator = self._list_all_legacy_queries()
+        # Redash APIs are very slow to paginate, especially for large number of dashboards, so we limit the listing
+        # to a small number of items in debug mode for the assessment workflow just to complete.
+        counter = itertools.count()
+        while self._debug_listing_upper_limit is None or self._debug_listing_upper_limit > next(counter):
+            try:
+                yield next(queries_iterator)
+            except StopIteration:
+                break
```
asnare, line 237: This can also be simplified using `itertools.islice()`.

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+                yield from self._ws.queries_legacy.list()  # TODO: Update this to non-legacy query
+            except DatabricksError as e:
+                logger.warning("Cannot list Redash queries", exc_info=e)
+
+    def _list_legacy_queries_from_dashboard(self, dashboard: Dashboard) -> Iterator[LegacyQuery]:
+        """List queries from dashboard."""
+        if self._include_query_ids:
+            query_ids = set(dashboard.query_ids) & set(self._include_query_ids)
+        else:
+            query_ids = set(dashboard.query_ids)
```
asnare, line 254: I normally prefer this for immutable collections. For mutable ones an alternative could be:
```suggestion
        query_ids = set(dashboard.query_ids)
        if self._include_query_ids:
            query_ids &= set(self._include_query_ids)
```

JCZuurmond:
The integration test CI only fails for test that is also failing on `test_running_real_assessment_job_ext_hms`

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+        sql_backend: SqlBackend,
+        schema: str,
+        *,
+        include_dashboard_ids: list[str] | None = None,
+        include_query_ids: list[str] | None = None,
+    ):
+        super().__init__(sql_backend, "hive_metastore", schema, "lakeview_dashboards", Dashboard)
+        self._ws = ws
+        self._include_dashboard_ids = include_dashboard_ids or []
+        self._include_query_ids = include_query_ids or []
```
asnare, line 304: The same comment as earlier applies here about what `[]` means as an argument.

tests/integration/assessment/test_workflows.py
```diff
@@ -31,6 +31,7 @@ def test_running_real_assessment_job(
     tmp_table = installation_ctx.make_table(schema_name=source_schema.name, ctas="SELECT 2+2 AS four")
     view = installation_ctx.make_table(schema_name=source_schema.name, ctas="SELECT 2+2 AS four", view=True)
     non_delta = installation_ctx.make_table(schema_name=source_schema.name, non_delta=True)
+
```
JCZuurmond, line 34: Making sure this tests passes

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,370 @@
...
+            This public method does not adhere to the common crawler layout, still, it is implemented to avoid/postpone
+            another crawler for the queries by retrieving the queries every time they are requested.
+
+            Different to the Redash crawler, Lakeview queries are part of the (serialized) dashboard definition.
+        """
+        sdk_dashboards = []
+        if dashboard:
+            sdk_dashboard = self._get_dashboard(dashboard_id=dashboard.id)
+            if sdk_dashboard:
+                sdk_dashboards.append(sdk_dashboard)
```
asnare, line 362: What do you think of this?
```suggestion
        if dashboard:
            sdk_dashboard = self._get_dashboard(dashboard_id=dashboard.id)
            sdk_dashboards = [sdk_dashboard] if sdk_dashboard else []
```

src/databricks/labs/ucx/source_code/queries.py
```diff
@@ -149,140 +154,80 @@ def _lint_queries(self, context: _ReportingContext) -> None:
...
-        for query in self._ws.queries_legacy.list():
-            if self._debug_listing_upper_limit is not None and items_listed >= self._debug_listing_upper_limit:
-                logger.warning(f"Debug listing limit reached: {self._debug_listing_upper_limit}")
-                break
-            legacy_queries.append(query)
-            items_listed += 1
-        return legacy_queries
+    def _list_queries(self) -> Iterable[Query]:
+        for crawler in self._dashboard_crawlers:
+            yield from crawler.list_queries()
```
asnare, line 159: Ah, now I understand.

src/databricks/labs/ucx/source_code/redash.py
```diff
@@ -18,41 +18,44 @@
...
 
-    def _list_dashboards(self, dashboard_id: str | None) -> list[Dashboard]:
-        try:
-            if dashboard_id is None:
-                return list(self._ws.dashboards.list())
-            return [self._ws.dashboards.get(dashboard_id)]
-        except DatabricksError as e:
-            logger.warning(f"Cannot list dashboards: {e}")
-            return []
+    def _list_dashboards(self, *dashboard_ids: str, force_refresh: bool = False) -> list[Dashboard]:
```
asnare, line 51: We shouldn't allow forcing a refresh here: that's supposed to be limited to just workflow tasks (directly) because we want it to be clear where refreshes are happening.

If the caller expects a refresh, it needs to be explicitly done by the workflow beforehand.

asnare (changes requested):
I've reviewed the non-test part of this code; and will continue with the tests tomorrow. Mostly nits and potential improvements rather than anything specifically related to correctness.

src/databricks/labs/ucx/source_code/redash.py
JCZuurmond: This is difficult as the revert dashboard requires the latest tags (to include the migration tag) and the dashboard migration is a cli command, not a workflow.

To resolve your comment, the dashboard migration needs to be transpiled to a workflow, which I do not want to include in the scope of this PR

asnare: If not the workflow, then the CLI should do this: the intent is that explicit refreshing is supposed to be a 'top-level' action and not occur implicitly during other activity.

src/databricks/labs/ucx/assessment/dashboards.py
JCZuurmond: Though, I like the use of `islice` it does not reach the same functionality, see `test_redash_dashboard_crawler_handles_databricks_error_on_iterate`. If we get a `TooManyRequests` half way listing, it returns the dashboards listed until then with the current code, with your seggestion it returns no dashboards

src/databricks/labs/ucx/assessment/dashboards.py
JCZuurmond: https://github.com/databrickslabs/ucx/pull/3397#discussion_r1877602313

src/databricks/labs/ucx/source_code/queries.py
JCZuurmond: Yeah, similarly to the notebook and such, we do not persist those resources in the inventory. (I think we should not do that.) However, that means we need to retrieve the code resources dynamically, thus I added this public method

src/databricks/labs/ucx/assessment/dashboards.py
asnare: That's a good concern, but I don't think it can happen? There's only a single call being made (`.list()`) which will either succeed with (a page of) results or fail immediately.

I don't think that unit test emulates an error that can occur; the only point that `TooManyRequests` can be raised is immediately upon receipt of the HTTP request, _before_ the first dashboard is returned. (Due to the way coroutines work raising the error is probably deferred until the _first_ dashboard is yielded, but after that I don't think it can happen and the consequence is the same: an empty list.)

Thinking about it some more… arguably the crawler for this particular case should also use a retry-handler: there's something to be said for returning partial results, but if the `.list()` itself fails, especially due to `TooManyRequests`, then it should retry after the delay including in the exception. I'm also a little uncomfortable about just returning an empty list: under the current code we've had a failure that prevents it from working properly but the `assessment` will succeed. What do you think about this?

JCZuurmond: yeah, the difficulty here is that I agree, but want to code defensively and avoid too many assumptions about the API inner working.

On the single `list`, it might call the API multiple times, see https://github.com/databricks/databricks-sdk-py/blob/980166030c688e08a52d0f55389abd766f1d928d/databricks/sdk/service/dashboards.py#L1350-L1357 

I expect if we set the page size to the debug limit, then it should call API just once, but I am not 100% certain

JCZuurmond: I am leaning to keeping the code as it is

JCZuurmond: we could log the error as an `logger.error` to make sure the assessment workflow fails

JCZuurmond:
Only `test_running_real_assessment_job_ext_hms` is failing in [last integration test run](https://github.com/databrickslabs/ucx/actions/runs/12259217739/job/34200847904?pr=3397)

tests/integration/source_code/test_queries.py
```diff
@@ -1,60 +1,77 @@
...
+            query_name=query_with_dfsa.name,
+            code='direct-filesystem-access-in-sql-query',
+            message='The use of direct filesystem references is deprecated: dbfs://some_folder/some_file.csv',
+        )
+    ]
+
+    dfsas = list(simple_ctx.directfs_access_crawler_for_queries.snapshot())
+    # By comparing the element instead of the list the `field(compare=False)` of the dataclass attributes take effect
+    assert len(dfsas) == 1, "Expected one DFSA"
+    assert dfsas[0] == DirectFsAccess(
```
asnare, line 31: Same suggestion as elsewhere, although to be clear it's not a strong suggestion but rather something to consider.

tests/integration/source_code/test_queries.py
```diff
@@ -1,60 +1,77 @@
...
+        is_read=True,
+        is_write=False,
+    )
+
+    used_tables = list(simple_ctx.used_tables_crawler_for_queries.snapshot())
+    # By comparing the element instead of the list the `field(compare=False)` of the dataclass attributes take effect
+    assert len(used_tables) == 1, "Expected one used table"
+    # The "query" in the source and object id, and "count" in the name are hardcoded in the
+    # `make_lakeview_dashboard` fixture
+    assert used_tables[0] == UsedTable(
```
asnare, line 55: … and here.

tests/unit/assessment/test_dashboards.py
```diff
@@ -0,0 +1,647 @@
...
+
+def test_redash_dashboard_crawler_list_queries_includes_query_ids(mock_backend) -> None:
+    ws = create_autospec(WorkspaceClient)
+    ws.queries_legacy.list.side_effect = list_legacy_queries
+    ws.queries_legacy.get.side_effect = get_legacy_query
+    crawler = RedashDashboardCrawler(ws, mock_backend, "test", include_query_ids=["qid1"])
+
+    queries = list(crawler.list_queries())
+
+    assert queries == [Query(id="qid1", name="First query", parent="parent", query="SELECT 42 AS count")]
```
asnare, line 239: 😍

tests/unit/assessment/test_dashboards.py
```diff
@@ -0,0 +1,647 @@
...
+
+    assert len(queries) == 0
+    assert "Cannot get Redash query: qid" in caplog.messages
+    ws.queries_legacy.get.assert_called_once_with("qid")
+
+
+def test_redash_dashboard_crawler_list_queries_stops_when_debug_listing_upper_limit_reached(mock_backend) -> None:
+    ws = create_autospec(WorkspaceClient)
+    legacy_queries = [LegacyQuery(id="qid1"), LegacyQuery(id="qid2")]
+    ws.queries_legacy.list.side_effect = lambda: (query for query in legacy_queries)
```
asnare, line 345: Subtle, but nice. 👍

tests/unit/source_code/test_redash.py
```diff
@@ -1,187 +1,145 @@
...
-    )
-
-    return workspace_client
+    query_mapping = {query.id: query for query in queries}
+    queries_matched = []
+    for query_id in dashboard.query_ids:
+        query = query_mapping.get(query_id)
+        if query:
+            queries_matched.append(query)
+    return queries_matched
```
asnare, line 56: Maybe?
```suggestion
    return [query_mapping[query_id] for query_id in dashboard.query_ids if query_id in query_mapping]
```

asnare (changes requested):
Looks pretty good… a few small questions and so forth but nothing major.

src/databricks/labs/ucx/assessment/dashboards.py
JCZuurmond: Sorry, missed those, thanks for catching!

src/databricks/labs/ucx/assessment/dashboards.py
JCZuurmond: Sorry, missed those, thanks for catching!

src/databricks/labs/ucx/assessment/dashboards.py
JCZuurmond: #3440

JCZuurmond: And promoted log message to ERROR level

src/databricks/labs/ucx/source_code/queries.py
```diff
@@ -149,140 +154,84 @@ def _lint_queries(self, context: _ReportingContext) -> None:
...
             for problem in problems:
                 query_problems.append(
                     dataclasses.replace(
                         problem,
-                        dashboard_id=dashboard_id,
-                        dashboard_parent=dashboard_parent,
-                        dashboard_name=dashboard_name,
+                        dashboard_id=dashboard.id,
+                        dashboard_parent=dashboard.parent or "PARENT",
+                        dashboard_name=dashboard.name or "UNKNOWN",
```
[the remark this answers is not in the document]
JCZuurmond: yeah, the query problems are persisted in the inventory, thus not easy to change.

tests/integration/source_code/test_dashboards.py
```diff
@@ -0,0 +1,154 @@
+from datetime import datetime, timezone, timedelta
```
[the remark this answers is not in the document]
JCZuurmond: Changed it, I agree and also prefer `import datetime as dt`, though we are incosistent.

Similar to dataclass, I prefer `import dataclasses`, then `dataclasses.dataclass` and `dataclasses.field`

tests/unit/assessment/test_dashboards.py
JCZuurmond: Yes,  me being inconsistent :/ Thanks for reviewing!

asnare (approved):

src/databricks/labs/ucx/assessment/dashboards.py
```diff
@@ -0,0 +1,416 @@
...
+            dashboards.append(dashboard)
+        return dashboards
+
+    def _list_dashboards(self) -> list[SdkRedashDashboard]:
+        if self._include_dashboard_ids is not None:
+            return self._get_dashboards(*self._include_dashboard_ids)
+        try:
+            dashboards_iterator = self._ws.dashboards.list()
+        except DatabricksError as e:
+            logger.warning("Cannot list Redash dashboards", exc_info=e)
```
JCZuurmond, line 185:
```suggestion
            logger.error("Cannot list Redash dashboards", exc_info=e)
```

PR #3211, opened 2024-11-05
[DOC] README.md rewrite

JCZuurmond:
## Changes
Rewrite with changes coming out of preparing the presentation on UCX. Based reviews as README file: https://github.com/databrickslabs/ucx/tree/docs/odds-and-ends-when-reading-our-docs-to-prepare-a-presentation.

Notable changes:
- Separate `workflow` from `process` sections. Before, the table migration workflow section contained the commands to execute before and after running the table migration workflow and the workflow itself. Now, these are separated.
- Make the `table migration process` more prominent by moving it up. And more clear by restructuring the steps into functionally similar sections 
- Introduce `Installation resource` sections that detail the installation resources.
- Add a `dashboard` section

In general, make the wording and structure more similar, e.g. see how the workflow steps are enumerated.

PR #3398, opened 2024-12-02
Add note about running assessment once

JCZuurmond:
Got lost in https://github.com/databrickslabs/ucx/pull/3211

### DIFF
--- a/README.md
+++ b/README.md
@@ -764,6 +764,10 @@ See [this guide](docs/assessment.md) for more details.
 Proceed to the [group migration workflow](#group-migration-workflow) below or go back to the
 [migration process diagram](#migration-process).
 
+The UCX assessment workflow is intended to only run once; re-running it is not supported. If the inventory and findings
+for a workspace need to be updated then first reinstall UCX by [uninstalling](#uninstall-ucx)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -764,6 +764,10 @@ See [this guide](docs/assessment.md) for more details.
 Proceed to the [group migration workflow](#group-migration-workflow) below or go back to the
 [migration process diagram](#migration-process).
 
+The UCX assessment workflow is intended to only run once; re-running it is not supported. If the inventory and findings
```
FastLee, line 767: We should say something like, rerunning it will not clear the results of the last run.
Users may need to run assessment in certain use cases.
The instruction should be.
a. Avoid
b. If need to, clear the table that needs to be refreshed.

FastLee (changes requested):
Let's reconsider the message

README.md
JCZuurmond: I have updated the docs with a part of your comment. 

The advice (as discussed with @nfx) is to reinstall ucx to clear the tables, as stated in the change

FastLee: KK

FastLee (approved):

Issue #392, opened 2023-10-06
Replace Usage of Mount Points in the Workspace and Repo


dipankarkush-db:
Related to https://github.com/databrickslabs/ucx/issues/391.

Once you find the Usage of Mount Points in the Workspace and Repo we need to have a way to replace those mount points referred in the code with cloud storage, volumes, etc.

nfx:
cc @zpappa

pohlposition:
depends upon mount-point mapping state

pritishpai:
Will work on directly rewriting known locations (dfsa locations for which the tables are known without the mapping). Will revisit this in the future.

PR #3427, opened 2024-12-10
Add cli command to create directfs to table mapping

pritishpai:
## Changes
The directfs access in the code that can be replaced by table would need a mapping to be reviewed/updated by the user. Introduces a cli command that creates the mapping file from the directfs usages.

### Linked issues
Introduces #392

### Functionality

- [x] added new CLI command

### Tests

- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -197,6 +197,20 @@ def create_table_mapping(
         if len(workspace_contexts) == 1:
             webbrowser.open(f"{w.config.host}/#workspace{path}")
 
+@ucx.command
+def create_directfs_mapping(
+    w: WorkspaceClient,
+    ctx: WorkspaceContext | None = None,
+    run_as_collection: bool = False,
+    a: AccountClient | None = None,
+):
+    """Create DirectFS mapping for all the direcfs references in the workspace"""
+    workspace_contexts = _get_workspace_contexts(w, a, run_as_collection)
+
+    if ctx:
+        workspace_contexts = [ctx]
+    for workspace_context in workspace_contexts:
+        workspace_context.directfs_mapping.create_directfs_mapping()
 
 @ucx.command
 def validate_external_locations(
--- a/src/databricks/labs/ucx/contexts/application.py
+++ b/src/databricks/labs/ucx/contexts/application.py
@@ -457,6 +457,10 @@ class GlobalContext(abc.ABC):
         return TableMapping(self.installation, self.workspace_client, self.sql_backend)
 
     @cached_property
+    def directfs_mapping(self) -> DirectFsMapping:
+        return DirectFsMapping(self.installation, self.workspace_client, self.sql_backend)
+
+    @cached_property
     def catalog_schema(self) -> CatalogSchema:
         return CatalogSchema(self.workspace_client, self.table_mapping, self.migrate_grants, self.config.ucx_catalog)
 


src/databricks/labs/ucx/hive_metastore/directfs_mapping.py
```diff
@@ -0,0 +1,114 @@
...
+        directfs_crawler: DirectFsAccessCrawler,
+        tables_crawler: TablesCrawler,
+        workspace_name: str,
+        catalog_name: str,
+    ) -> Iterable["DirectFsRule"]:
+        """
+        List all direct filesystem access records.
+        """
+        directfs_snapshot = []
+        directfs_crawler.dump_all(directfs_snapshot)
```
JCZuurmond, line 78: Why do you dump an empty snapshot here?

src/databricks/labs/ucx/hive_metastore/directfs_mapping.py
```diff
@@ -0,0 +1,114 @@
...
+        tables_snapshot = list(tables_crawler.snapshot())
+        if not tables_snapshot:
+            msg = "No tables found. Please run: databricks labs ucx ensure-assessment-run"
+            raise ValueError(msg)
+        if not directfs_snapshot:
+            msg = "No directfs references found in code"
+            raise ValueError(msg)
```
JCZuurmond, line 85: This is just a return

src/databricks/labs/ucx/hive_metastore/directfs_mapping.py
```diff
@@ -0,0 +1,114 @@
...
+
+        # TODO: very inefficient search, just for initial testing
+        for table in tables_snapshot:
+            for directfs_record in directfs_snapshot:
+                if table.location:
+                    if directfs_record.path in table.location:
```
JCZuurmond, line 91: What about dfsa's that do not have a match with tables? We want to include those too

JCZuurmond (commented):
The direction looks good. Just thinking again about the idea of merging this with the table_mapping instead of a separate mapping. Did we log our decision about this somewhere?

src/databricks/labs/ucx/hive_metastore/directfs_mapping.py
pritishpai: We wont be having a mapping with a table to replace them and ignore them right?

pritishpai:
> The direction looks good. Just thinking again about the idea of merging this with the table_mapping instead of a separate mapping. Did we log our decision about this somewhere?

Table mapping is used entirely for table migration, while this is for code migration. I think we should keep this seperate.

src/databricks/labs/ucx/hive_metastore/directfs_mapping.py
pritishpai: I have changed this part of the code and no longer using dump_all. I had a wrong idea about its functionality.

src/databricks/labs/ucx/hive_metastore/directfs_mapping.py
pritishpai: If we do not find any directfs references then we just raise the error and return

JCZuurmond:
> Table mapping is used entirely for table migration, while this is for code migration. I think we should keep this separate. But no we have not logged the decision anywhere.

- The table mapping is maybe also used for code migration, or is that done with the able migration status based of the table properties?
- The DFSA mapping is not just for code migration, it is also for table migration to create the tables for the DFSAs

Issue #3438, opened 2024-12-12
[BUG]: Is UCX catalog now a pre-requisite for the table migration workflow to run?


JCZuurmond:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

In #2929, the UCX catalog became a prequisite for the `migrate-tables` command. In #3239, it becomes a prerequisite for the table migration workflows, and the migration progress is tracked during the workflow too.

### Expected Behavior

- Consistent prerequiste: both for the command and workflow. Or neither.
- If prerequisite, document it as such.

### Steps To Reproduce

N.A.

### Cloud

Azure

### Operating System

macOS

### Version

latest via Databricks CLI

asnare:
The `ucx` catalog is where progress (across multiple installations/workspaces) is tracked for displaying in the progress-related dashboards. From a technical perspective it's intentional that the migration-related pieces therefore depend on it so that they can log their progress for it to be reported. From a technical point-of-view it would therefore be consistent for both the command and workflow to require it.

Stepping back a bit, the question in my mind is where should the `ucx` catalog be created. As far as I can tell:

 - It only needs to be created once, even though multiple UCX installations might be using (sharing) it.
 - The way to create it is with the `create-ucx-catalog` command, but this isn't documented nor is when it should it happen.

With this in mind, I think the following is needed:

 - The `create-ucx-catalog` command should be documented.
 - Arguably during `attach-metastore` we should also attempt to create it if it's not already present. (This needs to be documented.) If we ever need to upgrade this is also an obvious point to check for this. In the alternative, if we decide _not_ to create the `ucx` catalog automatically here then if the catalog doesn't exist we should at least log a reminder that it needs to be created via `create-ucx-catalog`.
 - If the user attaches the metastore themselves, it needs to be clear from the docs that the `create-ucx-catalog` is still needed.

JCZuurmond:
Thanks for the good write up. 

IMO we should make the UCX catalog required **before** the migrate tables workflow.

> The way to create it is with the create-ucx-catalog command, but this isn't documented nor is when it should it happen.
> The create-ucx-catalog command should be documented.

Already is document: https://github.com/databrickslabs/ucx?tab=readme-ov-file#create-ucx-catalog-command. 
But **not** the when, I think it should be part of https://github.com/databrickslabs/ucx?tab=readme-ov-file#new-unity-catalog-resources

> Arguably during attach-metastore

[...]

PR #3439, opened 2024-12-12
Remove `try-except` around verifying the migration progress prerequisites in the `migrate-tables` cli command

JCZuurmond:
## Changes
Remove `try-except` around verifying the migration progress prerequistes in the `migrate-tables` cli command as the warning is more specific inside the class

### Linked issues

Progresses #3438

### Functionality

- [x] modified existing command: `databricks labs ucx migrate-tables`

### DIFF
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -687,16 +687,7 @@ def migrate_tables(
     for workspace_context in workspace_contexts:
         deployed_workflows = workspace_context.deployed_workflows
 
-        try:
-            workspace_context.verify_progress_tracking.verify()
-        except RuntimeWarning:
-            logger.warning(
-                "We couldn't detect a successful run of the assessment workflow."
-                "The assessment workflow is a prerequisite for the migrate-tables workflow."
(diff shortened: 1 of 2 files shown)


src/databricks/labs/ucx/cli.py
```diff
@@ -687,16 +687,7 @@ def migrate_tables(
     for workspace_context in workspace_contexts:
         deployed_workflows = workspace_context.deployed_workflows
 
-        try:
```
JCZuurmond, line 690: @FastLee : You added this in #2929. From the context of the PR and linked issue, I cannot determine why it is added in that PR. Do you remember why?

JCZuurmond (commented):
@asnare and @FastLee : See the linked issue. By tracking the migration progress during `migrate-tables` workflow, we make the creation of the UCX catalog a prerequisite. We should decide if this is desired or if we make it optional. 

If required, we should always verify the prerequisites in the workflow (as Andrew is implementing). Optionally, we can verify the prerequisites in the cli command to fail early.

gueniai (approved):

PR #3442, opened 2024-12-12
Remove cached property from `JobProgressEncoder`

### LINKED ISSUE #3377: [TECH DEBT] Avoid `cached_property` on `JobsProgressEncoder`
This shouldn't be part of a cached property because there's no way to reset it (or signal a refresh) and the encoder itself is a singleton.

Instead, look at how we do this in `TableEncoder`:

 - Override `.append_inventory_snapshot()`, so it loads the enrichment data.
 - Implement a new per-record encoder, for example `._encode_table_as_historical(record, extra, things, needed, for, enrichment)`
 - The new per-record encoder invokes `super().encode_record_as_historical(record)` and then performs enrichment.

This achieves the goals of only loading the data once (per snapshot encode) while not inadvertently keeping (potentially stale) data cached.

_Originally posted by @asnare in https://github.com/databrickslabs/ucx/pull/3375#discussion_r1856313515_

JCZuurmond:
## Changes
Remove cached property from JobProgressEncoder

### Linked issues

Resolves #3377

### DIFF
--- a/src/databricks/labs/ucx/progress/jobs.py
+++ b/src/databricks/labs/ucx/progress/jobs.py
@@ -1,14 +1,20 @@
 import collections
+import logging
+from collections.abc import Iterable
 from dataclasses import replace
-from functools import cached_property
 
 from databricks.labs.lsql.backends import SqlBackend
 
 from databricks.labs.ucx.assessment.jobs import JobInfo, JobOwnership
+from databricks.labs.ucx.framework.utils import escape_sql_identifier
 from databricks.labs.ucx.progress.history import ProgressEncoder
 from databricks.labs.ucx.progress.install import Historical
 from databricks.labs.ucx.source_code.directfs_access import DirectFsAccessCrawler
 from databricks.labs.ucx.source_code.jobs import JobProblem
+
+
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


src/databricks/labs/ucx/progress/jobs.py
```diff
@@ -37,8 +43,18 @@ def __init__(
         self._direct_fs_access_crawlers = direct_fs_access_crawlers
         self._inventory_database = inventory_database
 
-    @cached_property
-    def _job_problems(self) -> dict[int, list[str]]:
+    def append_inventory_snapshot(self, snapshot: Iterable[JobInfo]) -> None:
+        job_problems = self._get_job_problems()
+        dfsas = self._get_direct_filesystem_accesses()
```
FastLee, line 48: rename

FastLee, line 48:
```suggestion
        directfs_access = self._get_direct_filesystem_accesses()
```

src/databricks/labs/ucx/progress/jobs.py
```diff
@@ -37,8 +43,18 @@ def __init__(
...
+        history_records = []
+        for record in snapshot:
```
FastLee, line 50:
```suggestion
        for job_info in snapshot:
```

FastLee (approved):

PR #3469, opened 2024-12-23
Removed "packaging" package dependency

FastLee:
Removed dependency on "packaging" package to address release issue.

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/federation.py
+++ b/src/databricks/labs/ucx/hive_metastore/federation.py
@@ -4,7 +4,6 @@ import re
 from dataclasses import dataclass, replace
 from functools import cached_property
 from typing import ClassVar
-from packaging.version import Version, InvalidVersion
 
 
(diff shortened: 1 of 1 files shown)


pritishpai (approved):
LGTM!

KennyShaevelA8:
@FastLee @pritishpai is there a reason that running `databricks labs install ucx` seems to not pull down this latest commit of the file? 

I am getting error `ModuleNotFoundError: No module named 'packaging'` and inspecting the file that gets downloaded/unpacked from the zipball shows [the version from the previous commit](https://github.com/databrickslabs/ucx/blob/e768af21c7cf168c7c97bdc57165ccdba87a4221/src/databricks/labs/ucx/hive_metastore/federation.py).

pritishpai:
@KennyShaevelA8 databricks labs install ucx will pull the latest released version. For this change to be pulled, you will have to install the unreleased version with databricks labs install ucx@main

PR #3455, opened 2024-12-19
Implement disposition field in SQL backend

micheledaddetta-databricks:
### Description

This feature aims to the chance of running SQL statement for assessment results export in case of large workspaces with a large amount of findings.  
Introducing the `query_statement_disposition` value during the UCX installation can allow users to select the proper disposition method to use when running large queries. This parameter is added in the `config.yml` file and used for the SqlBackend definition. Using this approach, large queries will not fail.

### Changes
Added SQL query statement disposition choice during the UCX installation
Added `query_statement_disposition` value in the config file

### Linked issues

Resolves #3447

### Functionality

- [X] modified existing commands: `databricks labs install ucx`, `databricks labs ucx export-assessment`

### Tests

- [X] manually tested

PR #3477, opened 2025-01-06
 Implement disposition field in SQL backend

### LINKED ISSUE #3447: [FEATURE]: Add Disposition parameter in SqlBackend creation

### Problem statement

When running the `databricks labs ucx export-assessment`, some customer are facing the following issue:

> BadRequest: Inline byte limit exceeded. Statements executed with disposition=INLINE can have a result size of at most 26214400 bytes. Please execute the statement with disposition=EXTERNAL_LINKS if you want to download the full result.

This configuration can also introduce limits in other executions in which SQL statement results are not negligible. 

### Proposed Solution

Requiring the `Disposition` value during the UCX installation can allow users to select the proper disposition method to use when running large queries. This parameter can be added then in the `config.yml` file and used for the `SqlBackend` definition. Using this approach, large queries will not fail.

### Additional Context

I can work on it and introduce this feature if it seems compliant with UCX framework.

JCZuurmond:
### Description

This feature aims to the chance of running SQL statement for assessment results export in case of large workspaces with a large amount of findings.  
Introducing the `query_statement_disposition` value during the UCX installation can allow users to select the proper disposition method to use when running large queries. This parameter is added in the `config.yml` file and used for the SqlBackend definition. Using this approach, large queries will not fail.

### Changes
Added SQL query statement disposition choice during the UCX installation
Added `query_statement_disposition` value in the config file

### Linked issues

Resolves #3447
From other PR #3455

### Functionality

- [X] modified existing commands: `databricks labs install ucx`, `databricks labs ucx export-assessment`

### Tests

- [X] manually tested


src/databricks/labs/ucx/config.py
```diff
@@ -92,6 +93,9 @@ class WorkspaceConfig:  # pylint: disable=too-many-instance-attributes
     # Skip TACL migration during table migration
     skip_tacl_migration: bool = False
 
+    # Select SQL query statement disposition, default to INLINE
+    query_statement_disposition: Disposition | None = None
```
FastLee, line 97:
```suggestion
    query_statement_disposition: Disposition = Disposition.INLINE
```

src/databricks/labs/ucx/install.py
```diff
@@ -259,6 +260,12 @@ def _prompt_for_new_installation(self) -> WorkspaceConfig:
         recon_tolerance_percent = int(
             self.prompts.question("Reconciliation threshold, in percentage", default="5", valid_number=True)
         )
+
```
FastLee, line 263: I would reconsider. This may be confusing to the users.

FastLee (changes requested):

src/databricks/labs/ucx/install.py
JCZuurmond: With reconsider, you mean "make the config field a hidden parameter"? @micheledaddetta-databricks: Do you have input on this?

JCZuurmond: I prefer the hidden parameter for this parameter for now. Would like confirmation from @micheledaddetta-databricks and @gueniai

gueniai (approved):
lgtm

src/databricks/labs/ucx/install.py
JCZuurmond: Made the parameter hidden for now. Changing the prompting mechanism is out-of-scope for this PR

PR #3485, opened 2025-01-07
Schedule the migration progress workflow to run daily

asnare:
## Changes

This PR updates the UCX installation so that the migration progress workflow runs automatically once per day.

Changes include:

 - Refactoring some of the plumbing that we use for managing/installing workflows.
 - Allowing workflows to have a (Cron-based) schedule attached to them.
 - ~Allowing the default schedule to be overridden via configuration during installation.~ (Will be addressed in a new PR.)
 - Configuring the migration progress workflow to run daily, by default at 5 a.m. (UTC).

### Functionality

- updated relevant user documentation

### Tests

- added and existing unit tests
- added and existing integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -919,11 +919,11 @@ The output is processed and displayed in the migration dashboard using the in `r
 
 ## [EXPERIMENTAL] Migration Progress Workflow
 
-The manually triggered `migration-progress-experimental` workflow populates the tables visualized in the
+The `migration-progress-experimental` workflow populates the tables visualized in the
 [migration progress dashboard](#dashboards) by updating a **subset** of the [inventory tables](#assessment-workflow)
-to [track Unity Catalog compatability](docs/migration-progress.md) of Hive and workspace objects that need to be migrated.
+to [track Unity Catalog compatability](docs/migration-progress.md) of Hive and workspace objects that need to be migrated. It runs automatically once a day, but can also be triggered manually.
 
-The following pre-requisites need to be fulfilled before running the workflow:
+The following pre-requisites need to be fulfilled before the workflow will run successfully:
 - [UC metastore attached to workspace](../README.md#assign-metastore-command)
 - [UCX catalog exists](../README.md#create-ucx-catalog-command)
 - [Assessment job ran successfully](../README.md#ensure-assessment-run-command)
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -919,11 +919,11 @@ The output is processed and displayed in the migration dashboard using the in `r
 
 ## [EXPERIMENTAL] Migration Progress Workflow
```
JCZuurmond, line 920: It is not experimental anymore if we automate to run it daily

src/databricks/labs/ucx/runtime.py
```diff
@@ -62,8 +53,8 @@ def all(cls):
             ]
         )
 
-    def tasks(self) -> list[Task]:
```
JCZuurmond, line 65: Looks like a property, not a method. And redundant, make the property in the init not hidden

src/databricks/labs/ucx/installer/workflows.py
```diff
@@ -837,10 +844,11 @@ def _job_settings(self, step_name: str, remote_wheels: list[str]) -> dict[str, A
             date_to_remove = self._get_test_purge_time()
             tags.update({"RemoveAfter": date_to_remove})
         return {
-            "name": self._name(step_name),
+            "name": self._name(workflow_name),
             "tags": tags,
             "job_clusters": self._job_clusters(job_clusters),
             "email_notifications": email_notifications,
+            "schedule": workflow.schedule,
```
JCZuurmond, line 851: This works with `None`? Probably covered in directly by integration tests

JCZuurmond (approved):
Looks solid, some small pointers

README.md
asnare: I thought about this, but reasoned that we really don't know how well it works yet because I don't think many people (manually) trigger it. As such I thought it prudent to first see what kind of reports come in (if any) before removing the _experimental_ label.

Does that make sense at all?

src/databricks/labs/ucx/installer/workflows.py
asnare: Yes: this dictionary is eventually used as `**kwargs` and schedule is an argument that defaults to `None`.

PR #3488, opened 2025-01-07
Test if known list needs updating

JCZuurmond:
## Changes
Test if known list needs updating by using the caplog. This was interfering with other tests.


src/databricks/labs/ucx/source_code/known.json
```diff
@@ -32815,6 +32815,21 @@
     "trl.trainer.sft_trainer": [],
     "trl.trainer.utils": []
   },
+  "typeguard": {
```
JCZuurmond, line 32818: See this CI: https://github.com/databrickslabs/ucx/actions/runs/12651632798/job/35252615103

src/databricks/labs/ucx/source_code/known.json
```diff
@@ -253,6 +253,14 @@
     "audioread.rawread": [],
     "audioread.version": []
   },
+  "autocommand": {
```
JCZuurmond, line 256: See CI: See https://github.com/databrickslabs/ucx/actions/runs/12651705849/job/35252839633?pr=3488

src/databricks/labs/ucx/source_code/known.json
```diff
@@ -991,6 +999,12 @@
     "backcall._signatures": [],
     "backcall.backcall": []
   },
+  "backports.tarfile": {
```
JCZuurmond, line 1002: See: https://github.com/databrickslabs/ucx/actions/runs/12651766195/job/35253010483?pr=3488

src/databricks/labs/ucx/source_code/known.json
```diff
@@ -6625,6 +6639,11 @@
     "importlib_resources.readers": [],
     "importlib_resources.simple": []
   },
+  "inflect": {
```
JCZuurmond, line 6642: See: https://github.com/databrickslabs/ucx/actions/runs/12651940679/job/35253538201?pr=3488

src/databricks/labs/ucx/source_code/known.json
```diff
@@ -6819,12 +6838,23 @@
     "jaraco.classes.meta": [],
     "jaraco.classes.properties": []
   },
+  "jaraco.collections": {
```
JCZuurmond, line 6841: See https://github.com/databrickslabs/ucx/actions/runs/12652031785/job/35253819247?pr=3488

src/databricks/labs/ucx/source_code/known.json
```diff
@@ -33802,4 +33847,4 @@
     "zipp.glob": []
   },
   "zstd": {}
-}
```
asnare, line 33805: EOL at EOF.

tests/unit/source_code/test_known.py
```diff
@@ -73,11 +74,11 @@ def test_error_on_missing_known_json() -> None:
         KnownList._get_known()  # pylint: disable=protected-access
 
 
-@pytest.mark.skip(reason="Interferes with tests reading known list at the same time of rebuilding (writing)")
-def test_rebuild_trivial() -> None:
-    # No-op: the known.json file is already up-to-date
-    cwd = Path.cwd()
-    KnownList.rebuild(cwd)
```
asnare, line 80: I've been thinking about this, and was thinking along the lines of:
a) The `.rebuild()` method gets an extra "update_known_json" parameter that defaults to `False`.
b) The `main()` method that calls it passes in `True`.
c) The branch for updating the file raises an exception if the distributions are different and it's not allowed to write. This would be what the test encounters.

What do you think?

asnare (approved):
Looks like an improvement, although I was thinking about doing this slightly differently.

tests/unit/source_code/test_known.py
asnare: I just realised these changes don't fix the problem: by the time we fail _this_ test the `known.json` has already been rewritten, and other (concurrent) tests may also have failed because they read the file while it was being updated.

Given that a test shouldn't be modifying anything in the source tree, I think this needs to be addressed.

asnare (changes requested):
I just realised these updates don't really address the underlying problem.

That said, they do improve things somewhat:
 - This forces the list to be maintained.
 - If there are no changes, we won't see spurious failures of concurrent tests (because the file isn't being rewritten).
 - If this test does fail, the file will be rewritten. This means that: a) concurrent tests might also fail; b) the workspace is now dirty.

(Although they do make it more likely that we won't face the prob

asnare (approved):
lgtm

PR #3499, opened 2025-01-08
Add dummy id for jobs to include on `MockInstallationContext.config`

JCZuurmond:
## Changes
Add dummy id for jobs to include on `MockInstallationContext.config` to avoid assessing all workflows.

### Linked issues

A more proper solution would be to resolve: https://github.com/databrickslabs/blueprint/issues/179

### Tests

- [x] manually tested


asnare (approved):
Agree, this isn't ideal but glad to see the underlying cause is being addressed upstream in databrickslabs/blueprint#180 (after which this can be removed).

PR #3501, opened 2025-01-08
Fix issue in installing UCX on UC enabled workspace

### LINKED ISSUE #3420: [BUG]: Installing ucx on UC enabled cluster creates Single User UC cluster instead of the expected non UC cluster

### Current Behavior

The assessment job for UCX installs the standard cluster and tacl cluster. When UCX is installed on a workspace which is already UC-enabled, the workflows are created with an enabled cluster.

### Expected Behavior

There should be consistency in the cluster types created when UCX is installed whether UC is enabled on the workspace or not.

### Steps To Reproduce

_No response_

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

_No response_

HariGS-DB:
## Changes

This PR updates the UCX policy that is created when installing UCX. It replaces the policy definition for spark_version from fixed to allowlist with a default value.
When UC is enabled on a workspaces, the cluster definition take the value as single_user and user_isolation instead of the Legacy_Single_User and Legacy_Table_ACL and this is occurring due to the policy which is overriding this. Changing the sparkversion from fixed to allowlist solved this issue. This could be a API issue which needs to be raised separately, but for now it addresses the problem by doing this simple change

It also updates the job definition to use the default value when not passed by setting the apply_policy_default_values to true

### Linked issues

Resolves #3420 

### Functionality

### Tests

- [ ] updated unit tests
- [ ] updated integration tests
- [ ] static installation test


HariGS-DB:
Before the change Tacl Cluster configuration on UC enabled cluster
image

After the change Tacl Cluster configuration on UC enabled cluster
image

asnare (commented):
I can see tests that check policy settings and the like, but is there an integration test somewhere that actually creates the cluster and verifies it's not UC-enabled?

(We have `VerifyHasMetastore.verify_metastore()` that can check this, but I don't see it being called from any integration tests.)

HariGS-DB:
> I can see tests that check policy settings and the like, but is there an integration test somewhere that actually creates the cluster and verifies it's not UC-enabled?
> 
> (We have `VerifyHasMetastore.verify_metastore()` that can check this, but I don't see it being called from any integration tests.)

I have done this testing manually by installing the UCX from the branch and verifying the cluster properties on a UC enabled workspace. But we don't have a int test at the moment, and I can write one . Good point

HariGS-DB:
@asnare added integration test to check the cluster data security mode when the workspace is UC enabled

tests/integration/install/test_installation.py
```diff
@@ -128,6 +128,28 @@ def test_job_cluster_policy(ws, installation_ctx) -> None:
...
+def test_job_cluster_on_uc_enabled_workpace(ws, installation_ctx) -> None:
+    installation_ctx.workspace_installation.config.override_clusters = ""
+
+    installation_ctx.workspace_installation.run()
+    job_id = installation_ctx.install_state.jobs["assessment"]
+    if installation_ctx.verify_has_metastore.verify_metastore():
+        job_clusters = installation_ctx.workspace_client.jobs.get(job_id).settings.job_clusters
+        for cluster in job_clusters:
+            if cluster.job_cluster_key == "main":
+                assert cluster.new_cluster.data_security_mode == compute.DataSecurityMode.LEGACY_SINGLE_USER_STANDARD
```
JCZuurmond, line 141: The security modes should be legacy when UC is enabled because before the concept of legacy did not exists as legacy signals pre-UC?

JCZuurmond (commented):
@HariGS-DB : Did you manually test this on our UC enabled workspace?

tests/integration/install/test_installation.py
HariGS-DB: Yes the security mode for tacl cluster is LEGACY_TABLE_ACL for main its LEGACY_SINGLE_USER_STANDARD and for user_isolation(UC cluster) is USER_ISOLATION. the legacy security mode names will be the same whether its UC enabled or not.

I have written this test case which runs on our UC enabled int env and checks the presence of right cluster configuration.

tests/integration/install/test_installation.py
```diff
@@ -128,6 +128,28 @@ def test_job_cluster_policy(ws, installation_ctx) -> None:
         assert policy_definition["aws_attributes.availability"]["value"] == compute.AwsAvailability.ON_DEMAND.value
 
 
+@retried(on=[NotFound, InvalidParameterValue], timeout=timedelta(minutes=3))
+def test_job_cluster_on_uc_enabled_workpace(ws, installation_ctx) -> None:
+    installation_ctx.workspace_installation.config.override_clusters = ""
```
JCZuurmond, line 133: Add a why-comment on why this is here

tests/integration/install/test_installation.py
JCZuurmond: If the mode names are the same whether its UC enabled or not, why is the `installation_ctx.verify_has_metastore.verify_metastore()` required?

JCZuurmond (approved):
LGTM, one nit and one clarification question to improve my understanding. Approving so you can go ahead

tests/integration/install/test_installation.py
HariGS-DB: good point, removed this condition

tests/integration/install/test_installation.py
HariGS-DB: added

PR #3546, opened 2025-01-17
Add assert to make sure installation is finished before re-installation

pritishpai:
## Changes
The reinstallation is starting before the first installation is done. This is causing the warning to not be raised and the test fails

### Tests
- [x] modified integration tests

### DIFF
--- a/tests/integration/install/test_installation.py
+++ b/tests/integration/install/test_installation.py
@@ -393,7 +393,9 @@ def test_check_inventory_database_exists(ws, installation_ctx):
 
 
 def test_compare_remote_local_install_versions(ws, installation_ctx):
-    installation_ctx.workspace_installation.run()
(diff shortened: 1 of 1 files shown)


pritishpai:
Will need to add a small sleep if this doesn't work.

pritishpai:
Can be merged after this passes: https://github.com/databrickslabs/ucx/actions/runs/12834034528

JCZuurmond (dismissed):
@pritishpai : I do not think the assert makes a differences, however, based on your findings from the `logs` we could add a timed out check that waits for the configuration file to be uploaded as that is missing on the second install. I prefer such a timed out check over a sleep. Could you make those changes?

JCZuurmond (approved):

gueniai (approved):
LGTM

Issue #3107, opened 2024-10-30
[FEATURE]: Add a cli command to migrate pipelines


pritishpai:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Migrate Pipelines cli command for users to migrate their DLT pipelines

### Proposed Solution

Add a cli command to migrate pipelines

PR #3579, opened 2025-01-28
Added the CLI command for migrating DLT pipelines

pritishpai:
## Changes
Added cli command for migration DLT pipelines with the flags for limiting the pipelines ids

### Linked issues
Introduces #3107 

### Functionality
- [x] added new CLI command

### Tests
- [x] manually tested
- [ ] added unit tests
- [ ] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/cli.py
+++ b/src/databricks/labs/ucx/cli.py
@@ -913,7 +913,12 @@ def migrate_dlt_pipelines(
     a: AccountClient | None = None,
     **named_parameters,
 ) -> None:
-    """Migrate DLT pipelines to UC"""
+    """
+        Migrate DLT pipelines to UC that were crawled in the assessment workflow
+        Flags
+        -- include-pipeline-ids: Comma separated list of pipeline ids to migrate
+        -- exclude-pipeline-ids: Comma separated list of pipeline ids to exclude from migration
+    """
     if ctx:
         workspace_contexts = [ctx]
     else:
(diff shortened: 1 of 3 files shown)


src/databricks/labs/ucx/cli.py
```diff
@@ -911,12 +911,18 @@ def migrate_dlt_pipelines(
     ctx: WorkspaceContext | None = None,
     run_as_collection: bool = False,
     a: AccountClient | None = None,
+    **named_parameters,
```
JCZuurmond, line 914: can't we make the flags explicit?

src/databricks/labs/ucx/contexts/application.py
```diff
@@ -374,8 +374,23 @@ def table_ownership_grant_loader(self) -> TableOwnershipGrantLoader:
 
     @cached_property
     def pipelines_migrator(self) -> PipelinesMigrator:
+        include_pipeline_ids = (
+            self.named_parameters.get('include_pipeline_ids', '').split(',')
+            if 'include_pipeline_ids' in self.named_parameters
+            else None
+        )
```
JCZuurmond, line 381: Kinda resolved already with the `get`

```suggestion
        include_pipeline_ids = self.named_parameters.get('include_pipeline_ids', '').split(',') or None
```

Same for `exclude_pipeline_ids` below

JCZuurmond (approved):
Minor details, approving so you can move forward

src/databricks/labs/ucx/contexts/application.py
pritishpai: I had this code before it resolves to an  empty list [''] instead of None so had to add an explicit else for None.

src/databricks/labs/ucx/cli.py
pritishpai: It aligns with all the other cli command methods (that have flags) and also the _get_workspace_contexts expects it.

PR #3520, opened 2025-01-14
Fix cli command to migrate local folder

JCZuurmond:
## Changes
Fix cli command to migrate local folder:
- Renamed `LocalFileMigrator` to `LocalCodeMigrator` to signal it migrates not just files
- Rename d`Fixer.name` to `Fixer.diagnostic_code` and make all fixer diagnostic code unique for matching when fixing code
- Cleaned some code, like reusing `apply_fixes`, prefering `Path.{read|write}_text`, forcing keyword argument if positional is unclear, matching test modules with ucx modules

### Linked issues

Resolves #3514

Stacked on:
- [x] #3535

### Functionality

- [x] modified existing command: `databricks labs ucx migrate-local-code`

### Tests

- [ ] manually tested
- [x] modified and added unit tests
- [x] modified and added integration tests

Issue #3514, opened 2025-01-13
[FEATURE] Add CLI command to migrate a folder of Python files and notebooks on local file system


JCZuurmond:
# Description

Add CLI command to migrate a folder of files/notebooks on local file system

PR #3586, opened 2025-01-29
Let `LocalFileLinter` reuse the `Dependency` instead falling back on its path

JCZuurmond:
## Changes

Let `LocalFileLinter` reuse the `Dependency` instead falling back on its path as we introduce duplicate ways of loading files. 

In addition to that, the PR contains a clean up of:
- Moving non-linting file classes from `source_code/linters/files.py` to `source_code/files.py` to avoid circular imports
- Extend `MockPathLookup` to verify which paths are resolved
- Make `current_session` public on `LinterContext` for reuse instead of passing it next to the `LinterContext`
- Improve testing and handling corner cases for `FileLoader`

### Linked issues

Progresses #3514
Breaks up #3520

### Functionality

- [x] modified existing command: `databricks labs ucx migrate-local-code`

### Tests

- [ ] manually tested
- [x] modified and added unit tests
- [x] modified and added integration tests


src/databricks/labs/ucx/source_code/files.py
```diff
@@ -0,0 +1,98 @@
...
+        if self.language == Language.PYTHON:
+            context = parent.new_dependency_graph_context()
+            analyzer = PythonCodeAnalyzer(context, self._original_code)
+            problems = analyzer.build_graph()
+            for idx, problem in enumerate(problems):
+                if problem.has_missing_path():
+                    problems[idx] = dataclasses.replace(problem, source_path=self._path)
+            return problems
+        # supported language that does not generate dependencies
+        if self.language == Language.SQL:
```
pritishpai, line 47: Can we combine this and the other return as SQL is also unsupported and probably doesn't need an if statement

src/databricks/labs/ucx/source_code/linters/context.py
```diff
@@ -72,14 +74,14 @@ def __init__(
...
         python_linters += [
-            DirectFsAccessPyLinter(session_state),
-            DBRv8d0PyLinter(dbr_version=session_state.dbr_version),
-            SparkConnectPyLinter(session_state),
-            DbutilsPyLinter(session_state),
+            DirectFsAccessPyLinter(self.session_state),
+            DBRv8d0PyLinter(dbr_version=self.session_state.dbr_version),
+            SparkConnectPyLinter(self.session_state),
+            DbutilsPyLinter(self.session_state),
             DirectFsAccessSqlPylinter(sql_direct_fs),
```
pritishpai, line 81: Should the sql_direct_fs also be a part of LinterContext object?

pritishpai (approved):
Commented a few nits. Approving so that you can merge in the AM.

src/databricks/labs/ucx/source_code/linters/context.py
JCZuurmond: All linters should be part of the linter context, see docstring added in https://github.com/databrickslabs/ucx/pull/3638/files

src/databricks/labs/ucx/source_code/files.py
JCZuurmond: I prefer to keep it like this because SQL is supported, it just does not return a dependency graph, see #3639

JCZuurmond:
Surpassed by #3641

PR #3588, opened 2025-01-30
Extend code migration progress documentation

### LINKED ISSUE #2231: [DOCS]: Document switching away from `hive_metastore` as the default workspace catalog

### Problem statement

When an existing workspace is UC-enabled, by the [default catalog is (by default) `hive_metastore`](https://docs.databricks.com/en/catalogs/default.html#the-default-catalog-configuration-when-unity-catalog-is-enabled).

### Proposed Solution

As discussed in discussed in #102, we wish to switch workspaces away from using the `hive_metastore` by default. Changing the workspace is an all-or-nothing change, and existing queries/jobs/etc. may break. As such:

 - There needs to be a workflow that can change the default catalog. (Can this be anything other than `hive_metastore`? If so a backup is required.)
 - A command needs to be implemented to allow for rollback.

### Additional Context

Implementation prerequisite:

 - UCX itself assumes it has been installed into the default catalog; if the default catalog is changed then currently UCX will break. This needs to be fixed. See: #2207

Runtime prerequisite:

 - A metastore catalog exists that can be made the default instead of `hive_metastore`.

JCZuurmond:
## Changes
Extend code migration progress documentation by detailing:
- Start with linting then do the migration
- Use dashboards
- Use local commands
- Reference default catalog resource
- Add section for finalising the migration

### Linked issues

Resolves #2231

### Functionality

### DIFF
--- a/docs/ucx/docs/process/index.mdx
+++ b/docs/ucx/docs/process/index.mdx
@@ -9,7 +9,8 @@ On a high level, the steps in migration process are:
 2. [group migration](/docs/reference/workflows#group-migration-workflow)
 3. [table migration](/docs/process/#table-migration-process)
 4. [data reconciliation](/docs/reference/workflows#post-migration-data-reconciliation-workflow)
-5. [code migration](/docs/reference/commands#code-migration-commands)
+5. [code migration](#code-migration)
+6. [final details](#final-details)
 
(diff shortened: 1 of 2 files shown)


FastLee (approved):
LGTM

PR #3589, opened 2025-01-30
Remove the linting false positive for missing table format warning when using `spark.table`

### LINKED ISSUE #3545: [BUG]: Eliminate linting false positive for `default-format-changed-in-dbr8` when reading a spark table

### Current Behavior

Eliminate linting false positive for `default-format-changed-in-dbr8` when reading a spark table

### Expected Behavior

No warning about this as the format does not matter when reading a table only on writing

### Steps To Reproduce

_No response_

### Cloud

Azure

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

```shell

```

JCZuurmond:
## Changes
Remove the linting false positive for missing table format warning when using `spark.table`

### Linked issues
Resolves #3545

### Functionality

- [x] modified liniting related logic

### Tests

- [x] modified unit tests

### DIFF
--- a/src/databricks/labs/ucx/source_code/linters/table_creation.py
+++ b/src/databricks/labs/ucx/source_code/linters/table_creation.py
@@ -108,8 +108,10 @@ class DBRv8d0PyLinter(PythonLinter):
 
         self._linter = NoFormatPythonLinter(
             [
+                # https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.writeTo.html
                 NoFormatPythonMatcher("writeTo", 1, 1),
                 NoFormatPythonMatcher("table", 1, 1),
+                # https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameWriter.saveAsTable.html
                 NoFormatPythonMatcher("saveAsTable", 1, 4, 2, "format"),
             ]
         )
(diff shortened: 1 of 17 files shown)


src/databricks/labs/ucx/source_code/linters/table_creation.py
```diff
@@ -106,10 +107,13 @@ def __init__(self, dbr_version: tuple[int, int] | None):
         version_cutoff = (8, 0)
         self._skip_dbr = dbr_version is not None and dbr_version >= version_cutoff
 
+        # A more precise match would check if the method names come from their respective parent classes. However, given
+        # the (current) uniqueness of the names within the Spark module it is not required (yet).
         self._linter = NoFormatPythonLinter(
             [
+                # https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.writeTo.html
                 NoFormatPythonMatcher("writeTo", 1, 1),
-                NoFormatPythonMatcher("table", 1, 1),
```
JCZuurmond, line 112: This was the "trick". Also see new comment on line 110

pritishpai (dismissed):
LGTM!

src/databricks/labs/ucx/source_code/linters/table_creation.py
pritishpai: For reading the table format does not matter, but will this raise an advice for saveAsTable without an explicit format parameter? I see it as optional. Also writeTo does not have a format parameter as per
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.writeTo.html

JCZuurmond: There will be **no** advice if `format` is defined when writing a table. There will be when it is missing and the runtime is unknown or lower than the cutoff

JCZuurmond: There will be **no** advice if `format` is defined when writing a table. There will be when it is missing and the runtime is unknown or lower than the cutoff

pritishpai (approved):

Issue #3585, opened 2025-01-29
[BUG]: Linter error message doesn't align with the source code


prashsub:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

<img width="764" alt="Image" src="https://github.com/user-attachments/assets/5ef9f6d4-dda4-4093-80c1-9f908f5993f5" />

We are were trying to analyze the results as part of the assessment, and found the issues as shown in the attached picture. We couldn't find references in the source code to the errors highlighted by the linter. We couldn't find any references to sc, sqlcontext, sparkcontext or _jvm. etc.

1. legacy-context-in-shared-cluster - sc is not supported on UC Shared Clusters. Rewrite it using spark
2. jvm-access-in-shared-clusters - Cannot access Spark Driver JVM on UC Shared Clusters

### Expected Behavior

These notebooks should not have been highlighted for this issue by the linter. If the issue does actually exist, it would nice to know the actual line number. The line and column numbers don't seem to match the source code right now.

### Steps To Reproduce

The source code is attached below. 

``` python
# Databricks notebook source
import os
import datetime
import time
import logging
import re
import pandas as pd
import pymongo
import pyspark
from pyspark.sql.functions import *
from datetime import datetime, timedelta, date, timezone
from pyspark.sql import *
from pyspark.sql.types import *
from pyspark.sql.functions import max
from os import path
from pytz import timezone
from pymongo import MongoClient
# email_sender = email_receiver = email_sender_name = 'xxx'
email_sender = email_receiver = email_sender_name = 'xxx'

# COMMAND ----------

# current date and time
now = datetime.now()

#Init Job Monitor Vars
job_id = 1
job_start_time = now.strftime("%Y-%m-%d %H:%M:%S.%f")
job_rows = 0
job_name = 'DBKS_DATALAKE_TO_MONGO_PUSH'

# COMMAND ----------

# MAGIC %run /UTILITIES/COMMON_UTILS

# COMMAND ----------

dbutils.widgets.text("MONGO_PORT","xxx")
dbutils.widgets.text("dbname","xxx")
dbutils.widgets.text("collection","xxx")
data_logpath=dbutils.widgets.get("logpath")

# COMMAND ----------

central = timezone('US/Central')
current_time = datetime.now() 
current_date = current_time.astimezone(central)
log_folder_path =  data_logpath + str(current_date.year) + '/' + str(current_date.month) + '/'
if not os.path.exists(log_folder_path):
[...]

PR #3643, opened 2025-02-05
[REFACTOR] Make `allow_list` / `KnownList` optional

JCZuurmond:
## Changes
Make `allow_list` / `KnownList` optional so that it does not need to passed around all the time. Only necessarily for testing as `KnownList` is an artifact in the UCX code base, hence static (and not context aware) 

### Linked issues

Progresses #3585


JCZuurmond:
Not really needed anymore as I am signaling a dependency is known with the `KnownDependency` in #3655, but still can be merged if we think this makes sense

pritishpai (approved):

PR #3645, opened 2025-02-05
[TECH DEBT] Remove walked paths parameter from dependency graph walker

JCZuurmond:
## Changes
Remove walked paths from dependency graph walker as it should not be used, the developer should relay on the `MockPathLookup.successfully_resolved_paths` as it is important that the paths are resolved not just added to the cache in the walker

### DIFF
--- a/src/databricks/labs/ucx/source_code/linters/graph_walkers.py
+++ b/src/databricks/labs/ucx/source_code/linters/graph_walkers.py
@@ -99,6 +99,7 @@ class DependencyGraphWalker(abc.ABC, Generic[T]):
 
     @property
     def lineage(self) -> list[LineageAtom]:
+        """The lineage for getting to the dependency."""
         lists: list[list[LineageAtom]] = [dependency.lineage for dependency in self._lineage]
         return list(itertools.chain(*lists))
 
--- a/src/databricks/labs/ucx/source_code/linters/graph_walkers.py
+++ b/src/databricks/labs/ucx/source_code/linters/graph_walkers.py
@@ -37,11 +37,12 @@ T = TypeVar("T")
 
 class DependencyGraphWalker(abc.ABC, Generic[T]):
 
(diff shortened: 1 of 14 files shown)


tests/integration/source_code/solacc.py
```diff
@@ -175,7 +175,7 @@ def _lint_dir(solacc: _SolaccContext, soldir: Path):
     linted_files = set(files_to_skip)
     # lint solution
     start_timestamp = datetime.now(timezone.utc)
-    advices = list(ctx.local_code_linter.lint_path(soldir, linted_files))
+    advices = list(ctx.local_code_linter.lint_path(soldir))
```
JCZuurmond, line 178: This needs to be addressed. An option would be to extend the `FolderLoader` with an `exclude_paths` parameter, then pass the `files_to_skip` to the `FolderLoader`. It would also be a speed up as the files while not be read. We could als use such a field in the `WorkspaceConfig` to exclude certain folders from being processed

src/databricks/labs/ucx/source_code/linters/graph_walkers.py
```diff
@@ -36,22 +36,37 @@
...
             # the dependency is a root, so its path is the one to use
             # for computing lineage and building python global context
-            root_path = dependency.path
-            yield from self._iter_one(dependency, self._graph, root_path)
+            yield from self._iter_one(dependency, self._graph, dependency.path)
 
     def _iter_one(self, dependency: Dependency, graph: DependencyGraph, root_path: Path) -> Iterable[T]:
+        """Iterate over a single dependency going depth first."""
         if dependency.path in self._walked_paths:
+            # TODO:
```
JCZuurmond: This is an open question to me, @pritishpai and @asnare , please have a look.

I am inclined to leave the question open but not remove the logic to be on the safer side until we answered the open question. 

Note that removing this logic does **NOT** make any unit test fail

pritishpai (approved):

PR #3626, opened 2025-02-03
Surface known problems during linting

JCZuurmond:
## Changes
Surface known problems during linting

### Linked issues

Resolves #3585

PR #2502, opened 2024-08-28
Implement dependency graph walker

ericvergnaud:
## Changes
There is redundant code for walking a dependency graph 'properly' i.e. from roots to leaves, populating python globals etc...
This PR makes it trivial to reuse the correct code, and reuses it for local code linting and jobs linting.
This will also be used by #2371 

### Linked issues
None

### Functionality
None

### Tests
- [x] ran existing unit tests

Issue #1527, opened 2024-04-24
[FEATURE]: Convert 'deprecated package' from dependency resolution checks to file linting ones


ericvergnaud:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Problem statement

Raising advices for deprecated packages is currently implemented as part of dependency graph building.
This is incorrect and should be moved to a regular linter.

### Proposed Solution

Create the required linter

PR #3654, opened 2025-02-06
Remove redundant `if`-condition in the `GraphWalker._iter_one`

JCZuurmond:
## Changes
Remove redundant `if`-condition in the `GraphWalker._iter_one`

### Linked issues

Progresses #3626

### Functionality

- [x] Code linting

### DIFF
--- a/src/databricks/labs/ucx/source_code/linters/graph_walkers.py
+++ b/src/databricks/labs/ucx/source_code/linters/graph_walkers.py
@@ -56,16 +56,15 @@ class DependencyGraphWalker(abc.ABC, Generic[T]):
         self._lineage.append(dependency)
         self._walked_paths.add(dependency.path)
         self._log_walk_one(dependency)
-        if dependency.path.is_file() or is_a_notebook(dependency.path):
-            inherited_tree = graph.root.build_inherited_tree(root_path, dependency.path)
-            path_lookup = self._path_lookup.change_directory(dependency.path.parent)
-            yield from self._process_dependency(dependency, path_lookup, inherited_tree)
-            maybe_graph = graph.locate_dependency(dependency.path)
-            # missing graph problems have already been reported while building the graph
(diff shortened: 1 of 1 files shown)


src/databricks/labs/ucx/source_code/linters/graph_walkers.py
```diff
@@ -56,16 +56,15 @@ def _iter_one(self, dependency: Dependency, graph: DependencyGraph, root_path: P
         self._lineage.append(dependency)
         self._walked_paths.add(dependency.path)
         self._log_walk_one(dependency)
-        if dependency.path.is_file() or is_a_notebook(dependency.path):
```
JCZuurmond, line 59: @pritishpai and @asnare : This is [another `if`-condition](https://github.com/databrickslabs/ucx/pull/3645#discussion_r1944716842) in this method about which I am unsure. However, I want to remove it as it is interfering with [surfacing known problems during linting](#1527).

Notes:
- Removing the `if`-condition does not let any unit test fail
- The condition was [introduced](https://github.com/databrickslabs/ucx/pull/2502/files) without notes, explanation or questions
- IMO the base graph walker class should not decide to skip a dependency. Classes that inherit form this class can decide to skip specific dependencies
- I introduced a new dependency to surface known problems, `KnownDependency`, which is not processed as the condition is not true for that dependency. Instead of extending the condition, I think it should be removed all together

pritishpai (approved):

src/databricks/labs/ucx/source_code/linters/graph_walkers.py
JCZuurmond: @ericvergnaud : Could you shed some light on this `if-`condition?

JCZuurmond: The inverse of the condition is not reached by the unit test code base

PR #3657, opened 2025-02-06
Add test to check if include_job_ids ins enforced while crawling jobs

pritishpai:
## Changes
The jobs crawler is not enforcing the include_jobs_ids property and is crawling all jobs in the workspace. Change makes sure that the config is enforced

### Linked issues

Resolves #3656

### Tests

- [x] added integration tests

PR #3658, opened 2025-02-06
Scope crawled jobs in `JobsCrawler` with `include_job_ids`

### LINKED ISSUE #3656: [BUG]: assess_jobs not taking include_job_ids into consideration

### Current Behavior

assess_jobs not taking include_job_ids into consideration

Image

### Expected Behavior

assess_jobs should only run for specific job_ids if it is present in config.yml

### Steps To Reproduce

1. Add specific job ids under include_job_ids in config.yml
2. Run assessment

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

```shell

```

mohanab-db:
## Changes
Check for include_job_ids in assess_jobs task

### Linked issues
https://github.com/databrickslabs/ucx/issues/3656

Resolves #3656 

### Functionality

- [x] modified existing workflow: jobs_crawler

### Tests

- [x] added integration tests

### DIFF
--- a/src/databricks/labs/ucx/assessment/jobs.py
+++ b/src/databricks/labs/ucx/assessment/jobs.py
@@ -94,7 +94,9 @@ class JobsMixin:
 
 
 class JobsCrawler(CrawlerBase[JobInfo], JobsMixin, CheckClusterMixin):
-    def __init__(self, ws: WorkspaceClient, sql_backend: SqlBackend, schema, include_job_ids: list[int] | None = None):
+    def __init__(
+        self, ws: WorkspaceClient, sql_backend: SqlBackend, schema, *, include_job_ids: list[int] | None = None
+    ):
         super().__init__(sql_backend, "hive_metastore", schema, "jobs", JobInfo)
         self._ws = ws
         self._include_job_ids = include_job_ids
--- a/src/databricks/labs/ucx/assessment/jobs.py
+++ b/src/databricks/labs/ucx/assessment/jobs.py
@@ -94,9 +94,10 @@ class JobsMixin:
 
 
 class JobsCrawler(CrawlerBase[JobInfo], JobsMixin, CheckClusterMixin):
-    def __init__(self, ws: WorkspaceClient, sql_backend: SqlBackend, schema):
+    def __init__(self, ws: WorkspaceClient, sql_backend: SqlBackend, schema, include_job_ids: list[int] | None = None):
         super().__init__(sql_backend, "hive_metastore", schema, "jobs", JobInfo)
         self._ws = ws
+        self._include_job_ids = include_job_ids
 
     def _crawl(self) -> Iterable[JobInfo]:
         all_jobs = list(self._ws.jobs.list(expand_tasks=True))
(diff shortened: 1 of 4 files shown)


pritishpai:
Needs an integration test: https://github.com/databrickslabs/ucx/pull/3657/files#diff-1dca10e5526032a0b818ce12b0a73066355672aa765216e8a3caf9e0e099c5aa

src/databricks/labs/ucx/assessment/jobs.py
```diff
@@ -94,9 +94,10 @@ def _job_clusters(job: BaseJob) -> Iterable[tuple[BaseJob, ClusterSpec]]:
 
 
 class JobsCrawler(CrawlerBase[JobInfo], JobsMixin, CheckClusterMixin):
-    def __init__(self, ws: WorkspaceClient, sql_backend: SqlBackend, schema):
+    def __init__(self, ws: WorkspaceClient, sql_backend: SqlBackend, schema, include_job_ids: list[int] | None = None):
```
JCZuurmond, line 97:
```suggestion
    def __init__(self, ws: WorkspaceClient, sql_backend: SqlBackend, schema, *, include_job_ids: list[int] | None = None):
```

tests/integration/assessment/test_jobs.py
```diff
@@ -15,14 +15,19 @@
 @retried(on=[NotFound], timeout=timedelta(minutes=5))
 def test_job_crawler(ws, make_job, inventory_schema, sql_backend):
     new_job = make_job(spark_conf=_SPARK_CONF)
-    job_crawler = JobsCrawler(ws=ws, sql_backend=sql_backend, schema=inventory_schema)
+    skip_job = make_job(spark_conf=_SPARK_CONF)
+
+    ws.config.include_job_ids = [new_job.job_id]
```
JCZuurmond, line 20: Why add it to the config if it is passed as parameter below?

JCZuurmond (commented):
@mohanab-db : Thank you for reporting this issue and resolving it!

I have added some minor comments. A more fundamental implementation question would be: what is the bottle neck when crawling all the jobs?

Is the bottleneck:
- Listing all the jobs
- Listing all clusters
- Combining jobs with clusters
- Assessing the jobs (actually we only assess the clusters related to the jobs)

JCZuurmond:
@pritishpai : Could you update the `WorkflowLinter` to use the `JobsCrawler` instead of crawling the jobs itself? Can be a `tech debt` issue separate from this PR. See the `QueryLinter` as an example

pritishpai:
> @pritishpai : Could you update the WorkflowLinter to use the JobsCrawler instead of crawling the jobs itself? Can be a tech debt issue separate from this PR. See the QueryLinter as an example

https://github.com/databrickslabs/ucx/issues/3662

JCZuurmond (approved):

FastLee:
I would vote against this change.

JCZuurmond:
> I would vote against this change.

@FastLee : Could you add your motivation?

I am **pro** this change as it parallel with how we do this with other `include_` config attributes that scope the resources, which makes the code easier to follow as it is according to expectations (of (new) developers and users like @mohanab-db). 

This outweighs IMO restricting `include_job_ids` to be used in the `WorkflowLinter` only, which is a more straightforward implementation in the absolute sense, but relative to the code base it is deviating from expectations. (The `WorkflowLinter` class has more expensive logic than the `JobCrawler` - and most assessment logic - thus the scoping of the workflows/jobs has most impact in that class.)

PR #3661, opened 2025-02-10
Rename file language helper function

JCZuurmond:
## Changes
Rename and document getting supported file language helper function to make more clear what it does.

IMO it was not clear that this function contains essential decision power as it is the filter for passing files to the linter only if they are supported (in contrast with inferring any language - also the non-supported ones). 

For example, before #3655, the `FileLinter` contains filters for extensions, however this redundant as that is already handled by this helper function.

The rename makes it more explicit what this function does. The documentation makes it more clear how to use the function

### Functionality

- [x] modified linter related code.

### DIFF
--- a/src/databricks/labs/ucx/source_code/linters/files.py
+++ b/src/databricks/labs/ucx/source_code/linters/files.py
@@ -203,7 +203,7 @@ class NotebookLinter:
         # TODO deal with workspace notebooks
         language = get_file_language_if_supported(resolved)
         # we only support Python notebooks for now
-        if language is not Language.PYTHON:
+        if language != Language.PYTHON:
             logger.warning(f"Unsupported notebook language: {language}")
             return None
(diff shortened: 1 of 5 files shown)


src/databricks/labs/ucx/source_code/linters/files.py
```diff
@@ -201,9 +201,9 @@ def _resolve_and_parse_notebook_path(self, path: Path | None) -> Notebook | None
         if resolved is None:
             return None  # already reported during dependency building
         # TODO deal with workspace notebooks
-        language = file_language(resolved)
+        language = get_file_language_if_supported(resolved)
         # we only support Python notebooks for now
-        if language is not Language.PYTHON:
+        if language != Language.PYTHON:
```
pritishpai, line 206: Nice!

pritishpai (approved):

Issue #3687, opened 2025-02-13
[BUG]: lint-local-code command does not recognize local directory in windows


sreekanthmunigati:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

lint-local-code command does not recognize local directory in windows but it works find if you give the full path to a single file. 

![Image](https://github.com/user-attachments/assets/ba1fb257-d453-4d37-b82f-135bb0b325f0)

![Image](https://github.com/user-attachments/assets/fdd13095-a674-455b-837c-6d923758140d)

### Expected Behavior

lint-local-code command should lint the local directory with python files as documented. It works find in macOS  but not on Windows laptop. 

### Steps To Reproduce

1. Install UCX on windows laptop
2. run databricks labs ucx lint-local-code 
3. and select the default directory 

### Cloud

AWS

### Operating System

Windows

### Version

latest via Databricks CLI

### Relevant log output

```shell

```

JCZuurmond:
@sreekanthmunigati : Please share the debug logs as discussed. Preferably here, if they are sensitive via a non-public route.

gueniai:
This issue is dependant on getting a Windows env

PR #3711, opened 2025-02-18
Use cached property for table migration index on local checkout context

JCZuurmond:
## Changes

Use cached property for table migration index on local checkout context to avoid multiple (re)crawling in case of an empty migration index

### Linked issues

Relates to #3687

### Functionality

- [ ] modified existing command: `databricks labs ucx lint-local-code`

### Tests

- [ ] manually tested


FastLee (commented):
Can you explain why migration index will never have to be updated?

FastLee (approved):
Confirmed with Cor

PR #3734, opened 2025-02-21
Create a message to run create-ucx-catalog after assign-metastore command

FastLee:
Related to #2737

### DIFF
--- a/src/databricks/labs/ucx/account/metastores.py
+++ b/src/databricks/labs/ucx/account/metastores.py
@@ -45,11 +45,11 @@ class AccountMetastores:
             else:
                 metastore_id = list(metastore_choices.values())[0]
         self._ac.metastore_assignments.create(workspace_id, metastore_id)
+        logger.info(f"Metastore {metastore_id} assigned to workspace {workspace_id}")
         # set the default catalog using the default_namespace setting API
         if default_catalog is not None:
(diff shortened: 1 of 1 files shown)


src/databricks/labs/ucx/account/metastores.py
```diff
@@ -48,6 +48,8 @@ def assign_metastore(
         # set the default catalog using the default_namespace setting API
         if default_catalog is not None:
             self._set_default_catalog(workspace_id, default_catalog)
+        logger.info(f"Metastore {metastore_id} assigned to workspace {workspace_id}")
```
JCZuurmond, line 51: Move this to line 48

src/databricks/labs/ucx/account/metastores.py
```diff
@@ -48,6 +48,8 @@ def assign_metastore(
...
+        logger.info("Run the 'create-ucx-catalog' command to create the UCX catalog in case it does not exist")
```
JCZuurmond, line 52: Do we want to explain why?

```suggestion
        logger.info("Run the `create-ucx-catalog` command to create the UCX catalog for tracking your migration progress")
```

"in case it does not exists" is not required IMO

JCZuurmond (approved):
Made some suggestions, approving already so that you can merge after deciding what to do with the suggestions

PR #3741, opened 2025-02-24
Scoping "Seen Tables". Improving table scan performance.

FastLee:
Scoped down table scan for migrated tables.

1. When upgrading tables, looks only for the mapped tables.
2. Skipping, Delta Sharing Catalogs and Information Schema
3. Parallelizing Scan


src/databricks/labs/ucx/hive_metastore/table_migration_status.py
```diff
@@ -171,7 +191,10 @@ def _iter_catalogs(self) -> Iterable[CatalogInfo]:
 
     def _iter_schemas(self) -> Iterable[SchemaInfo]:
         for catalog in self._iter_catalogs():
-            if catalog.name is None:
+            if catalog.name is None or catalog.catalog_type in (
```
JCZuurmond, line 194: This has probably the same effect as the filter on line 186. If not, move the filter there so that filtering catalogs happens inside the `_iter_catalog`

src/databricks/labs/ucx/hive_metastore/table_migration_status.py
```diff
@@ -92,29 +96,45 @@ def __init__(self, ws: WorkspaceClient, sql_backend: SqlBackend, schema, tables_
...
         seen_tables: dict[str, str] = {}
+        tasks = []
+        schema_scope = {(table.catalog.lower(), table.database.lower()) for table in scope} if scope else None
+        table_scope = {table.full_name.lower() for table in scope} if scope else None
         for schema in self._iter_schemas():
-            if schema.catalog_name is None or schema.name is None:
+            if (
+                schema.catalog_name is None
+                or schema.name is None
+                or schema.name == "information_schema"
```
JCZuurmond, line 108: Move this into `_iter_schema`

src/databricks/labs/ucx/hive_metastore/table_migration_status.py
```diff
@@ -92,29 +96,45 @@ def __init__(self, ws: WorkspaceClient, sql_backend: SqlBackend, schema, tables_
     def index(self, *, force_refresh: bool = False) -> TableMigrationIndex:
         return TableMigrationIndex(self.snapshot(force_refresh=force_refresh))
 
-    def get_seen_tables(self) -> dict[str, str]:
+    def get_seen_tables(self, *, scope: set[Table] | None = None) -> dict[str, str]:
         seen_tables: dict[str, str] = {}
```
JCZuurmond, line 100: You can initialize scope as `scope = scope or set()` so that you can eliminate some of the `if (schema_|table_)scope` below

src/databricks/labs/ucx/hive_metastore/table_migration_status.py
```diff
@@ -92,29 +96,45 @@ def __init__(self, ws: WorkspaceClient, sql_backend: SqlBackend, schema, tables_
...
+                partial(
+                    self._iter_tables,
+                    schema.catalog_name,
+                    schema.name,
+                )
+            )
+        tables: list = []
+        logger.info(f"Scanning {len(tasks)} schemas for tables")
+        table_lists = Threads.gather("list tables", tasks, self.API_LIMIT)
+        # Combine tuple of lists to a list
```
JCZuurmond, line 122: Use `itertools.chain` instead

src/databricks/labs/ucx/hive_metastore/table_migration_status.py
```diff
@@ -92,29 +96,45 @@ def __init__(self, ws: WorkspaceClient, sql_backend: SqlBackend, schema, tables_
...
+        for table_list in table_lists[0]:
+            tables.extend(table_list)
+        for table in tables:
+            if not isinstance(table, TableInfo):
```
JCZuurmond, line 126: Is this supposed to happen? It should not based on the type hinting, right?

src/databricks/labs/ucx/hive_metastore/table_migration_status.py
```diff
@@ -92,29 +96,45 @@ def __init__(self, ws: WorkspaceClient, sql_backend: SqlBackend, schema, tables_
...
+            tasks.append(
+                partial(
+                    self._iter_tables,
+                    schema.catalog_name,
+                    schema.name,
+                )
+            )
+        tables: list = []
+        logger.info(f"Scanning {len(tasks)} schemas for tables")
+        table_lists = Threads.gather("list tables", tasks, self.API_LIMIT)
```
JCZuurmond, line 121: It is not an API limit, but the number of threads

JCZuurmond (changes requested):
@FastLee : I have to think about the choice to add a `scope` for a bit. 

It feels redundant to pass the in-scope tables, extract the schemas names from those tables, then skip the in-scope schemas when iterating over all schemas. Why not directly loop over the in-scope schemas? 

A similar approach is repeated for the table names, which makes the schema loop even more redundant. Why not directly go to the in-scope tables?

Moreover, we typically provide the scope using an `include_` at initialization, which might not work here, but I have to think about that

PR #3799, opened 2025-02-28
Pin dependencies to minor releases

JCZuurmond:
## Changes
Pin dependencies to minor releases to test dependency changes before users are affected by them.

The reasoning for choosing between pinned vs loose requirements is:
1. For applications (like UCX), pin the dependency minor versions to minimize instability due to changes in dependencies while still receiving dependency bug/patch fixes. We automatically test dependency changes in Dependabot's PRs.
2. For libraries (like pytest, lsql and blueprint), prefer loose requirements so that the libraries and projects that use your library have the freedom to chose their preferred version, especially important to increase the likelihood for finding a library version that fits the constrains of all dependencies and sub-dependencies (dependencies of dependencies of dependencies ...).

We prefer the greater than or equal to & lower than syntax over the tilde as it is more clear what the version range is

The sdk version pinning is covered by #3686
Supersedes #3766

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,15 +45,15 @@ classifiers = [
 ]
 
 dependencies = ["databricks-sdk>=0.40,<0.42",
-                "databricks-labs-lsql>=0.14.0,<0.15",
-                "databricks-labs-blueprint>=0.9.1,<0.11",
-                "PyYAML>=6.0.0,<7.0.0",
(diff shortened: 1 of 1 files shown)


JCZuurmond:
@gueniai : Can you check if Dependabot is configure to also bump the lower version range in case of a new minor version release of a dependency?

pritishpai (approved):

PR #3826, opened 2025-03-08
[Tech Debt] removing notfound if not required in test_install.py 

### LINKED ISSUE #2700: [TECH DEBT] Raise `NotFound` exception as `side_effect` in `test_install.py` unit tests
Currently we have raise `NotFound` as a `side_effect` in the `test_install.py` as follows:

``` python
ws.workspace.get_status = not_found
```

with:  https://github.com/databrickslabs/ucx/blob/2679758d817888f9d6eb929fc7ca5a752661eff1/tests/unit/install/test_install.py#L98-L100

### Solution

The following should suffice and is more correct:

``` python
ws.workspace.get_status.side_effect = NotFound
```

Note: Please verify if the side effect is needed at all. It might not be in all the test cases

HariGS-DB:
### Linked issues

Resolves #2700 

### Tests

- [ ] added unit tests

### DIFF
--- a/tests/unit/install/test_install.py
+++ b/tests/unit/install/test_install.py
@@ -96,11 +96,6 @@ def mock_installation_extra_jobs():
     )
 
 
-def not_found(_):
-    msg = "save_config"
-    raise NotFound(msg)
-
-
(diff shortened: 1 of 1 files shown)


FastLee (approved):
LGTM

PR #3963, opened 2025-03-26
Limit crawl workflows task in assessment to workflows that ran in the last 30 days.

### LINKED ISSUE #3960: assess_workflows task fails frequently when the workspace has a large number of workflows.

### Current Behavior

In cases where a workspace has a large number of workflows or very complex workflows the assess workflow task can take a long time (days) to complete and frequently fails. Linting cannot utilize the cluster and relies only on the driver which is wasteful.

### Expected Behavior

We need to scope down the task to concentrate on a smaller number of workflows.

### Steps To Reproduce

_No response_

### Cloud

AWS

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

```shell

```

FastLee:
Fix #3960

### DIFF
--- a/src/databricks/labs/ucx/progress/workflows.py
+++ b/src/databricks/labs/ucx/progress/workflows.py
@@ -181,7 +181,7 @@ class MigrationProgress(Workflow):
         """Scans all jobs for migration issues in notebooks.
         Also stores direct filesystem accesses for display in the migration dashboard."""
         # TODO: Ensure these are captured in the history log.
-        ctx.workflow_linter.refresh_report(ctx.sql_backend, ctx.inventory_database)
+        ctx.workflow_linter.refresh_report(ctx.sql_backend, ctx.inventory_database, last_run_days=30)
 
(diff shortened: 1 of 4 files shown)


src/databricks/labs/ucx/progress/workflows.py
```diff
@@ -181,7 +181,7 @@ def assess_workflows(self, ctx: RuntimeContext):
         """Scans all jobs for migration issues in notebooks.
         Also stores direct filesystem accesses for display in the migration dashboard."""
         # TODO: Ensure these are captured in the history log.
-        ctx.workflow_linter.refresh_report(ctx.sql_backend, ctx.inventory_database)
+        ctx.workflow_linter.refresh_report(ctx.sql_backend, ctx.inventory_database, last_run_days=30)
```
pritishpai, line 184: We should make `last_run_days` configurable and supplied by the customer if they have different requirements for e.g. last 2-3 months

PR #3995, opened 2025-04-11
Pause migration progress workflow schedule

FastLee:
As the workflow is experimental and fails if no UCX catalog was created by the customer, we are pausing this workflow.
We suggest that the user unpause it after running the [`create-ucx-catalog`](https://databrickslabs.github.io/ucx/docs/reference/commands/#create-ucx-catalog) command.

### DIFF
--- a/src/databricks/labs/ucx/progress/workflows.py
+++ b/src/databricks/labs/ucx/progress/workflows.py
@@ -29,7 +29,7 @@ class MigrationProgress(Workflow):
     def schedule(self) -> CronSchedule:
         """Schedule the migration progress workflow to run by default daily at 5:00 a.m. (UTC)."""
         return CronSchedule(
-            quartz_cron_expression="0 0 5 * * ?", timezone_id="Etc/UTC", pause_status=PauseStatus.UNPAUSED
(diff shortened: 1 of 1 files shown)


tests/integration/progress/test_workflows.py
```diff
@@ -54,4 +54,4 @@ def test_migration_progress_job_has_schedule(ws: WorkspaceClient, installation_c
     assert schedule is not None, "No schedule found for the migration-progress workflow."
     assert schedule.quartz_cron_expression, "No cron expression found for the migration-progress workflow."
     assert schedule.timezone_id, "No time-zone specified for scheduling the migration-progress workflow."
-    assert schedule.pause_status == PauseStatus.UNPAUSED, "Workflow schedule should not be paused."
+    assert schedule.pause_status == PauseStatus.PAUSED, "Workflow schedule should not be paused."
```
pritishpai, line 57: Will need to change the error message accordingly.

pritishpai (commented):
Minor change

tests/integration/progress/test_workflows.py
FastLee: Fixed

pritishpai (approved):

PR #4009, opened 2025-04-23
Skip integration tests with legacy dashboard creation as it is deprecated

pritishpai:
## Changes
Skip integration tests with legacy dashboard creation as it is deprecated. Skipping until we can work with product team and enable it only for Labs workspaces

### Tests
- [x] modify integration tests


FastLee (approved):
LGTM

PR #4016, opened 2025-05-01
Warns instead of an error while finding an acc group in workspace

pritishpai:
## Changes
After the introduction of creation of nested account groups from workspace local groups, the `reflect_account_groups_on_workspace` shoots out an error which instead should be a warning.

### Functionality

- [x] change error level

### Tests

- [x] existing tests


FastLee (approved):
LGTM

PR #4030, opened 2025-05-15
Update unit tests to configure log-capturing when verifying logs

asnare:
## Changes

This PR updates unit tests that verify logs to ensure they set up log-capturing to capture the level of the logs they're looking for. Previously this wasn't the case, and they relied on the underlying logging configuration have a level that would emit the record.

### Linked issues

Relates databrickslabs/blueprint#238.

### Tests

- updated unit tests


FastLee (approved):
LGTM

PR #4363, opened 2025-08-07
Use the new property `id` from RunAs in pytester

pritishpai:
## Changes
Refactor to use the newly introduced property `id` from RunAs


FastLee (approved):
LGTM

PR #4387, opened 2025-08-27
Fixed create federated catalog to add all external locations to the allowed paths.

FastLee:
Fixed "create federated catalog" to scan all external locations and add their paths to the allowed paths for the federated catalog.

### DIFF
--- a/src/databricks/labs/ucx/hive_metastore/federation.py
+++ b/src/databricks/labs/ucx/hive_metastore/federation.py
@@ -271,28 +271,10 @@ class HiveMetastoreFederation(SecretsMixin):
             return self._get_existing_connection(name)
 
     def _get_authorized_paths(self) -> str:
-        existing = {}
-        for external_location in self._ws.external_locations.list():
-            existing[external_location.url] = external_location
         authorized_paths = []
-        current_user = self._ws.current_user.me()
-        if not current_user.user_name:
-            raise NotFound('Current user not found')
-        # Get the external locations. If not using external HMS, include the root DBFS location.
-        external_locations = self._external_locations.external_locations_with_root()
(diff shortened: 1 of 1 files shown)


src/databricks/labs/ucx/hive_metastore/federation.py
```diff
@@ -271,28 +271,19 @@ def _get_or_create_glue_connection(self, name: str, aws_glue: AWSGlue, credentia
...
-                logger.warning(f'External location {location} not found')
+        for external_location in self._ws.external_locations.list():
+            if not external_location.url:
+                logger.warning("Location with no URL.")
                 continue
-            location_name = existing_location.name
+            location_name = external_location.name
             if not location_name:
-                logger.warning(f'External location {location} has no name')
+                logger.warning("Location with no name.")
```
pritishpai, line 282: Nit: Should we add some identifier here to make it easier to track down in case there are failures

src/databricks/labs/ucx/hive_metastore/federation.py
```diff
@@ -271,28 +271,19 @@ def _get_or_create_glue_connection(self, name: str, aws_glue: AWSGlue, credentia
...
                 continue
-            self._add_missing_permissions_if_needed(location_name, current_user.user_name)
-            authorized_paths.append(location)
+            authorized_paths.append(external_location.url)
```
pritishpai, line 284: Looks like the order has changed from the original where the location is appended to the list after the _add_missing_permissions_if_needed call

pritishpai (commented):
2 minor comments.

pritishpai (approved):
Minor Nits.

PR #4712, opened 2025-11-10
Fixed api compatibility with DBR16

### LINKED ISSUE #4331: Find long-term solution to convert managed tables to external without using unstable internal JVM/Scala APIs

### Current Behavior

We currently have an internal workflow to convert managed Spark tables to external tables. This relies on constructing CatalogTable objects directly from PySpark using internal JVM/Scala APIs.

This approach has proven brittle because:
- CatalogTable constructor signatures frequently change between Spark versions (e.g., Spark 3.5.x added extra fields for Delta runtime properties).
- Accessing Scala companion objects from Python via Py4J is complex (CatalogTable$.MODULE$, apply()), and breaks depending on Spark/Databricks runtime.
- There is no stable public API to directly toggle a table from managed to external.

Proposed Next Steps:
Investigate if there is a public API or safe SQL-based flow for converting a managed table to an external table.
For example : 
Use ALTER TABLE ... SET LOCATION and recreate table metadata if required or a supported SparkSession.catalog interface for externalization.

Explore implementing a JVM-side helper (Scala) to encapsulate this logic and avoid Python → Scala constructor hacks. Review and track other places where internal JVM/Scala references are used (e.g., SessionCatalog, CatalogStorageFormat.copy, CatalogTableType.EXTERNAL), and create follow-up issues for stabilizing those.

### Expected Behavior

- A robust, maintainable solution to convert managed → external tables that:
- Does not depend on unstable internal constructors.
- Works across Spark versions and Databricks runtimes.
- Can be extended to other internal APIs where we rely on Scala internals today.

### Steps To Reproduce

Run `test_table_migration_convert_manged_to_external` with a DBR 16+ cluster and check the `convert_managed_to_external` task

### Cloud

Azure

### Operating System

macOS

### Version

latest via Databricks CLI

### Relevant log output

```shell
py4j.protocol.Py4JError: An error occurred while calling None.org.apache.spark.sql.catalyst.catalog.CatalogTable. Trace:
py4j.Py4JException: Constructor org.apache.spark.sql.catalyst.catalog.CatalogTable([class org.apache.spark.sql.catalyst.TableIdentifier, class org.apache.spark.sql.catalyst.catalog.CatalogTableType, class org.apache.spark.sql.catalyst.catalog.CatalogStorageFormat, class org.apache.spark.sql.types.StructType, class scala.Some, class scala.collection.immutable.Nil$, class scala.None$, class java.lang.String, class java.lang.Long, class java.lang.Integer, class java.lang.String, class scala.collection.immutable.HashMap$HashMap1, class scala.None$, class scala.None$, class scala.None$, class scala.collection.mutable.ArrayBuffer, class java.lang.Boolean, class java.lang.Boolean, class scala.collection.immutable.Map$Map1, class scala.None$, class scala.collection.immutable.Nil$]) does not exist
	at py4j.reflection.ReflectionEngine.getConstructor(ReflectionEngine.java:203)
	at py4j.reflection.ReflectionEngine.getConstructor(ReflectionEngine.java:220)
	at py4j.Gateway.invoke(Gateway.java:255)
	at py4j.commands.ConstructorCommand.invokeConstructor(ConstructorCommand.java:80)
	at py4j.commands.ConstructorCommand.execute(ConstructorCommand.java:69)
	at py4j.ClientServerConnection.waitForCommands(ClientServerConnection.java:197)
	at py4j.ClientServerConnection.run(ClientServerConnection.java:117)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

FastLee:
fixes #4331

### DIFF
--- a/tests/unit/hive_metastore/test_table_migrate.py
+++ b/tests/unit/hive_metastore/test_table_migrate.py
@@ -3,7 +3,7 @@ import logging
 import sys
 from collections.abc import Generator
 from itertools import cycle
-from unittest.mock import create_autospec, Mock
+from unittest.mock import create_autospec
 
 import pytest
 from databricks.labs.lsql.backends import MockBackend, SqlBackend
@@ -54,11 +54,25 @@ def mock_pyspark(mocker):
 
 
 @pytest.fixture
-def mock_hms_table_dbr15(mocker):
+def mock_table_location(mocker):
+    table_location = mocker.Mock()
+    table_location.inputFormat.return_value = "org.apache.hadoop.mapred.TextInputFormat"
+    table_location.outputFormat.return_value = "org.apache.hadoop.mapred.TextOutputFormat"
+    table_location.serde.return_value = "org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe"
+    table_location.compressed.return_value = False
+    table_location.properties.return_value = {
+        "serialization.format": "1",
+        "field.delim": ",",
+    }
+    return table_location
+
+
+@pytest.fixture
+def mock_hms_table_dbr15(mocker, mock_table_location):
     table_metadata = mocker.Mock()
     table_metadata.identifier.return_value = "hive_metastore.db1_src.managed_dbfs"
     table_metadata.tableType.return_value = "MANAGED"
-    table_metadata.storage.return_value = {"location": "dbfs:/mnt/foo"}
+    table_metadata.storage.return_value = mock_table_location
     table_metadata.schema.return_value = "db1_src"
     table_metadata.provider.return_value = "DELTA"
     table_metadata.partitionColumnNames.return_value = []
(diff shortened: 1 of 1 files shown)


src/databricks/labs/ucx/hive_metastore/table_migrate.py
```diff
@@ -322,54 +322,77 @@ def _catalog_storage(self):
...
-        #    methods will be automatically created on the proxy but fail when invoked.
-        # Instead the only approach is to use dir() to check if the method exists _prior_ to trying to access it.
-        # (After trying to access it, dir() will also include it even though it doesn't exist.)
-        return table_metadata.entityStorageLocations() if 'entityStorageLocations' in dir(table_metadata) else None
-
     def _convert_hms_table_to_external(self, src_table: Table) -> bool:
         """Converts a Hive metastore table to external using Spark JVM methods."""
         logger.info(f"Changing HMS managed table {src_table.name} to External Table type.")
         inventory_table = self._tables_crawler.full_name
+        database = self._spark._jvm.scala.Some(src_table.database)  # pylint: disable=protected-access
```
pritishpai: This can be retained under the try right?

PR #4730, opened 2026-02-19
Paginate through Account Groups list request

pritishpai:
## Changes
Customer facing issues where an account group is not matching. This is due to the list account groups not paginating through the API calls. Since it is a bigger change and an account api client is not always available at workspace level, we replicate the sdk pagination here.

### Tests
- [ ] manually tested
- [ ] added unit tests
- [ ] added integration tests


pritishpai:
Functionally superseeded by #4732

PR #4739, opened 2026-04-15
Fix/supply chain lockdown

FastLee:
Based on https://gist.github.com/asnare/bd504345c1f70f9cfeecb038e6f9b9fe#3-lock-down-everything-else
We locked down the UCX deployment and replaced hatch for uv.


asnare (commented):
This looks pretty good, although some things have been missed:

 - GitHub workflows don't have top-level permissions, and write permissions haven't been scoped to just the jobs that need them. In addition the release workflow is still present. Refer to the _Workflow permissions_ section of the migration guide on what needs to happen here, as well as items 3 and 4 in the checklist.
 - Project documentation hasn't been updated: there are still lots of hatch references. Step 13 of the _Migrating to uv_ section, and item 18 on the checklist.

asnare (commented):
Some other things I noticed:

 - Jobs need to run on `larger-runners`/`larger` group/label, not just `larger`.
 - Jobs need to run on `databrickslabs-protected-runner-group`/`linux-ubuntu-latest`, not `ubuntu-latest`.
 - Although you've added the script and composite workflow needed to authenticate against JFrog, it's not used anywhere. This is needed in all the jobs after the `setup-uv` step. This is Step 9 of the _Migration from hatch to uv_ section. (Those jobs will also need `id-token: write` permission.)

.github/workflows/no-cheat.yml
```diff
@@ -10,9 +10,13 @@ on:
     branches:
       - main
 
+permissions: {}
```
asnare: I think `content: read` is needed here: otherwise the `actions/checkout` action doesn't have access to read the repo.

asnare (commented):
I think this is fine, although there's a single permissions setting that needs to be updated.

PR #4803, opened 2026-04-28
Fix three flaky integration tests

FastLee:
## Summary

Three previously-failing integration tests, plus a CI fix for a coverage-config issue that was masking the runner's exit status.

### Test fixes

- **`test_build_notebook_dependency_graphs_installs_pytest_from_index_url`** (and its installed-twice sibling) — generate the notebook fixture at test time using whichever pip mirror is configured locally. The helper shells out to `python -m pip config list` (not `pip config get`, because `get` only inspects `user`/`global`/`site` scopes and ignores the `env` scope that `PIP_CONFIG_FILE` loads as) and parses the `global.index-url='...'` line. Falls back to public PyPI when no mirror is configured. The same test now exercises the JFrog mirror in CI and the dev proxy locally without any environment knowledge baked into the test code. Depends on #4804 keeping the JFrog URL credential-free.
- **`test_installation_when_dashboard_id_is_invalid`** — skipped; it exercises a deprecated dashboard API, matching the existing skip on line 185 of the same file.
- **`test_create_account_level_groups_nested_groups`** — scoped the final assertion to the 4 groups the test actually creates, instead of asserting a global "no mismatches anywhere" log line that gets polluted by stale UCX groups left in the shared workspace by earlier runs.

### CI: pin `COVERAGE_RCFILE`

The integration job was failing with `failed: trigger: run: unknown: exit status 3` even when zero tests failed. Root cause: the `databrickslabs/sandbox/acceptance` wrapper invokes pytest in multiple per-directory sessions, and coverage.py walks up from CWD looking for config. At least one CWD wasn't surfacing the project's `pyproject.toml`, so coverage fell back to defaults (`branch=false`) and wrote line data while sibling sessions wrote arc data. The final session's `cov.combine()` then raised `Can't combine arc data with line data`. Setting `COVERAGE_RCFILE: ${{ github.workspace }}/pyproject.toml` forces every session to read the same `[tool.coverage.run]` config regardless of CWD.

## Test plan

- [x] `make fmt` (black, ruff, mypy, pylint 10.00/10)
- [x] `make test` (2011 passed, coverage 89.83%)
- [x] `labs test-one` for each of the three originally-failing integration tests
- [x] CI integration job green (76 ✅ / 0 ❌ / 11 ⏭️, all 11 pytest sessions exit 0)

### DIFF
--- a/tests/integration/source_code/test_libraries.py
+++ b/tests/integration/source_code/test_libraries.py
@@ -3,6 +3,7 @@ These tests uses part of unit testing framework to mock the path lookup, and the
 because it uses the context and the time it takes to run the test.
 """
 
+import configparser
 import logging
 import os
 import re
@@ -16,15 +17,22 @@ from tests.unit.conftest import MockPathLookup
 
 
 def _resolved_index_url() -> str:
-    """Return the pip index URL configured for the current environment.
+    """Return a pip-usable index URL for the current environment.
 
-    Local dev points at the Databricks pypi proxy; CI configures JFrog via PIP_INDEX_URL/UV_INDEX_URL.
+    The notebook fixture invokes a real `pip install --index-url <url>`. CI's jfrog-auth action
+    writes a pip.conf at ``$PIP_CONFIG_FILE`` whose ``[global] index-url`` carries the
+    JFrog token; ``UV_INDEX_URL`` exists too but lacks credentials (uv stores auth in its keyring),
+    so passing it to pip would 401. Prefer ``PIP_CONFIG_FILE`` so pip can authenticate; fall back
+    to ``PIP_INDEX_URL`` and finally to the dev proxy used in local development.
     """
-    return (
-        os.environ.get("PIP_INDEX_URL")
-        or os.environ.get("UV_INDEX_URL")
-        or "https://pypi-proxy.dev.databricks.com/simple/"
-    )
+    pip_config_file = os.environ.get("PIP_CONFIG_FILE")
+    if pip_config_file and Path(pip_config_file).is_file():
+        parser = configparser.ConfigParser()
+        parser.read(pip_config_file)
+        url = parser.get("global", "index-url", fallback=None)
+        if url:
+            return url
+    return os.environ.get("PIP_INDEX_URL") or "https://pypi-proxy.dev.databricks.com/simple/"
 
 
 def _write_pytest_with_index_url_notebook(directory: Path) -> str:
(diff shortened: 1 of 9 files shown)


tests/integration/source_code/test_libraries.py
```diff
@@ -13,6 +16,46 @@
...
+    to ``PIP_INDEX_URL`` and finally to the dev proxy used in local development.
+    """
+    pip_config_file = os.environ.get("PIP_CONFIG_FILE")
+    if pip_config_file and Path(pip_config_file).is_file():
+        parser = configparser.ConfigParser()
+        parser.read(pip_config_file)
+        url = parser.get("global", "index-url", fallback=None)
+        if url:
+            return url
+    return os.environ.get("PIP_INDEX_URL") or "https://pypi-proxy.dev.databricks.com/simple/"
```
asnare: This needs to be changed: we shouldn't be embedding environment details in the code like this and trying to emulate pip's configuration loading code is brittle. In addition, neither the dev environment (if set up using the normal instructions) nor CI/CD have `PIP_INDEX_URL` set by default.

The approach to use here is, I think:
1. Use `pip3 config get global.index-url` to figure out if a mirror has been configured.
2. If nothing has been configured, default to `https://pypi.python.org/simple`.

This means the code should be something like:
```python
cmd = [sys.executable, "-m", "pip", "config", "get", "global.index-url"]
result = subprocess.run(cmd, capture_output=True, text=True, check=False, timeout=10)
index_url = result.stdout.strip() if result.returncode == 0 else "https://pypi.python.org/simple"
```

One thing worth noting: under CI/CD, index-url includes a secret that is somewhat sensitive and could be exposed by this test especially when it fails. I'm going to change the pip setup under CI/CD to help address this, but that will be a separate PR.

asnare (changes requested):
The updates for 2 of the 3 are good, but we need to adjust the approach for `test_build_notebook_dependency_graphs_installs_pytest_from_index_url`: we shouldn't be embedding environment details in the test like that, and the approach is brittle.

tests/integration/source_code/test_libraries.py
asnare: > One thing worth noting: under CI/CD, index-url includes a secret that is somewhat sensitive and could be exposed by this test especially when it fails. I'm going to change the pip setup under CI/CD to help address this, but that will be a separate PR.

I've made the change described here in #4804.

FastLee: Addressed

asnare (approved):

PR #4804, opened 2026-04-29
Refactor pip/JFrog integration during CI/CD

asnare:
## Changes

The purpose of this PR is to adjust the way pip is configured to use JFrog during CI/CD: in particular the username/password are no longer embedded in the URL of the mirror but are instead stored separately in a `.netrc` file. This is necessary because the mirror URL isn't always handled carefully and can easily be exposed.

Further to the above:

 - The setup for `uv` has been adjusted to use the same `.netrc` mechanism, which is slightly cleaner.
 - The rest of the `jfrog-auth` action has been synced with the latest version we have.

Because the rest of the action has been synced, the commit to review is really a07085ecf7460e3bc8bda1b975285e12779ea065.

### Linked issues

Relates #4803.

### Tests

 - CI/CD tests

### DIFF
--- a/.github/actions/jfrog-auth/action.yml
+++ b/.github/actions/jfrog-auth/action.yml
@@ -1,5 +1,12 @@
 name: 'Authenticate for JFrog'
 description: 'Authenticate with JFrog using OIDC based on the GitHub repository.'
+# Some things to note:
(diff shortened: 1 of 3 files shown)


FastLee (approved):
LGTM

PR #4857, opened 2026-06-04
Fix 403 on databricks labs install lsql by passing GITHUB_TOKEN

pritishpai:
## Summary

- `databricks labs install lsql` was returning a 403 in CI because the GitHub token was not being passed to the environment during the install step.
- Adds `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}` to the `env` block of the relevant CI job in `push.yml`.

## Test plan

- [x] Confirm the `lsql fmt` step in CI no longer fails with a 403 on `databricks labs install lsql`


FastLee (approved):
Perfect!

